PR #60, opened 2012-11-06
A nice alternative to Xlib key hooking

TiMBuS:
This commit adds a new module to steal keys from /dev/input/event* instead of using the Xlib record extension.
*This code is not ready to be pulled*, but I believe it is ready for a review.

A few notes:
* This module needs the evdev module, found here: http://gvalkov.github.com/python-evdev/
* There needs to be a config option to set which /dev/input/* should be captured, as well as a config option to switch between the evdev and X11 modules -- I would have done this, but I have never programmed in python so it's a tad hard to grok a few of the finer details about this application's structure.


ezyang:
Here's another bug with this patchset: curly braces don't work, e.g. {#leftbrace} in the dictionary format. I haven't diagnosed the problem yet.

morinted:
Closing for continued discussion in #238

PR #87, opened 2013-05-30
Toggle steno input with the Sidewinder calculator button

gcr:
This change allows users to toggle Plover by pressing the calculator button.

My Sidewinder keyboard has this awesome little "Calculator" button in the upper right hand corner. With this patch, the calculator button becomes a first-class steno stroke, which means I can map it to `{PLOVER:TOGGLE}`, or any other key combination in `dict.json`. This is more convenient than mapping an actual steno stroke to `{PLOVER:TOGGLE}` because the calc button doesn't cause any on-screen input, so there's no need to press backspace to clean the gibberish up.

A better patch would add support for dedicated {f1}...{f12} keys or perhaps allow for modifiers {shift+calc}, etc.

This patch probably isn't a great way of doing this, but it scratches my itch, so I thought I'd share.

### DIFF
--- a/plover/assets/dict.json
+++ b/plover/assets/dict.json
@@ -1,4 +1,5 @@
 { 
+"{calc}": "{PLOVER:TOGGLE}",
 "KH-FG": " {#grave}",
 "-FPLT": "{.}",
 "STPH": "{?}",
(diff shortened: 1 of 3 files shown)


morinted:
I'm going to close this as I think that if we were to implement a toggle-by-key, the user should be able to configure which key they would like. But it isn't a huge priority at the moment in the face of other important changes. Thanks, though!

PR #163, opened 2013-12-13
fix stentura high CPU usage

rbrewer123:
This patch is a fix for high CPU usage in stentura.py.  It sets a small default timeout on the serial port (0.05 seconds) for reading data.  That timeout becomes the granularity for interrupting the process of reading data in _read_data().  _read_data() attempts to read as much data as it can within that small timeout period.  I moved the check for the overall timeout up to _read_packet().

Verified it functions with low CPU usage on Ubuntu 13.04.

### DIFF
--- a/plover/machine/stentura.py
+++ b/plover/machine/stentura.py
@@ -427,7 +427,7 @@ def _validate_response(packet):
 
 
 # Timeout is in seconds, can be a float.
-def _read_data(port, stop, buf, offset, timeout):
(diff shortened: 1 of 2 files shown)


morinted:
Still breaking unit tests. I think I prefer the solution in #200. Please comment if you think otherwise! I don't have a Stentura to test, either.

rbrewer123:
I'm not in a position to test either fix right now, but I agree that #200 looks like a nicer solution.

Issue #138, opened 2013-11-15
Spelling mistakes


balshetzer:
weapon + ized -> weaponnized. Should be weaponized.
heptamer + ized -> heptamerrized. Should be heptamerized.

morinted:
Short of the really long medical terms here, all these spelling mistakes are fixed with the new orthography rules and the incoming dictionary update.

stenoknight:
Fantastic!

On Mon, Sep 28, 2015 at 5:44 PM, Ted Morin <notifications@github.com> wrote:

> Short of the really long medical terms here, all these spelling mistakes
> are fixed with the new orthography rules and the incoming dictionary update.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openstenoproject/plover/issues/138#issuecomment-143882767>
> .
>



-- 
Mirabai Knight, CCP, RDR
StenoKnight CART Services
917 576 4989
mkk@stenoknight.com
http://stenoknight.com

PR #190, opened 2014-04-08
Try to further improve the english orthography rules

samlh:
Inspired by issue #138

Warning: I wrote this on a machine without python, so it may be entirely broken

### DIFF
--- a/plover/test_orthography.py
+++ b/plover/test_orthography.py
@@ -97,7 +97,14 @@ class OrthographyTestCase(unittest.TestCase):
             ('white', 'er', 'whiter'),
             ('crap', 'y', 'crappy'),
             ('lad', 'er', 'ladder'),
-        
+            ('translucent', 'cy', 'translucency'),
+            ('bankrupt', 'cy', 'bankruptcy'),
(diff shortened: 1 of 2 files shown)


plover/test_orthography.py
```diff
@@ -97,7 +97,14 @@ def test_add_suffix(self):
             ('white', 'er', 'whiter'),
             ('crap', 'y', 'crappy'),
             ('lad', 'er', 'ladder'),
-        
+            ('translucent', 'cy', 'translucency'),
+            ('bankrupt', 'cy', 'bankruptcy'),
+            ('inadequate', 'cy', 'inadequatecy'), # TODO: inadequacy
+            ('secret', 'cy', 'secretcy'), # TODO: secrecy
```
morinted: Would you like to finish your TODOs before this pull request is tested?

morinted:
Can you confirm what the purpose of your TODOs are? Would you like to handle those cases?

samlh:
Sorry, I had totally forgotten this. I've pushed a 2 commits to remove the remaining TODOs - I think this should be good to go.

morinted:
Thanks @samlh, it all checks out.

PR #198, opened 2014-06-11
Added Taskbar Icon

David-Allison:
Added an icon to the taskbar showing the status of Plover. 
Allows toggling of Plover on double-click.


morinted:
Before this is merged I would like to see at least:

- The option to disable Plover in the taskbar
- The option to minimize to the task bar
- A context-menu for the task bar icon that provides shortcuts to "Configuration..." and "About...", as well as "Enable/Disable" (contextual) and "Quit"

Are you willing to try this? If you are unable, I can help test on Mac, Linux, and Windows.

morinted:
I'm going to close this for now. Feel free to re-open it if you make it more configurable. I'll be available for testing cross platform and then we can reopen when it is near-ready.

PR #200, opened 2014-07-29
Remove busy loop from the stentura serial port reading

Elktro:
The while loop polling serial port buffer caused 100% CPU usage and was replaced by code that rises the timeout exception if read() method returns less bytes as requested.

### DIFF
--- a/plover/machine/stentura.py
+++ b/plover/machine/stentura.py
@@ -427,7 +427,7 @@ def _validate_response(packet):
 
 
 # Timeout is in seconds, can be a float.
-def _read_data(port, stop, buf, offset, timeout):
+def _read_data(port, stop, buf, offset, num_bytes):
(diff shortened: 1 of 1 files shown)


morinted:
Can you confirm this is still a problem in Plover's code and that this fix works? I don't have a Stentura and would love your feedback to help get this into the code base.

Elktro:
I can confirm that this is still problem with the SHA: e5c4aeb91f03f49e5e10d7f54c1dc2c2f35c3f2f. I merged stentura_busy_loop branch to it which fixed the issue.

I have been using this fix for quite some time without any problem with my Stentura 400 SRT machine.

morinted:
Okay, would you like to set the commit's author to have the same email address as you so that the history has your GitHub account?

benoit-pierre:
Note: this change broke the Stentura tests.

PR #204, opened 2014-08-12
Retrospective commands

mighele:


mighele:
I have just re-tested all of this and am still happy that it all works. Unit tests have been put in place for the new functionality and old tests still pass.

PR #213, opened 2014-10-30
Rewrite keyboard emulation for Windows

morinted:
The keyboard emulation through SendKeys had some issues. Namely, it was not possible to emulate Windows key or a lot of special keys. It also didn't emulate keys, rather it sent characters to the foreground application.

I have rewritten the emulation using SendInput, which is a little more low-level. It can emulate most keys on the OS level, and it uses CTypes. It also has support for outputting Unicode characters using VK_PACKET. For more information, the commit messages have details surrounded around the various commits.

### DIFF
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -13,13 +13,27 @@ emulate keyboard input.
 
 """
 
-import re
+import collections
+import ctypes
 import functools
 import threading
-
+import types
+import win32api
 import pyHook
-from pywinauto.SendKeysCtypes import SendKeys as _SendKeys
 
+SendInput = ctypes.windll.user32.SendInput
+MapVirtualKey = ctypes.windll.user32.MapVirtualKeyW
+GetKeyboardLayout = ctypes.windll.user32.GetKeyboardLayout
+LONG = ctypes.c_long
+DWORD = ctypes.c_ulong
+ULONG_PTR = ctypes.POINTER(DWORD)
+WORD = ctypes.c_ushort
+KEYEVENTF_EXTENDEDKEY = 0x0001
+KEYEVENTF_KEYUP = 0x0002
+KEYEVENTF_SCANCODE = 0x0008
+KEYEVENTF_UNICODE = 0x0004
+INPUT_MOUSE = 0
+INPUT_KEYBOARD = 1
 
 # Global state used by KeyboardCapture to support multiple instances.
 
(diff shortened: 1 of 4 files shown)


morinted:
Is there anything I can do to help you with this pull request? I've been using it since November and it really does work better in general in terms of program compatibility on Windows.

shayneholmes:
I've built this independently and verified that it works with Z-Type at least (fun!), and doesn't appear to break anything else. My use cases include terminal services (remote desktop), universal apps on windows 10 preview, and a lot of free form writing in WriteMonkey. +1 for this change!

morinted:
Thanks Shayne,

I should add that I have used this change for the months since releasing
it, and have had no issues, and do benefit from additions like media keys,
the use of the Windows key, on top of the increased application support.

Ted

mighele:
Hi Ted.

I've just tested this and it doesn't seem to work with a UK keyboard layout. I have a dictionary entry for "That's" and "It's" and it shows them as "That#s" and "It#s".

I also gave it a test with a French keyboard layout and the results were "That²s" and "It²s".

Thanks
Mike

morinted:
Right, because this emulates the physical pressing of keys, that makes
sense. That's a big shame. I wonder if there's a way to reconcile the
benefits of each method.

balshetzer:
I seem to remember that there is a Windows API for accessing the keyboard
layout. Then you can use your method for keys available on the keyboard by
using keycodes and falling back to the old method for chars not on the
current keyboard.

morinted:
@mighele I think that non-US layouts should work on this now, so you should try it out :)

Also, this is ready for review. Please scrutinize away. (pinging also @chajadan)

I've also made a build for anyone who just wants to try it without actually building: https://www.dropbox.com/s/lp2q77mapulekkq/plover.exe?dl=0

morinted:
Playing around some more, I think that:

- the global manipulating function needs to add not found characters to the key name to Unicode list.
- we might consider changing Alt-Gr characters to Unicode only, because Alt-Gr can have some weird effects (like the effects of ` and ^ are weird because when Alt-Gr'd, they are supposed to be used as accents)

I will be investigating this tonight.

morinted:
I've got some work to do on this, going to clean up the variables a little bit and try to simplify things, then flatten the history. Overall, though, I still find myself needing this change in my day-to-day on Windows.

morinted:
@benoit-pierre I rewrote it. Try it out! :D

PR #215, opened 2014-11-03
Add different output modes: Caps, Title, Lowercase, Snake, Camel

morinted:
This pull adds several different cases. Namely:

- CAPITAL CASE
- Title Case
- lower case
- snake_case
- CamelCase

They are controlled by "MODE" commands, similar to "PLOVER" commands. They can be toggled with CAPS, TITLE, LOWER, SNAKE, and CAMEL. They can all be turned off with CLEAR.

CAPS, Title case, and lower case are mutually exclusive, so turning on one will turn off the others. CamelCase is mutually exclusive to all other cases. Snake case is only exclusive with camel case.

This means CAPS, title, and lower case can be used with snake case. This functionality is useful for programmers and regular stenographers alike. The cases are useful for programmers while coding, such as in variable names. For regular stenographers, CAPS can represent yelling, or scene cues, for example. Title case is useful for journalism.

I've tried to keep some sanity while coding these. The initial implementation used global variables, but this wasn't testable. Now, a dict of the cases has been added as a variable to the Action class. Since the cases are persistent, the case is passed along on each stroke. The MODE command directly modifies the mode of the action.

Some sample strokes for testing:

    "KA*PS": "{MODE:CAPS}",
    "K-BGS": "{MODE:CAMEL}",
    "TAO*EULT": "{MODE:TITLE}",
    "STPHA*EUBG": "{MODE:SNAKE}",
    "HRO*ER": "{MODE:LOWER}",
    "KHRAO*ER": "{MODE:CLEAR}",

Hope the code is up to standard -- if not, I'd be happy to make it so.

Ted

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -11,16 +11,18 @@ from os.path import commonprefix
 from collections import namedtuple
 from plover import orthography
 import re
+import string
+
 
 class Formatter(object):
     """Convert translations into output.
 
     The main entry point for this class is format, which takes in translations
     to format. Output is sent via an output class passed in through set_output.
-    Other than setting the output, the formatter class is stateless. 
+    Other than setting the output, the formatter class is stateless.
 
-    The output class can define the following functions, which will be called if
-    available:
+    The output class can define the following functions, which will be called
+    if available:
 
     send_backspaces -- Takes a number and deletes back that many characters.
 
@@ -35,7 +37,7 @@ class Formatter(object):
     """
 
     output_type = namedtuple(
-        'output', ['send_backspaces', 'send_string', 'send_key_combination', 
+        'output', ['send_backspaces', 'send_string', 'send_key_combination',
                    'send_engine_command'])
 
     def __init__(self):
@@ -67,7 +69,8 @@ class Formatter(object):
         self._output = output_type(*[getattr(output, f, noop) for f in fields])
 
     def set_space_placement(self, s):
-        """Set whether spaces will be inserted before the output or after the output"""
+        # Set whether spaces will be inserted
+        # before the output or after the output
         if s == 'After Output':
             self.spaces_after = True
         else:
@@ -78,12 +81,12 @@ class Formatter(object):
 
         Arguments:
 
-        undo -- A sequence of translations that should be undone. The formatting
-        parameter of the translations will be used to undo the actions that were
-        taken, if possible.
+        undo -- A sequence of translations that should be undone. The
+        formatting parameter of the translations will be used to undo the
+        actions that were taken, if possible.
 
-        do -- The new actions to format. The formatting attribute will be filled
-        in with the result.
+        do -- The new actions to format. The formatting attribute will be
+        filled in with the result.
 
         prev -- The last translation before the new actions in do. This
         translation's formatting attribute provides the context for the new
@@ -93,14 +96,16 @@ class Formatter(object):
         for t in do:
             last_action = _get_last_action(prev.formatting if prev else None)
             if t.english:
-                t.formatting = _translation_to_actions(t.english, last_action, self.spaces_after)
+                t.formatting = _translation_to_actions(t.english, last_action,
+                                                       self.spaces_after)
             else:
-                t.formatting = _raw_to_actions(t.rtfcre[0], last_action, self.spaces_after)
+                t.formatting = _raw_to_actions(t.rtfcre[0], last_action,
+                                               self.spaces_after)
             prev = t
 
         old = [a for t in undo for a in t.formatting]
         new = [a for t in do for a in t.formatting]
-        
+
         min_length = min(len(old), len(new))
         for i in xrange(min_length):
             if old[i] != new[i]:
@@ -113,6 +118,7 @@ class Formatter(object):
 
         OutputHelper(self._output).render(old[i:], new[i:])
 
+
 class OutputHelper(object):
     """A helper class for minimizing the amount of change on output.
 
@@ -124,7 +130,7 @@ class OutputHelper(object):
         self.before = ''
         self.after = ''
         self.output = output
-        
+
     def commit(self):
         # Python narrow Unicode is useless for long characters
         # UTF-32 is a good container to count characters
@@ -151,17 +157,18 @@ class OutputHelper(object):
                 self.before += a.text
 
         self.after = self.before
-        
+
         for a in reversed(undo):
             if a.text:
                 self.after = self.after[:-len(a.text)]
             if a.replace:
                 self.after += a.replace
-        
+
         for a in do:
             if a.replace:
                 if len(a.replace) > len(self.after):
-                    self.before = a.replace[:len(a.replace)-len(self.after)] + self.before
+                    self.before = a.replace[
+                        :len(a.replace)-len(self.after)] + self.before
                     self.after = ''
                 else:
                     self.after = self.after[:-len(a.replace)]
@@ -175,9 +182,11 @@ class OutputHelper(object):
                 self.output.send_engine_command(a.command)
         self.commit()
 
+
 def _get_last_action(actions):
     """Return last action in actions if possible or return a blank action."""
     return actions[-1] if actions else _Action()
+
 
 class _Action(object):
     """A hybrid class that stores instructions and resulting state.
@@ -187,8 +196,13 @@ class _Action(object):
     context to render future translations.
 
     """
-    def __init__(self, attach=False, glue=False, word='', capitalize=False, 
-                 lower=False, upper=False, upper_carry=False, orthography=True, text='', replace='', combo='',
+
+    CASE_NONE, CASE_UPPER, CASE_LOWER, CASE_TITLE = range(4)
+
+    def __init__(self, attach=False, glue=False, word='', capitalize=False,
+                 lower=False, orthography=True, space_char=' ',
+                 upper=False, upper_carry=False,
+                 case=CASE_NONE, text='', replace='', combo='',
                  command=''):
         """Initialize a new action.
 
@@ -214,6 +228,11 @@ class _Action(object):
         othography -- True if orthography rules should be applies when adding
         a suffix to this action.
 
+        space_char -- this character will replace spaces after all other
+        formatting has been applied
+
+        case -- an integer to determine which case to output after formatting
+
         text -- The text that should be rendered for this action.
 
         replace -- Text that should be deleted for this action.
@@ -221,7 +240,7 @@ class _Action(object):
         combo -- The key combo, in plover's key combo language, that should be
         executed for this action.
 
-        command -- The command that should be executed for this actions.
+        command -- The command that should be executed for this action.
 
         """
         # State variables
@@ -233,13 +252,17 @@ class _Action(object):
         self.upper = upper
         self.upper_carry = upper_carry
         self.orthography = orthography
-                
+
+        # Persistent state variables
+        self.space_char = space_char
+        self.case = case
+
         # Instruction variables
         self.text = text
         self.replace = replace
         self.combo = combo
         self.command = command
-        
+
     def copy_state(self):
         """Clone this action but only clone the state variables."""
         a = _Action()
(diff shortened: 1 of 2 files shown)


hhirsch:
Why isn't this merged?

morinted:
Two things other than the quality of code itself that I can think of: title
case doesn't work properly with multi-part words, especially because python
applies it to words like "That'S". Maybe we could make our own local title
case function to be more useful for most Plover users. Second thing: maybe
a counter could be implemented. Right now we have the ability to change
modes, it might be cool to add a counter. Omitted or 0 means "always
on/infinite", and 1 would mean "next one translation" and 10 would mean
"next 10 translations". That way, you could say "caps lock mode for 1 word"
and then do it and not have to remember to clear it. That would speed up
certain things. Or you could have a stroke for snake case for only 2
strokes if you wanted to use it to append strokes. (also, I specify
translation because of multi-word entries. I think that counting words and
trying to define what's a word and what isn't is no fair to the program or
its users. One translation is clear.)

In addition to those, Hesky would have to review my code, and I imagine
that it's not the ideal implementation, as I "wrote first, designed later"
with this one. This is the kind of thing that might warrant a rewrite if we
can thing of a better design. That being said, Hesky, if you can think of a
better design, I'd be willing to rewrite as I think this feature is
critical for Plover, and I use it now as a regular feature on my end of the
woods.

Ted

balshetzer:
The main reason why a pull requests isn't merged, unless I've given another reason, is just that I haven't had the time for a full review and testing.

hhirsch:
Life is short. I pull it now and we can fix it later.

plover/formatting.py
```diff
@@ -733,37 +795,131 @@ def _atom_to_action_spaces_after(atom, last_action):
...
+
+
+def _apply_case(text, case, begin):
+    action = _Action()
+    if case == action.cases['LOWER']:
+        text = text.lower()
+    elif case == action.cases['UPPER']:
+        text = text.upper()
+    elif case == action.cases['TITLE']:
+        text = text.title()
```
morinted: .title() is not ideal because it title cases "isn't" as "Isn'T".

morinted:
@mighele @chajadan same for this one, time to review if you can -- I fixed the issues that I had with this. Test it out :smile_cat: Title case now works quite well!

morinted:
I really hate title case, I must say. Now when fingerspelling, subsequent capitals get lowercased. For example, if I fingerspell "CAT" the output will invariably be "Cat". While it is "true" title case, I don't like this behavior. It would turn iPhone into Iphone or iphone (depending on if the user has set their stroke with {>})

I'll be investigating this tonight.

morinted:
The way I addressed this was to actually ignore glued/appended strokes with Title Casing. The Way That It Works Is Basically, As I Type Output Is Titled. However, Using A Manual Space Or A TK-LS Or Any Glue Command Will Set Plover To {^} Mode. This Is Interesting Because Now if I Wanted to Lowercase Certain Words in a Sentence All I Need to do is Space ("S-P") and My Output Is Normal.

Furthermore, it's cool because now the user can fingerspell without the case mode getting in the way, just in case they need to type something styled like ShoeBOX or iPhone in a title.

morinted:
Hi @benoit-pierre, if you could, I'd appreciate if you could check this PR as well. I'm hoping to merge it, still.

benoit-pierre:
One test is failing:

```python

self = <plover.test_formatting.FormatterTestCase testMethod=test_apply_case>

    def test_apply_case(self):
        an_action = action()
        # case, text, begin
        test = ' some test '
        test2 = 'test Me'
        test3 = ' SOME TEST '
        cases = [
            # UNDEFINED
            ((test, '', False), (test)),
            ((test, 'TEST', False), (test)),
            # TITLE
            ((test, an_action.cases['TITLE'], False), (' Some Test ')),
            ((test, an_action.cases['TITLE'], True), (' Some Test ')),
            # TITLE responds to 'lower'
            ((test2, an_action.cases['TITLE'], True), ('test Me')),
            # LOWER
            ((test, an_action.cases['LOWER'], False), (' some test ')),
            ((test3, an_action.cases['LOWER'], False), (' some test ')),
            ((test2, an_action.cases['LOWER'], True), ('test me')),
            # UPPER
            ((test.upper(), an_action.cases['UPPER'], False), (' SOME TEST ')),
            ((test3, an_action.cases['UPPER'], False), (' SOME TEST ')),
            ((test2, an_action.cases['UPPER'], True), ('TEST ME')),
        ]

>       self.check_arglist(formatting._apply_case, cases)

plover/test_formatting.py:858:

plover/test_formatting.py:53: in check_arglist
    self.assertEqual(actual, expected)
E   AssertionError: ' some test ' != ' Some Test '

```

morinted:
Ah, yeah, I redefined the behavior of Title Case to be less intrusive. Updated the test to reflect the behavior.

plover/formatting.py
```diff
@@ -415,6 +454,7 @@ def _atom_to_action_spaces_before(atom, last_action):
     last_upper = last_action.upper
     last_upper_carry = last_action.upper_carry
     last_orthography = last_action.orthography
+    begin = False  # for meta attach
```
morinted: Don't think I'm using this anymore

plover/formatting.py
```diff
@@ -347,6 +372,17 @@ def _translation_to_actions(translation, last_action, spaces_after):
 META_ATTACH_FLAG = '^'
 META_KEY_COMBINATION = '#'
 META_COMMAND = 'PLOVER:'
+META_MODE = 'MODE:'
+MODE_CAPS = 'CAPS'
```
morinted: I'm thinking these should be put into a tuple, MODES

plover/formatting.py
```diff
@@ -187,8 +196,13 @@ class _Action(object):
     context to render future translations.
 
     """
-    def __init__(self, attach=False, glue=False, word='', capitalize=False, 
-                 lower=False, upper=False, upper_carry=False, orthography=True, text='', replace='', combo='',
+
+    cases = {'NONE': 0, 'UPPER': 1, 'LOWER': 2, 'TITLE': 3}
```
morinted: Rename to CASES?

morinted:
This wasn't working with "after" spaces. I fixed that. Also made the changes outlined above, except for my MODES set. Discovered bug: if a stroke is undefined, the mode gets unset. This needs to be fixed.

morinted:
Thinking about it a little more, I don't mind losing the case mode on misstroke. @benoit-pierre, I'm going to merge this. Any last words before it's in?

plover/formatting.py
```diff
@@ -187,8 +196,13 @@ class _Action(object):
     context to render future translations.
 
     """
-    def __init__(self, attach=False, glue=False, word='', capitalize=False, 
-                 lower=False, upper=False, upper_carry=False, orthography=True, text='', replace='', combo='',
+
+    CASES = {'NONE': 0, 'UPPER': 1, 'LOWER': 2, 'TITLE': 3}
```
benoit-pierre: IMHO this would be better:
```python
CASE_NONE, CASE_UPPER, CASE_LOWER, CASE_TITLE = range(4)
```

plover/formatting.py
```diff
@@ -187,8 +196,13 @@ class _Action(object):
...
+
+    def __init__(self, attach=False, glue=False, word='', capitalize=False,
+                 lower=False, orthography=True, space_char=' ',
+                 upper=False, upper_carry=False,
+                 case=0, text='', replace='', combo='',
```
benoit-pierre: Should be:
```python
                 case=CASES['NONE'], text='', replace='', combo='',
```

plover/formatting.py
```diff
@@ -347,6 +372,17 @@ def _translation_to_actions(translation, last_action, spaces_after):
...
+MODE_CAPS = 'CAPS'
+MODE_TITLE = 'TITLE'
+MODE_LOWER = 'LOWER'
+MODE_SNAKE = 'SNAKE'
+MODE_SET_SPACE = 'SET_SPACE:'
+MODE_CLEAR_SPACE = 'CLEAR_SPACE'
+MODE_CLEAR_CASE = 'CLEAR_CASE'
+MODE_CAMEL = 'CAMEL'
+MODE_CLEAR = 'CLEAR'
+
```
benoit-pierre: I think `RESET` instead of `CLEAR` might be clearer.

plover/formatting.py
```diff
@@ -730,44 +794,147 @@ def _atom_to_action_spaces_after(atom, last_action):
...
+    command should be:
+        CAPS, LOWER, TITLE, CAMEL, SNAKE, CLEAR_SPACE,
+            CLEAR_CASE, SET_SPACE or CLEAR
+
+        CAPS: Uppercase
+        LOWER: Lowercase
+        TITLE: Titlecase
+        CAMEL: Titlecase, no space
+        SNAKE: Underscore space
+        CLEAR_SPACE: no space
```
benoit-pierre: Or rather: `CLEAR_SPACE: reset space character to default (' ')`, no?

plover/formatting.py
```diff
@@ -730,44 +794,147 @@ def _atom_to_action_spaces_after(atom, last_action):
...
+    else:
+        return text
+
+
 def _get_meta(atom):
     """Return the meta command, if any, without surrounding meta markups."""
     if (atom is not None and
         atom.startswith(META_START) and
-        atom.endswith(META_END)):
+            atom.endswith(META_END)):
```
benoit-pierre: Why the indentation change?

plover/formatting.py
```diff
@@ -730,44 +794,147 @@ def _atom_to_action_spaces_after(atom, last_action):
...
     return atom.replace(META_ESC_START, META_START).replace(META_ESC_END,
                                                             META_END)
 
+
 def _get_engine_command(atom):
     """Return the steno engine command, if any, represented by the atom."""
     if (atom and
         atom.startswith(META_START + META_COMMAND) and
-        atom.endswith(META_END)):
+            atom.endswith(META_END)):
```
benoit-pierre: Ditto.

plover/test_formatting.py
```diff
@@ -721,9 +757,116 @@ def test_atom_to_action_spaces_after(self):
...
+    def test_apply_space_char(self):
+        # (text, space_char)
+        test = ' some test '
+        test2 = 'don\'t'
+        cases = [
+            ((test, '_'), ('_some_text_')),
+            ((test, ''), ('sometext')),
+            ((test2, '_'), (test2)),
+            ((test2, ''), (test2)),
+        ]
```
benoit-pierre: No checks?

morinted:
Thanks for the feedback! I'll get to this soon.

plover/formatting.py
morinted: PEP8 E125,

Wants a difference in the indentation from the next line to show separation:

1. Valid:

```python
if (thing and
    and thing
        and thing):
    do_thing()
```

2. Invalid:

```python
if (thing and
    and thing
    and thing):
    do_thing()
```

3. Valid:

```python
if (thing and
        and thing
        and thing):
    do_thing()
```

I could switch to (3).

benoit-pierre: Meh, 1 looks ugly, sadly 2 is invalid, 3 is better then... It would have been better if those formatting and whitespace changes had been done in a separate commit: make using `git blame` easier.

morinted: You're right. I was a bad person. :cry:

morinted:
All right, I fixed the loss of case when you use an undefined stroke, and I updated according to your review.

morinted:
If all is good, I'll squash and merge, just give me the go ahead.

benoit-pierre:
:+1:

arxanas:
For anyone reading this after the fact, `MODE:CLEAR` is now called `MODE:RESET`. The full list of mode strings can be consulted on [L375-384](https://github.com/openstenoproject/plover/pull/215/files#diff-7c608006e88cb876893096ed894f2577R375). Not sure if the modes are documented anywhere else.

morinted:
Only on the release notes, I plan on making a dictionary format doc which
will cover case modes, key combos, and more.

davidkitfriedman:
So with no mode setting doing TUPBT/S correctly yields: 

"opportunities"

But with the `MODE:CAPS` setting doing TUPBT/S yields:

"OPPORTUNITYIES"

I haven't looked further, but other words that end in "Y" work fine in both modes such as "city", and "sky".

Issue #216, opened 2014-11-05
Unicode Characters above 16 bytes cause problems for backspacing and output


morinted:
NOTE: This Issue will not display correctly on some browsers/OS's. Firefox on any platform and OS X should work well.

Unicode characters that are above 16 bytes, like emoji and many non-latin characters, cause issues for Plover. This is because the "len" operator on these does not return valid values.

I discovered this issue while playing with emoji. To replicate the issue, create an entry with an emoji:

    "S": "😸",

Type the character, then "*" undo command. 2 backspaces will be sent instead of one.

Another place that the issue shows itself is if you are replacing an emoji with another one that has the same starting byte. Try these strokes:

    "S": "😸",
    "S/S": "😹",

The first stroke will output correctly, but the second will backspace once and then output a non-existent character. This is because Plover is trying to send half a byte.

morinted:
The fix is merged and emojis and other long characters work 💩 🐈

Achim63:
Great, thank you! The issue with incorrect backspacing is really unnerving at times.

PR #217, opened 2014-11-05
Fix counting and output of long Unicode characters

morinted:
Addresses #216 by encoding into UTF-32 before calculating common prefix,then decoding back to Unicode for output through the OS layer.

It works because UTF-32 is fixed-width, so each character always takes 4 "lengths". That's why when we get common prefix, I divide then multiply by 4. That rounds the result to get you the right count. I've tested against the issues I mentioned in the issue.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -106,11 +106,17 @@ class OutputHelper(object):
         self.output = output
         
     def commit(self):
-        offset = len(commonprefix([self.before, self.after]))
-        if self.before[offset:]:
-            self.output.send_backspaces(len(self.before[offset:]))
-        if self.after[offset:]:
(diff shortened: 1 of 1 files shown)


morinted:
@balshetzer I'd recommend taking a look at this one. It's a very small change and makes a big difference. As you've probably heard, I'm on OS X now and using long Unicode characters would actually crash Safari when Plover outputted them wrong, so I had to change to my branch to fix it.

You can try it out with these translations:

```
"A": "😢"
"A/A": "😢😭😓"
```

You'll see it on Windows that you lose a space and one of the emojis comes out wrong, but you'll see it on Mac clearly as it really messes things up quickly.

balshetzer:
Thanks Ted!

PR #219, opened 2014-11-22
Blackhole keyboard output on X11 using XGrabKeyboard, fixes #41

### LINKED ISSUE #41: Capture keystrokes without outputting them on linux
Currently the qwerty-to-steno implementation of Plover sends qwerty versions of keystrokes entered by the user, sends a corresponding number of backspace commands to delete the keystrokes, and then sends the strokes corresponding to the correct steno translation of the original keystrokes. This works fairly well in most applications, but in cases where single alphanumeric keys are mapped to commands (such as in Vim or Firefox), Plover cannot be used with a qwerty keyboard because the qwerty strokes will execute before Plover has a chance to delete them. In order to solve this, Plover needs to control the keyboard at a lower level, so that it can prevent the strokes from being sent out in the first place.

As of the most recent release, this problem is fixed in the Windows version.

ezyang:
The upside: delete emulation is no longer necessary!

The downside: GTK's default textbox seems to hide the cursor when
someone else has grabbed the keyboard (for example, when using gedit),
so the editing experience gets worse for many situations.

Signed-off-by: Edward Z. Yang <ezyang@cs.stanford.edu>

### DIFF
--- a/plover/gui/keyboard_config.py
+++ b/plover/gui/keyboard_config.py
@@ -3,12 +3,19 @@
 
 import wx
 from wx.lib.utils import AdjustRectToScreen
+import sys
 
 DIALOG_TITLE = 'Keyboard Configuration'
 ARPEGGIATE_LABEL = "Arpeggiate"
 ARPEGGIATE_INSTRUCTIONS = """Arpeggiate allows using non-NKRO keyboards.
 Each key can be pressed separately and the space bar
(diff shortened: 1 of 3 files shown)


morinted:
#259 was merged and it does not hide the cursor except while the user is holding down steno keys. Thanks for the effort and correspondence, though

PR #222, opened 2014-11-29
Linux: fix SetTopApp crash

feend78:
It looks like when the LOOKUP feature was added and O/S-specific utilities were refactored out into plover.gui.util, the Linux version of SetTopApp was still using the TITLE global (now defined in both lookup.py and add_translation_py).

Modified the SetTopApp function in plover.gui.util to require a dialog object, confirms it has rendered, and passes the title to wmctrl to bring it to front. Also modified plover.gui.lookup and plover.gui.add_translation
to accommodate the util change.

For the problem of returning focus, it seems as though calling Destroy() on a child window returns focus to the parent. The original on_close handlers called SetForegroundWindow before Destroy. Reversed that order, and it also seemed to require a call to Update after Destroy to make sure the dialog has un-drawn and the default focus change happened before returning focus to the original application.

This is all working as I would expect, but I do notice that if I run plover from the command line and open, then close either dialog, the first stroke entered into the original application prints an error on the console:
```
<class 'Xlib.error.BadWindow'>: code = 3, resource_id = Xlib.xobject.resource.Resource(0x042001ab), sequence_number = 69, major_opcode = 25, minor_opcode = 0
```
The dialogs behave as expected, and the stroke prints correctly, but the message still appears, but only for the first stroke after closing the dialog.

I'm really new to the codebase (and any kind of desktop app development), nor do I totally have my head wrapped around the guts of Plover, so I don't really know whether this is a serious problem. Any advice on running it down?

### DIFF
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -130,12 +130,13 @@ class AddTranslationDialog(wx.Dialog):
 
     def on_close(self, event=None):
         self.engine.translator.set_state(self.previous_state)
+        self.other_instances.remove(self)
+        self.Destroy()
+        self.Update()
         try:
             util.SetForegroundWindow(self.last_window)
         except:
             pass
-        self.other_instances.remove(self)
-        self.Destroy()
 
     def on_strokes_change(self, event):
         key = self._normalized_strokes()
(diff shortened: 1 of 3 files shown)


mighele:
Hi feend78.

I can't reproduce the "Xlib.error.BadWindow" problem on windows. Looks like it might just be a linux thing.

I've tested this change on windows and it doesn't seem to break anything. I'm unable to test it on Linux as I don't have a Linux machine setup. I'm happy to take your word that it fixes the issue :)

I notice the change affects add_translation and lookup dialogs. Does it also need to be added for dictionary_editor and paper_tape dialogs?

Thanks
Mike

PR #223, opened 2014-12-01
add installation instructions for Fedora

rekado:
I documented what packages were needed to get plover to run on Fedora 20.


morinted:
I don't plan on updating pypi for now so if you'd like to resubmit this, just update and reopen this with the full build instructions.

PR #224, opened 2014-12-05
replace fixed key mapping with customisable keymap

rekado:
A keymap is used as a dictionary from an input key (say "q") on a
regular keyboard to a steno key (such as "S-").  Since there may be more
than one key on a keyboard mapped to a single steno key, the dictinary
is generated from a list of steno keys mapped to a list of keyboard
keys.  The keymap is stored in the configuration file as a JSON array.

An ordered dictionary is used for display purposes only, i.e. to list
the key assignments in the same order in the configuration dialogue.

This feature can be used by Dvorak typists who do not normally use a
QWERTY layout, or those with keyboards featuring unusual key positions.

PR #232, opened 2015-02-06
adding preliminary support for the kinesis advantage keyboard

ecraven:
This adds a copy of sidewinder.py which matches on keycodes instead of keystrings (which are a bad idea anyway, as they depend on your keyboard layout...).
However, this does *not* use the thumb keys, but instead uses V/B and N/M. To support the thumb keys, changes in plover internals would be necessary, as they are all special characters (backspace, delete, return and space), which are not handled correctly right now.

### DIFF
--- a/plover/app.py
+++ b/plover/app.py
@@ -23,6 +23,7 @@ import plover.oslayer.keyboardcontrol as keyboardcontrol
 import plover.steno as steno
 import plover.machine.base
 import plover.machine.sidewinder
+import plover.machine.kinesis
 import plover.steno_dictionary as steno_dictionary
 import plover.steno as steno
 import plover.translation as translation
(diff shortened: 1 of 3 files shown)


ecraven:
Just to elaborate, as the original keypresses go through to the client right now (and are later eliminated by backspaces), this obviously doesn't work well for delete and backspace (which cannot be eliminated by backspaces). space and enter often have side-effects in programs (like sending off a html form on a website), so they *should* also be intercepted and only sent to the client when actually intended.
one way to achieve this under linux is to grab the keyboard evdev device, create a new uinput device, and send all keypresses you want the client to get on that device. that would be a rather large change in xkeyboardcontrol.py, but it would also make plover work without Xorg :)

ecraven:
see also: https://github.com/ecraven/plover/tree/evdev-backend

ecraven:
hm.. after playing around with this in earnest for some time, it seems the kinesis does not have good enough NKRO, when pressing 7 or more keys, I see some ghosting. So there's probably no need to actually add separate support for it, unfortunately. However, basing the code in sidewinder.py on keycodes instead of keystrings might be worth thinking about.

morinted:
I really appreciate the thought of moving from keystrings to keycodes. Maybe it could be done such that there's a dictionary of keycodes along with strings to help the developer know what they are doing... ideally, though, long term -- the end user would be able to do it in `plover.cfg` using #224 to customize on runtime. Ideas, ideas. Thanks for the submission, though, shame about Kinesis' limits.

Then again, it's probably the case where most steno learners will upgrade to a machine with serial protocol, but who knows. ¯\_(ツ)_/¯

PR #238, opened 2015-03-06
Plover evdev support

ezyang:
This is an updated and improved version of https://github.com/openstenoproject/plover/pull/60
It supports automatic hotplug, but needs evdev and pyudev to be installed, as well as udev rules. Here are the udev rules I installed in `/etc/udev/rules.d/99-plover.rules`:

```
# Microsoft Sidewinder (BOTH input devices)
SUBSYSTEM=="input", ATTRS{idVendor}=="045e", ATTRS{idProduct}=="0768", GROUP="ezyang", MODE="0660"

# Stenoboard (Arduino Leonardo)
SUBSYSTEM=="input", ATTRS{idVendor}=="1d6b", ATTRS{idProduct}=="0001", GROUP="ezyang", MODE="0660"

# Make /dev/uniput world writeable
KERNEL=="uinput", MODE:="0666"
```

Known bugs:

* Special characters like curly braces and backticks don't work
* No UI for switching between xkeyboard and evdev
* Needs udev rules so that devices show up as evdev
* Only knows how to hook up with SideWinder and Stenoboard (Arduino)

There is some relevant commentary in the old PR which is worth checking out.

### DIFF
--- a/plover/app.py
+++ b/plover/app.py
@@ -211,6 +211,7 @@ class StenoEngine(object):
         self.full_output.send_backspaces = o.send_backspaces
         self.full_output.send_string = o.send_string
         self.full_output.send_key_combination = o.send_key_combination
+        self.full_output.render_key_combination = o.render_key_combination
         self.full_output.send_engine_command = o.send_engine_command
         self.command_only_output.send_engine_command = o.send_engine_command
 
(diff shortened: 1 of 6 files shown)


morinted:
I don't feel as if this is ready for production. The inability to output special characters is rather critical, and the manual tasks that need to be accomplished paired with the sparse support for machines makes it too confusing to document around. I think that it's a great educational reference for when trying to rewrite hooking, but I'm not ready to pull in a change. You can continue with this idea if you can find a way to make it work for the general population, but with the recent XGrabKey change, the difference will probably not be that great.

PR #245, opened 2015-07-29
Adds stroke suggestions window

balthamos:
- Adds stroke suggestions window that updates realtime while writing
- Looks for longest possible phrases in fewest possible strokes and reports
  suggestions to user
    - For example, writing 'of' followed by 'the' will give a suggestion for 'of
      the', since the combined phrase exists in the dictionary
- Allows user to fingerspell a word and receive instant feedback for possible
  strokes
- Additionally, prefixes and suffixes can be fingerspelled, and stroke
  suggestions will be given

screen shot 2015-07-28 at 11 33 51 pm

screen shot 2015-07-28 at 11 36 14 pm

### DIFF
--- a/plover/gui/config.py
+++ b/plover/gui/config.py
@@ -98,7 +98,7 @@ class ConfigurationDialog(wx.Dialog):
         self.dictionary_config = DictionaryConfig(self.engine, self.config, 
                                                   notebook)
         self.logging_config = LoggingConfig(self.config, notebook)
-        self.display_config = DisplayConfig(self.config, notebook)
+        self.display_config = DisplayConfig(self.config, notebook, self.engine)
         self.output_config = OutputConfig(self.config, notebook)
 
         # Adding each tab
@@ -455,7 +455,7 @@ class DisplayConfig(wx.Panel):
     SHOW_SUGGESTIONS_BUTTON_TEXT = "Open stroke suggestions"
     
     """Display configuration graphical user interface."""
-    def __init__(self, config, parent):
+    def __init__(self, config, parent, engine):
         """Create a configuration component based on the given Config.
 
         Arguments:
@@ -467,6 +467,7 @@ class DisplayConfig(wx.Panel):
         """
         wx.Panel.__init__(self, parent, size=CONFIG_PANEL_SIZE)
         self.config = config
+        self.engine = engine
         sizer = wx.BoxSizer(wx.VERTICAL)
         
         show_strokes_button = wx.Button(self, 
--- a/plover/gui/config.py
+++ b/plover/gui/config.py
@@ -20,6 +20,7 @@ from plover.machine.registry import machine_registry
 from plover.exception import InvalidConfigurationError
 from plover.dictionary.loading_manager import manager as dict_manager
 from plover.gui.paper_tape import StrokeDisplayDialog
+from plover.gui.suggestions import SuggestionsDisplayDialog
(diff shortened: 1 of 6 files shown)


plover/gui/config.py
```diff
@@ -476,15 +479,29 @@ def __init__(self, config, parent):
...
     def save(self):
         """Write all parameters to the config."""
         self.config.set_show_stroke_display(self.show_strokes.GetValue())
+        self.config.set_show_suggestions_display(self.show_suggestions.GetValue())
 
     def on_show_strokes(self, event):
         StrokeDisplayDialog.display(self.GetParent(), self.config)
 
+    def on_show_suggestions(self, event):
+        SuggestionsDisplayDialog.display(self.GetParent(), self.config, self.engine)
```
mighele: Hi Balthamos.

I'm getting an error on this line.

'DisplayConfig' object has no attribute 'engine'

Thanks
Mike

balthamos: Hi Mike,

Sorry about that. Looks like that part of the patch got lost somehow. It should be corrected now with 530130c.

Thanks,
Kevin

mighele:
Hi Kevin.

This is really great :)

I just have a couple of things before merging it..

Can you make the window resizable? I ended up writing the stroke for "electrocardiogram" and noticed it didn't quite fit. It would be a nice touch if we could resize it.

I know this might be a real headache but would it be possible for the "last word" to pick up on fingerspelled words. For instance, when a beginner wants to write a word but doesn't know the strokes, they could fingerspell the word and then see on your screen what the strokes are. This would be a great training aid for steno newbies. Let me know what you think.

The changes to the file "osxkeyboardcontrol.py" don't seem to be related to the new feature. Is it code that's been pulled across from some other change?

Thanks again
Mike

balthamos:
Hey Mike,

It should be fairly straightforward to make the window resizable. I just copied the code from the steno stroke window to keep the same look and feel. Probably just a one line change to allow resizing.

Also, the "last word" should already pick up on fingerspelled words. That was one of the primary motivations behind this feature, specifically for the reasons you stated. If it isn't working out for you, let me know.

And yes, the changes in the osxkeyboardcontrol are unrelated to this new feature; they can be safely omitted. They were necessary for me since I was developing with newer Python libraries. I don't know the GitHub-kosher way of pulling these changes out (making a new branch, new pull req, etc), so I'll defer to your knowledge.

Best,
Kevin

lifton:
Seems like you should simply be able to omit commit 14a471d9bbbaa01077088b259b9ecb9e84a3a5ca. There are several ways of doing this. For example, you could rewind your branch, cherry-pick the remaining two commits, and force push to the same branch, which should automatically update this pull request. You could also simply create a new pull request from a new branch.

balthamos:
Thanks for the suggestions lifton. Making a new branch/PR is probably the cleanest/safest option. Force pushing is rather dangerous and I've seen it make a mess before, though it probably wouldn't cause any issues in this scenario (famous last words). I could also just reset that single commit, but it would make the history unnecessarily noisy.

I'll go ahead and close this PR and open another one on a clean branch with the required commits, plus the window resizing changes Mike was asking about.

balthamos:
New PR is: Adds stroke suggestions window #249

lifton:
@balthamos thanks for revising. I'll leave the new pull request to @mighele since he already reviewed it once and is probably more familiar with this part of the code base anyway.

PR #248, opened 2015-08-14
Respect install prefix for data files

Twey:
Use a relative path for `data_files` in `setup.py` so as to respect prefix.  This doesn't seem to break anything.
(without this change the package will not build on NixOS and other distros that don't have or use `/usr/share`)


balshetzer:
With this change does "python setup.py install" work on both distros that do and don't have /use/share?

Twey:
It should do, according to the Python docs:
>  If directory is a relative path, it is interpreted relative to the installation prefix (Python’s `sys.prefix` for pure-Python packages, `sys.exec_prefix` for packages that contain extension modules).
— https://docs.python.org/2/distutils/setupscript.html

So if Python is installed in `/usr`, as is generally the case, this should work just the same as without the patch.  I haven't tested it, though.

morinted:
This has been fixed thanks to Benoit-Pierre's great development environment rewrite.

PR #249, opened 2015-08-19
Adds stroke suggestions window

balthamos:
This is a redo of the following pull request: Adds stroke suggestions window #245

This removes an unnecessary commit and adds a window resizing to the new window.


mighele:
Hi Kevin.

Thanks for resubmitting the pull.

I'm still having a few problems with the "last word" not showing what I would expect. Here are my findings:

Using the following dictionary only:
{
"-Z": "{^s}",
"K*": "{&k}",
"S*": "{&s}",
"TKAEU": "day"
}

Write "K*"
Expect last word "k"
Actual is "{&k}"

Write "S*"
Expect last word "s" or "ks" if written after above case
Actual is "{^s}"

Write "TKAEU"
Expect last word "day"
Yes, this works
Now write "-Z"
Expect last word "days"
Actual is "{^s}"

Normal words and phrases work perfectly

mighele:
It might be worth noting that plover builds the words up for you both for fingerspelling and for suffixes. If you look in your function "show_stroke", the object called "new" contains an array of objects. The "word" property on these objects will contain a built up word like "ks" and "days" in my previous examples. However, using this exclusively stops the phrases from working. I.e. Using "this is" just gives "is" as that was the last whole word.

Achim63:
I don't quite get your example - you don't even have a "K*" entry in your dictionary but expect it to procude the word "k"?
Using the standard dict.json, your examples work as I'd expect them to ( "K*/S" or "K*/S*" produce "ks", "TKAEU/-Z" produces "days").

mighele:
Hi Achim.

I've just tested it with the standard dictionary from here:
https://github.com/openstenoproject/plover/blob/master/plover/assets/dict_unsorted.json

"TKAEU/-S" now works as "days", but only because there is an entry in the dictionary specifically for "TKAEU/-S" (two strokes). If the dictionary only contains a single stroke version "TKAEUS", the problem still occurs. I maintain that this is still not ideal functionality, but it's only minor.

Also using the standard dictionary, my fingerspelling example doesn't work.

Fingerspelling "ks" just shows last word "s"
Fingerspelling "day" (TK\*/A\*/KWR\*) does not show last word "day" but "y"

Achim63:
Strange - for me fingerspelling works fine using the 2.5.8 release version on OSX.
If I stroke a word that has no specific entry for the plural, appending "s" also works as expected.

mighele:
Hi Achim.

I just want to confirm that it's the "last word" line at the top of the suggestions window that I'm referring to. The actual output from Plover is working perfectly with fingerspelling etc..

The way I'm testing this is by downloading the current Plover master branch and applying the changes from this pull request to that branch. I'm not using the release version of Plover.

Achim63:
Sorry, we talked about different things then. I cannot try the master branch since I didn't manage yet to build it on OSX (but then, I didn't try really hard to make it work).

balthamos:
Mike, most of your observed behavior is correct/intended. The suggestions window will only lookup things in your dictionary and show how to stroke them efficiently, if possible. If you get a dictionary miss, the only "suggestion" it could possibly give is what you just stroked, which isn't super useful. I can see the merit of showing what you stroked, but I like the 1:1 correspondence between the suggestions window and your dictionary, as there isn't any question of what is/isn't in your dictionary.

In your test with the reduced dictionary stroking "TKAEU/-S", last word should be "days" with "No suggestions" in the window, not "{&s}". Same with your attempt at fingerspelling "day". Last word should be: "d", "da", "day". One reason you may be seeing different results than me is that I am developing on Mac OS X. I thought this feature was placed in a platform-independent part of the code, but perhaps it isn't?

On a final note, you may see decorated last words, like "{a^}", depending on the situation. This is intended to alert the user that the suggested strokes are for the prefix "a", not the word "a". However, the only time you should ever see suggestions for a prefix/suffix is when that is the only possible dictionary hit, so my example of "{a^}" and "a" is impossible with the default dictionary. Words are given priority over prefixes/suffixes during the lookup.

mighele:
In my reduced dictionary example: If I add one more entry of "TKAEUZ": "days" and then stroke "TKAEU/-Z". I will have just written the word "days" with two strokes however there is a dictionary entry to stroke it in one. I would expect this to show in the suggestions as a way of stroking it more efficiently.

The fingerspelling definitely doesn't work for me on a windows 8.1 machine.

balthamos:
That is exactly what should happen. If your dictionary consists only of "TKAEUZ" and you stroke "TKAEU/-Z", last word should be "days" and the only suggestion displayed should be "TKAEUZ".

I'm betting that a core assumption isn't working on Windows (or at least your system), which is why the word building and fingerspelling are not working for you. I'll try to locate a Windows box for debug/confirmation.

mighele:
*"If your dictionary consists only of "TKAEUZ" and you stroke "TKAEU/-Z", last word should be "days" and the only suggestion displayed should be "TKAEUZ"."*

It doesn't show this suggestion for me though. Last word shows as "{^s}" with a suggestion of "-Z".

balshetzer:
Perhaps someone could write unit test that replicates the problem.

balthamos:
Ok, I got Plover up and running in Windows and the suggestions window works as I expect it to. I don't know what is different about your setup, Mike.

mighele:
I have created some unit tests to illustrate my problem but because the logic for the suggestions is inside the GUI and not separated out, the tests aren't particularly clean. The file can be found here:
https://github.com/mighele/plover/blob/patch-1/plover/gui/test_suggestions.py

I think the problem stems from the the "phrase" variable that is passed into "lookup_suggestions".

For me, when I stroke "day" and then "{^s}", the "phrase" variable is "day s" (with a space). This is because of the preceeding line:
phrase = ' '.join(phrase)

When I fingerspell the word day, the "phrase" variable is "TK * A * KWR *" (notice the spaces between the letters and the asterisks).

Just to clarify what I'm doing to reproduce my "day" / "^s" problem:
I open this branch in the github website: https://github.com/balthamos/plover/tree/suggestions-window
I click "Download ZIP".
I extract the zip to my C drive.
I open the folder and create a launch.py file.
I run the launch.py file
Plover opens
The dictionary I have loaded is:
{
"TKAEU": "day",
"-Z": "{^s}",
"TKAEUZ": "days"
}
I activate Plover.
I stroke "TKAEU"
Plover writes "day" to my text editor
Plover stroke suggestion window shows last word "day"
I stroke "-Z"
Plover changes "day" to "days" in my text editor
Plover stroke suggestion window shows last word "{^s}"
Plover stroke suggestion window does not show last word "days" as I would expect.

Are we definitely using the same code base? Could it be that you have some local changes that haven't been committed to the "balthamos:suggestions-window" branch? Just trying to work out why we're getting different results. Let me know if yours still works even after following my steps above.

Hopefully we'll get to the bottom of this elusive problem soon :)
Mike

balthamos:
There is definitely something amiss here. Just looking through your unit test, what you are feeding to the suggestion engine is not what I see.

```python
        # This is how phrase looks when I stroke day and then stroke {^s}
        phrase = 'day s'

        # This is how phrase looks when I stroke {^d} {^a} {^y}
        phrase = 'TK * A * KWR *'
```

To remove any possibility of floating changes, I downloaded the same zip and walked through the same steps. I am unable to recreate the issue you are seeing.

I inserted some print statements after the `lookup_suggestions` call in `plover/gui/suggestions.py`:
```diff
diff --git a/plover/gui/suggestions.py b/plover/gui/suggestions.py
index a4d131a..1b72e58 100644
--- a/plover/gui/suggestions.py
+++ b/plover/gui/suggestions.py
@@ -121,9 +121,11 @@ class SuggestionsDisplayDialog(wx.Dialog):
         split_words = PAT.findall(self.words)
         interp_phrase = split_words[-1:]
         self.listbox.Clear()
+        print 'Starting loop for split_words:%s' % (split_words)
         for phrase in SuggestionsDisplayDialog.tails(split_words):
             phrase = ' '.join(phrase)
             found, interp_phrase, suggestions = self.lookup_suggestions(phrase)
+            print '  [%4s] phrase:"%s" interp:"%s" suggs:%s' % ('STOP' if found else 'CONT', phrase, interp_phrase, suggestions)
             if found:
                 # Limit arbitrarily to 10 suggestions
                 for suggestion in suggestions[:10]:
```

Here are the strokes I performed:
```
TH      S-    AEU   TKAEU/-Z   TK*  A*  KWR* S*
This    is     a     days       d   a    y   s
```

And this is the output:
```
Starting loop for split_words:['this']
  [STOP] phrase:"this" interp:"this" suggs:[('TH',), ('TH-RBGS',)]
Starting loop for split_words:['this', 'is']
  [STOP] phrase:"this is" interp:"this is" suggs:[('TH-S',)]
Starting loop for split_words:['this', 'is', 'a']
  [CONT] phrase:"this is a" interp:"this is a" suggs:[]
  [STOP] phrase:"is a" interp:"is a" suggs:[('SA*EU',)]
Starting loop for split_words:['this', 'is', 'a', 'day']
  [CONT] phrase:"this is a day" interp:"this is a day" suggs:[]
  [CONT] phrase:"is a day" interp:"is a day" suggs:[]
  [CONT] phrase:"a day" interp:"a day" suggs:[]
  [STOP] phrase:"day" interp:"day" suggs:[('TAEU',), ('TKAEU',)]
Starting loop for split_words:['this', 'is', 'a', 'days']
  [CONT] phrase:"this is a days" interp:"this is a days" suggs:[]
  [CONT] phrase:"is a days" interp:"is a days" suggs:[]
  [CONT] phrase:"a days" interp:"a days" suggs:[]
  [STOP] phrase:"days" interp:"days" suggs:[('TKAEUS',), ('TKAEU', '-S')]
Starting loop for split_words:['this', 'is', 'a', 'days', 'd']
  [CONT] phrase:"this is a days d" interp:"this is a days d" suggs:[]
  [CONT] phrase:"is a days d" interp:"is a days d" suggs:[]
  [CONT] phrase:"a days d" interp:"a days d" suggs:[]
  [CONT] phrase:"days d" interp:"days d" suggs:[]
  [STOP] phrase:"d" interp:"d" suggs:[('SP-S', 'TK*'), ('TK*', 'SP-S')]
Starting loop for split_words:['this', 'is', 'a', 'days', 'da']
  [CONT] phrase:"this is a days da" interp:"this is a days da" suggs:[]
  [CONT] phrase:"is a days da" interp:"is a days da" suggs:[]
  [CONT] phrase:"a days da" interp:"a days da" suggs:[]
  [CONT] phrase:"days da" interp:"days da" suggs:[]
  [STOP] phrase:"da" interp:"da" suggs:[('TKA',)]
Starting loop for split_words:['this', 'is', 'a', 'days', 'day']
  [CONT] phrase:"this is a days day" interp:"this is a days day" suggs:[]
  [CONT] phrase:"is a days day" interp:"is a days day" suggs:[]
  [CONT] phrase:"a days day" interp:"a days day" suggs:[]
  [CONT] phrase:"days day" interp:"days day" suggs:[]
  [STOP] phrase:"day" interp:"day" suggs:[('TAEU',), ('TKAEU',)]
Starting loop for split_words:['this', 'is', 'a', 'days', 'days']
  [CONT] phrase:"this is a days days" interp:"this is a days days" suggs:[]
  [CONT] phrase:"is a days days" interp:"is a days days" suggs:[]
  [CONT] phrase:"a days days" interp:"a days days" suggs:[]
  [CONT] phrase:"days days" interp:"days days" suggs:[]
  [STOP] phrase:"days" interp:"days" suggs:[('TKAEUS',), ('TKAEU', '-S')]
```

As you can see, my "phrase" variable has already been translated and is ready for output. You can also see how fingerspelling constructs the word on the fly, rather than giving steno strokes.

The next steps would be to trace it back and find where the discrepancy is coming from. It's looking like it's coming from the engine or translator... This has the signature of a platform-dependent problem, but I tested this myself in Windows 10 from an entirely clean setup and had no issues, so it's not that simple. Maybe it's a configuration/environment thing.

We'll figure it out soon enough :)

mighele:
Hi Kevin, I hope you had a good weekend.

I've figured it out!  The reason it's not working for me is because I have my "Space Placement" set to "After Output".

Go into Plover Configuration
Go to the Output tab
Change your Space Placement to After Output.

Now try the new features.

I've just tested it with Space Placement to "Before Output" and it all works beautifully.

If you can get it working with "After Output" I'll be more than happy to merge this pull request.

Thanks
Mike

balthamos:
Glad to hear it was something simple, Mike. I am able to reproduce the bad behavior by changing the "Space Placement" parameter.

To support both options, I'll need to study the code and find a better approach. The original implementation was trying to grab the plain, interpreted text before it was munged and fed through the OS layers, but apparently it was grabbing from the wrong layer.

At a glance, the space placement code look rather unfriendly, so hopefully I won't need to touch that... It may be worth code reviewing that piece to make it more maintainable.

mighele:
I would highly recommend trying to use the "word" property for this.

For instance, in show_stroke where you get the parameters "old" and "new", they will contain a "word" property which is perfect for what you're doing. It will already glue together fingerspelled words and suffixes / prefixes. The space after / space before code makes sure that "word" will always be correct regardless of where the space was put. There are lots of unit tests that ensure the "word" is correct for all cases.

I would look into using that if possible. For the cases where you've written a phrase, the word property won't work but the text property will be longer than the word property so perhaps use text in that case.

Would you like me to try and put an example together?

balthamos:
Thanks for the suggestion, Mike. If you could point me in the right direction with a snippet or some pseudocode, that would be a big help. You know the design better than me :)

balthamos:
Reviewed your patch and it looks good. Changes are merged and ready for upstream.

Achim63:
I played a while with this today and I have one major gripe with it: it doesn't work for the first word(s) in a sentence - i.e. it's case dependent.

morinted:
@Achim63 This is already merged, so you can open an issue if you think it's a feature you'd like. I think that case dependent is all right because sometimes you want to check if, for example, "Covenant" is defined as opposed to "covenant". But I don't imagine it would be hard to remove the case dependence.

balthamos:
The reason case dependence is in there is to give suggestions for proper nouns (eg. Mary vs marry). Though, the logic can probably be modified to fall back on all lower case in the event the mixed case word cannot be found, giving the best of both worlds.

Achim63:
The fallback solution sounds good to me!

PR #250, opened 2015-08-19
Unit test cleanup

mighele:
Various changes to fix all broken unit tests.


plover/dictionary/test_default_dict.py
```diff
@@ -4,7 +4,7 @@
 import json
 import unittest
 
-DICT_PATH='plover/assets/dict.json'
+DICT_PATH='plover/assets/dict_unsorted.json'
```
balshetzer: This doesn't really fix the test. 

The default dictionary now consists of all the .json files combined. The test would need to load all of them rather than just dict_unsorted.json.

plover/assets/dict_unsorted.json
```diff
@@ -121866,7 +121866,6 @@
 "KWRA/AUL": "y'all",
 "KWRAOU/AUL": "y'all",
 "KWRAB/PWA/TKAB/PWA": "yabba dabba",
-"KWRAT": "yacht",
```
balshetzer: Why remove these entries? Are these changes related to fixing unit tests?

plover/dictionary/test_loading_manager.py
```diff
@@ -25,8 +25,6 @@ def __call__(self, filename):
         loader = MockLoader(files)
         with patch('plover.dictionary.loading_manager.load_dictionary', loader):
             manager = loading_manager.DictionaryLoadingManager()
-            manager.start_loading('a')
-            manager.start_loading('b')
```
balshetzer: This line tested that we only loaded file 'b' once even though we tried to load it twice.

plover/dictionary/test_loading_manager.py
```diff
@@ -25,8 +25,6 @@ def __call__(self, filename):
         loader = MockLoader(files)
         with patch('plover.dictionary.loading_manager.load_dictionary', loader):
             manager = loading_manager.DictionaryLoadingManager()
-            manager.start_loading('a')
```
balshetzer: This line tests that the start_loading function works on its own. The other dictionaries are loaded with the load function.

plover/assets/dict_unsorted.json
Achim63: Please don't delete those - that's exactly how I'd try to write "yacht" so it's a perfectly valid entry. And who doesn't like the Flintstones?

mighele: The reason I've removed KWRAT = yacht is because it conflicts with an older entry for "{^@^}". Since the latter entry was in the dictionary long before the "yacht" entry I decided to remove the yacht one.

plover/dictionary/test_loading_manager.py
mighele: Ah ok. The start_loading function gets called later on in the normal way so I assumed this was redundant code. I'll re-instate.

plover/dictionary/test_default_dict.py
mighele: True. I just figured it was worth a quick fix so at least the tests don't error.

plover/assets/dict_unsorted.json
Achim63: There's no entry for "@" in my release version of the dictionaries ... I had to add it to my personal dictionary. Might be a good idea to add it, though, maybe use "KRWA_Asterisk_T" to leave yacht as is?

mighele: I've just updated this to KWRA\*T.

Issue #174, opened 2014-01-13
Invisible Dictionary Update Window


stenoknight:
It's distracting for captioning clients to have to look at the dictionary update popup whenever the captioner adds an entry. If there were an option to make it transparent or invisible, so that all the definitions were done without any visual feedback, that would be perfect. Failing that, having the option to make the pop-up much smaller and/or less opaque would be a start. This is obviously an advanced feature, not intended for beginners, so it would be disabled by default.

On the one hand, I'm probably one of the few people currently using Plover to provide daily onsite realtime captioning on a professional basis. (Yep, as of last week I've switched over completely to Plover from my proprietary software at my day job, and I don't intend to go back.) On the other hand, I anticipate that our numbers won't stay small forever. So while this feature would be extremely useful for me, it might not be high priority right now unless it's fairly simple to implement. In order to become truly professional captioning software, though, Plover will need this feature at some point.

jeremy-w:
The suggestions and steno tape windows could also usefully be configurably transparent, similar to how Terminal.app has an alpha slider for its window style. I like to pin suggestions to always on top so I can peep over to see suggestions seamlessly, but sometimes I need to see what's behind them on a laptop screen, and not having to drag them out of the way would be swell.

PR #251, opened 2015-08-24
Fix issue #174: Add Translation Transparency

mighele:
To fix issue https://github.com/openstenoproject/plover/issues/174

Added functionality to set the transparency of the Add Translations window.


mighele:
I'm going to close this request and re-create it so that there are no merge conflicts

PR #252, opened 2015-08-27
Add Translation Transparency

mighele:
Wishlist Feature:
Issue https://github.com/openstenoproject/plover/issues/174

Added functionality to set the transparency of the Add Translations window.

### DIFF
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -24,6 +24,10 @@ class AddTranslationDialog(wx.Dialog):
                            pos, wx.DefaultSize, 
                            wx.DEFAULT_DIALOG_STYLE, wx.DialogNameStr)
 
+        transparency = config.get_translation_frame_transparency()
+
+        self.SetTransparent(255 - (2.55 * transparency))
+
         self.config = config
 
         # components
@@ -48,7 +52,7 @@ class AddTranslationDialog(wx.Dialog):
         sizer.Add(label, 
                   flag=wx.TOP | wx.RIGHT | wx.BOTTOM | wx.ALIGN_CENTER_VERTICAL, 
                   border=self.BORDER)
-        sizer.Add(self.translation_text          , 
+        sizer.Add(self.translation_text,
                   flag=wx.TOP | wx.RIGHT | wx.BOTTOM | wx.ALIGN_CENTER_VERTICAL, 
                   border=self.BORDER)
         sizer.Add(button, 
--- a/plover/gui/config.py
+++ b/plover/gui/config.py
@@ -44,6 +44,7 @@ SPACE_PLACEMENTS_LABEL = "Space Placement:"
 SPACE_PLACEMENT_BEFORE = "Before Output"
 SPACE_PLACEMENT_AFTER = "After Output"
 SPACE_PLACEMENTS = [SPACE_PLACEMENT_BEFORE, SPACE_PLACEMENT_AFTER]
+TRANSLATION_TRANSPARENCY_LABEL = "Add Translations Dialog Transparency"
 CONFIG_PANEL_SIZE = (-1, -1)
 UI_BORDER = 4
 COMPONENT_SPACE = 3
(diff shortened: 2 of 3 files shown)


chajadan:
Hey Mike! I have several comments to make so I'll let you know when I'm fully done reviewing. The good part is, it works! So the rest is just details.

chajadan:
The file was largely wrapping lines at 80 characters. Do you mind keeping that convention in place? (I tend to find it an outdated convention, but it's still prevalent, and I tend to conform to it.)

plover/gui/add_translation.py
```diff
@@ -24,6 +24,10 @@ def __init__(self, parent, engine, config):
                            pos, wx.DefaultSize, 
                            wx.DEFAULT_DIALOG_STYLE, wx.DialogNameStr)
 
+        transparency = config.get_translation_frame_transparency()
+
+        self.SetTransparent(255 - (2.55 * transparency))
```
chajadan: SetTransparent takes an int but we'll be passing in a float this way. Can we make sure to explicitly pass an int?

A very minor note: just int'ing only yields [0, 254] on my system but rounding gives the full [0, 255]:

self.SetTransparent(255 - int(round(2.55 * transparency)))

plover/gui/config.py
```diff
@@ -44,6 +44,7 @@
 SPACE_PLACEMENT_BEFORE = "Before Output"
 SPACE_PLACEMENT_AFTER = "After Output"
 SPACE_PLACEMENTS = [SPACE_PLACEMENT_BEFORE, SPACE_PLACEMENT_AFTER]
+TRANSLATION_TRANSPARENCY_LABEL = "Add Translations Dialog Transparency"
```
chajadan: I'm not a fan of this label, because it reads funny. I would prefer "Translation Dialog Transparency", after a bit of thought. What do you think?

plover/gui/config.py
```diff
@@ -469,26 +470,59 @@ def __init__(self, config, parent, engine):
...
+        translation_transparency_sizer = wx.BoxSizer(wx.VERTICAL)
+
+        translation_transparency_label = wx.StaticText(self, label=TRANSLATION_TRANSPARENCY_LABEL)
+        translation_transparency_sizer.Add(translation_transparency_label,
+                border=COMPONENT_SPACE,
+                flag=wx.ALL)
+
+        translation_transparency_slider_sizer = wx.BoxSizer(wx.HORIZONTAL)
+
+        self.translation_transparency_slider = wx.Slider(self)
```
chajadan: SetMin and SetMax are superfluous here, since they are the defaults. Also, the constructor would take these values, for more compact code.

plover/gui/config.py
```diff
@@ -469,26 +470,59 @@ def __init__(self, config, parent, engine):
...
+        self.translation_transparency_slider.SetMin(0)
+        self.translation_transparency_slider.SetMax(100)
+        self.translation_transparency_slider.SetTick(1)
```
chajadan: This doesn't do what I think you think it does. It puts a tick mark along the bar, not determine the minimum jump from one point to another. It also only functions, on windows at least, if the SL_AUTOTICKS style is provided.

plover/gui/config.py
```diff
@@ -469,26 +470,59 @@ def __init__(self, config, parent, engine):
...
+        self.translation_transparency_slider.SetValue(self.config.get_translation_frame_transparency())
+        self.translation_transparency_slider.Bind(wx.EVT_SLIDER, self.on_transparency_slider_move)
+
+        translation_transparency_slider_sizer.Add(self.translation_transparency_slider)
+
+        self.translation_transparency_value = wx.StaticText(self,
```
chajadan: This looks better with a percent sign. Also would be necessary on the slider move update.

chajadan:
I'm not a fan of the dialog spacing. I've attached before and after shots of the dialog. Before, it was, to the eye, clearly two parts. The new third blends in with the one above. Some extra space would help.

image

image

chajadan:
The sizer code got a lot more involved, and I don't see any purpose behind the extra lines. One sizer for all six elements makes sense to me. I could understand using sizers to facilitate spacing, but you even broken the existing part into separate sizers. :-p Maybe you're like me and ~love~ a container store!!

chajadan:
@mighele I'm done reviewing. =) Could have said that before but forgot after I got back from the gym.
Thanks for contributing! If you disagree with any of my points please feel free to share your views.

morinted:
Hey @mighele, I'd still really love to see this change in Plover. Will you be working on it again to get it up for submission?

jeremy-w:
Testing this after rebasing on master. Issues:

- [ ] Add translation window defaults to 0% alpha. This will confuse the heck out of someone new to the app! If the config value is not set (as in case of upgrade), it should default to 100%.
- [ ] Lack of live-application of the alpha slider makes it hard to dial in an alpha you're happy with as a user, if you want anything between 0% and 100%.
    - This surprised me as a Mac user, since the standard on that platform is for settings to affect things directly on change, not on OK. In the event there's an Apply/Cancel dialog, it just gives you a chance to back out the change before committing permanently, but the effects are expected always to be visible on-the-fly.

The live-application one I don't see as blocking merger, the 0% default might just be a small presentation issue of how the slider is set up, since my window didn't show up with 0%. :)

jeremy-w:
Hacking on this, need to create a new PR.

jeremy-w:
Ah, I didn't realize till now that it's legitimately setting transparency rather than alpha. I'm used to thinking in terms of opacity, I guess. So that default issue is fine.

PR #254, opened 2015-09-12
Allow Windows keys behavior, and other passthrough keys enhancements.

chajadan:
This pull request contains two commits.

The first is a simplification and correction, whereby meta keys are monitored as a group, rather than individually (since no behavior is determined based on the particular key in question), and also now their down state is properly monitored. See commit message for details.

The second commit adds Windows key commands to the list of actions that will act as usual and be ignored by plover.

### DIFF
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -45,7 +45,8 @@ class KeyboardCapture():
     CONTROL_KEYS = set(('Lcontrol', 'Rcontrol'))
     SHIFT_KEYS = set(('Lshift', 'Rshift'))
     ALT_KEYS = set(('Lmenu', 'Rmenu'))
-    PASSTHROUGH_KEYS = CONTROL_KEYS | SHIFT_KEYS | ALT_KEYS
+    WIN_KEYS = set(('Lwin', 'Rwin'))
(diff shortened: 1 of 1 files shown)


morinted:
I will pull this soon. The code is very sane and I'll just build it on my machine, do some stress testing and confirm. Hopefully by Monday evening.

morinted:
My Python install is messed up on Windows and I wasn't able to do this tonight.

morinted:
@chajadan I'm finding that I can break this. Try hitting both shifts, a control, and pressing tab. Removing all at once, Plover now isn't working on input. Can you confirm this and propose a fix?

chajadan:
I can reproduce this as you described. I looked into it a bit more and it
turns out you can do it with a just a single shift key. Just hold it down.
To me it seems as though obviously the key is repeating down keys, which
are accrued in my counter, and then only registering a single up key. I
thought I could get away with a simple counter, but nope. And  pyhook
doesn't pass out the flag about repeat strokes I hear may be available.

I'll update my pull request to something that address this. Good find!
Thanks!

chajadan:
@morinted Change completed. I switched to a set to ignore the repeat keys

morinted:
Looks good :) Thanks.

PR #259, opened 2015-09-25
Fix/improve Linux keyboard supression

benoit-pierre:
- make it consistent with how the Windows/OS X versions work: suppress the whole alpha rows (including previously unsuppressed keys, e.g. 'z').
- use grabs on the suppressed keys (only when no modifier except for numlock is pressed).


morinted:
Thanks Ben, I think this is a great incremental code change that will help Linux users. May I ask what your Linux set up is? I'll test it on something different. For clarity's sake, we can continue discussion of the issue in #41. I like how you copied the behavior from the other operating systems. I'll review the code and do some testing by Monday.

benoit-pierre:
I'm on Arch Linux, and my daily driver is a Nopoo Choc Mini (with either red or brown Cherry MX switches).

morinted:
Ben, does your implementation face the same issue as mentioned in #219?

benoit-pierre:
No, since the whole keyboard is not grabbed (XGrabKey versus XGrabKeyboard).

morinted:
I did some testing -- it works really, really well. This is so awesome to see. Unicode is the next issue to tackle, but this is super duper cool regardless. Thanks so much for the contribution!

PR #260, opened 2015-09-25
Change arrow key dictionary entries to be more like traditional layout

morinted:
- Swap behavior of B and G, so now the layout acts like a regular arrow pad:

```
 P  =  u
RBG = ldr
```

- Add {^} to the end of the movement strokes to prevent accidental spaces


morinted:
@stenoknight and @chajadan, can I get your blessing on this pull request (you can merge it, too)

chajadan:
Makes sense -- no doubt it crossed my mind the first times I used those
arrow keys too.

PR #265, opened 2015-09-26
Dictionary update (huge)

morinted:
- Remove NULL, FLUSH, and some nonsense entries
- Add in bulk of MKKexport and GoogleDoc dictionaries from Mirabai
- Create commands dictionary with some new commands (including toggle, on, off for plover)
- PHROFR is now "Plover"... finally!
- Delores also makes an appearance
- Add my own (Ted's) personal entries -- many tech companies and websites -- Tumblr, reddit -- also many phonetic definitions of words that are too "brief"-y for beginners (I learned while building this dictionary)

Warnings:
- I tried to remove as many medical terms as I could, but there are *so many*. Thousands. I removed medical terms that were 1 stroke and 2 or more words, but anything else I can't really guarantee. As far as I can tell, it shouldn't make a different to people who don't try to stroke them. Opinions?


morinted:
Also adds Stenosaurus and StenoTray and Stenoboard :)

morinted:
This breaks unit tests. Ted, please fix this and we'll reevaluate.

morinted:
I've been using this for the past couple days and I've got to say it's a nice improvement over the previous dictionary. I'm also pretty sure that as of today I've cleaned out all the junk strokes. In addition, there's copy and paste for both Windows and Mac.

Issue #264, opened 2015-09-26
Suggestion window doesn't work for first word in a sentence


Achim63:
Since the first word in a sentence is always capitalized, the suggestion window doesn't show any results for lower case words, so it stays empty most of the time in this case.
I'd expect the search to be case independent, maybe marking the result somehow if it's not an exact match (attach e.g. " (no caps)" or some such to the stroke definition).

morinted:
@balthamos would you be able to make a fix for this? I'm not sure where you can fit it in the UI. I think the user should be indicated *somehow* that the case doesn't match. But definitely would be useful in case you have a stroke such as "Android" and you get the suggestion for "Android" after you type "android".

Maybe it could be like this:

User types cat. Stroke suggestions:

```
cat:
KAT
KAET
Cat:
KA*T
CAT:
KA*T/KA*T
```

Thoughts?

Achim63:
That looks fine - similar to what I do in a bash script to look up entries.

balthamos:
As you pointed out, the current UI is insufficient for displaying multiple suggestions. It could probably be changed to mimic the list you wrote out, with maybe the words bolded to make the sections more distinct and easier to read.

Have any other ideas for how to display this information cleanly?

morinted:
Well, if you implement it such that the information is returned in a list
of lists (assuming that you would have perfect match first, then ordered by
parity difference or position of caps?), then we can always change the UI
later. I'm alright with just providing bold rows with colons for the
definition (maybe in a regular sans serif font) then fixed-width stroke
definitions as it is now. Maybe a *slightly* lighter gray for the non-exact
matches. But we can tweak all this to our liking after trying it out :)

balthamos:
Yea, I'm not particularly concerned about the backend logic; it should be a fairly straightforward implementation. Just thought I'd shake the bushes for any good ideas in the UI department. Measure twice, cut once kind of thing :)

morinted:
Cool. Yeah, I'd just say something along the lines of:

**cat:**
`KAT`
**CAT:**
`KA*T`

Yeah, and with the non-exact matches with a dark grey instead of pure
black. Then maybe take a screenshot and share it here and everyone can
comment.

stenoknight:
A bonus feature might be a basic parser that looks for the phonetic
definition based on a simple steno phonetics table + English phonetic
dictionary match and then highlights that entry with another color or
boldface or something. Highlighting the phonetic entry automatically would
[...]

PR #267, opened 2015-10-01
Adds multiple suggestions to Suggestions window

balthamos:
See #264 for more details.

### DIFF
--- a/plover/gui/suggestions.py
+++ b/plover/gui/suggestions.py
@@ -13,7 +13,7 @@ ON_TOP_TEXT = "Always on top"
 UI_BORDER = 4
 MAX_STROKE_LINES = 38
 LAST_WORD_TEXT = 'Last Word: %s'
-DEFAULT_LAST_WORD = '[]'
+DEFAULT_LAST_WORD = 'N/A'
 
 class SuggestionsDisplayDialog(wx.Dialog):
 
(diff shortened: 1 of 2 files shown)


morinted:
When there are no suggestions, it displays last word as "N/A". It would be nice to see the word that it can't match. Because if my dictionary doesn't have FlowType in it, I'd like to see "FlowType: no suggestions" instead of "N/A: no suggestions". This will help users decide when it's time to add a new dictionary entry.

balthamos:
I went ahead and changed "Last Word" to always show the last stroked word.

Ellipses do work for me, though it only suggests HR-PZ (and also "last word" displays "." instead of "..."). 

The reason that SKWR-RBGS does not show up is because it is defined as "{^...}{ }{-|}", which is too complicated to be interpreted by the current lookup logic.

morinted:
I'll build and use it during the day tomorrow before pulling. I've got to
say, it's really awesome. Changes the way I feel about the default
dictionary. Thanks so much for these changes.

morinted:
Nitpicking -- could you reword your commit message on 0d5e8b5b50 to say "264" instead of "246"? Building now and I'll pull tonight.

balthamos:
Glad I was able to contribute something useful :). Commit message should now reference the correct issue.

Also, the reason it is picking up HR-PZ but not HR-PS is because HR-PS is `"..."` and HR-PZ is `". . ."` (has spaces between the dots).

The current word detection logic uses this regular expression `'[-\'"\w]+|[^\w\s]'` to figure out what is a word and what isn't. It does not consider `.` to a part of the same word, always separate. So `...` is never constructed as a "word" by the suggestion logic.

morinted:
Right, I get what you mean. All right, well, thanks so much for these changes. I'll probably get to styling sometime soon with an HTML list box.

Issue #105, opened 2013-07-20
'*' key not working properly as backspace


caru:
OS: Windows7 64bit
Plover 2.4.1

Steps:
1. write a dozen of words
2. delete all words with '*' key.

Issue description:
Step 2 cannot be completed, it deletes just a few words, than the '*' stops working, ie: it doesn't fire even a single backspace anymore, when translation history stops to work as can be seen in log tail, attached below.

This issue is critical for users of Plover without a QWERTY.

EDIT:  Actually, one can just define a backspace shortcut probably, so it is not so critical as I initially wrote...

plover.log:
2013-07-20 04:20:46,084 Stroke(-R R-)
2013-07-20 04:20:46,084 Translation(('R-R',) : {#Return})
2013-07-20 04:20:46,316 Stroke(-P -B)
2013-07-20 04:20:46,316 Translation(('-PB',) : an)
2013-07-20 04:20:46,729 Stroke(-L)
2013-07-20 04:20:46,729 Translation(('-L',) : None)
2013-07-20 04:20:47,052 Stroke(P- -L)
2013-07-20 04:20:47,055 Translation(('P-L',) : None)
2013-07-20 04:20:47,845 Stroke(-B -R P- -L T-)
2013-07-20 04:20:47,845 Translation(('TP-RBL',) : None)
2013-07-20 04:20:49,273 Stroke(P- -B -R -L -T T-)
2013-07-20 04:20:49,273 Translation(('TP-RBLT',) : None)
2013-07-20 04:20:49,898 Stroke(-T P- T-)
2013-07-20 04:20:49,898 Translation(('TP-T',) : if the)
2013-07-20 04:20:50,891 Stroke(-T -R P- T- -L)
2013-07-20 04:20:50,891 Translation(('TP-RLT',) : None)
2013-07-20 04:20:51,996 Stroke(-T -R P- T- -L)
2013-07-20 04:20:51,996 Translation(('TP-RLT',) : None)
2013-07-20 04:20:52,753 Stroke(-T -R P- -L T-)
2013-07-20 04:20:52,753 Translation(('TP-RLT',) : None)
2013-07-20 04:20:53,801 Stroke(-T -R P- -L T-)
2013-07-20 04:20:53,802 Translation(('TP-RLT',) : None)
2013-07-20 04:20:54,081 Stroke(-R)
2013-07-20 04:20:54,082 Translation(('-R',) : are)
[...]

PR #268, opened 2015-10-01
Add word deletion on empty buffer

morinted:
Addressing issue #105 

Adds an in-built way to handle the empty buffer. After there are no strokes in the buffer, Plover emits "Alt_L(BackSpace)" (Mac) or "Control_L(BackSpace)" (Win & Linux) in order to not halt the user. Also had to include the "{^}" operator because this kind of spacing removes the previous word, not the previous word and its space :)

Could I ask you, @chajadan, to review this? I'm looking for general code style advice, as I think this is very fringe behavior for Plover. Tell me what you think.

Ted

### DIFF
--- a/plover/test_translation.py
+++ b/plover/test_translation.py
@@ -9,6 +9,7 @@ from mock import patch
 from steno_dictionary import StenoDictionary, StenoDictionaryCollection
 from translation import Translation, Translator, _State, _translate_stroke, _lookup
 import unittest
+import sys
 from plover.steno import Stroke, normalize_steno
 
 def stroke(s):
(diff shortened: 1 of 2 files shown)


morinted:
@mighele can I ask you to review this? I imagine you are much more familiar with this part of the code that I am and your feedback would be invaluable.

mighele:
Hi Ted.

Yes I'll look at this

Mike

morinted:
@benoit-pierre can you test this functionality on Linux please? I'll also squash these before any submission.

benoit-pierre:
It does not work great on my setup because of the multiple keyboards, some with colemak with capslock and backspace switched, some with colemak without the switch: somehow depending on the keyboard I use, Plover will send a control+capslock instead of control+backspace...

But that's an unusual setup so let's merge this.

benoit-pierre:
Of course, twice the size may not work on other platforms, and we may have to take other factors, such has the phase of the moon, ambient temperature, etc... So please test, if that magic multiplier is not enough, will increase it...

P.S.: I'm definitively adding wxWidgets to the list of GUI toolkit to avoid.

PR #277, opened 2015-10-09
Better normalize number strokes to insert dash, fixes #276

### LINKED ISSUE #276: Number bar entries aren't interpreted liberally
I was trying to define a stroke with a number in the middle of it, and discovered that Plover doesn't process "18" in a stroke, but rather expects "1-8". Because there are no duplicate numbers in the layout, we should not require the explicit dash. Not to mention, the Plover add stroke dialogue doesn't add the dash so entries added with Plover won't work by default.

morinted:
Interpret number strokes so that an entry like 18 gets read as 1-8. I don't think that users should have to think about explicit dashes when dealing with numeric entries.

Additionally, this change makes it so that the Plover add translation dialogue correctly inserts "1-8" into the dictionary. The regex string for this kind of stroke is [1-4][6-9]. That means that numbers on opposite sides of the keyboard in the middle this stroke effectively count as an implicit asterisk.

The only issue I can think of would be that the user could have duplicate strokes in their dictionary if they have both "19" and "1-9" defined, for example. But is that a big enough issue compared to not being able to define strokes with numbers in them? I don't think it is, but you, the reviewer, can weigh in.

### DIFF
--- a/plover/steno.py
+++ b/plover/steno.py
@@ -16,6 +16,7 @@ import re
 STROKE_DELIMITER = '/'
 IMPLICIT_HYPHENS = set('AOEU*50')
 
+
 def normalize_steno(strokes_string):
     """Convert steno strings to one common form."""
     strokes = strokes_string.split(STROKE_DELIMITER)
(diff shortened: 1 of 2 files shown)


morinted:
@mighele This should be a quick review if you would like to check it over.

morinted:
I've been running this for a few a couple weeks and it has been reliable. I'm going to submit.

PR #286, opened 2015-10-21
Produce separate dictionaries by default, fixes #109

### LINKED ISSUE #109: Request - Two dictionary files
Plover should come with two dictionary files. The base one dict.json and the user's custom one (starts off empty) called userdict.json.

That way if a release of Plover requires the users to update their base dictionary, their own entries will not be affected.

Also, when using the new feature to add my own dictionary entry it re-orders the dictionary file to be alphabetical. So if I want to see my custom strokes that I added to my dictionary with a text editor, they are now no longer at the top of the file where I put them, but they are mixed in amongst all the other entries in the file.

If the userdict.json file is the last dictionary in the list in Plover, all new custom entries will be put into this file and the base dictionary file will not be affected.

I am willing to make the code change myself if developer time is an issue.

Thanks
Mike N
Windows 7
Plover 2.5.1

morinted:
- Instead of `dict.json` we now have, in order of least precedence:
	- `main.json`
	- `commands.json`
	- `user.json`
- Tests know how to handle these as well

User being a blank JSON file that the user can add to without worrying about their updates being lost in the huge default dictionary.

@stenoknight Could I get you to look over this? The code is mostly simple and you don't really have to dig into it -- the behavior is a more generalized version of what was there before.

The effective change is that:

- We don't rename the dictionary from what it is in the repo. E.g. 'main.json' and 'commands.json' are both in the repository and in the user's default config folder. Previously, we had `main.json` and `commands.json` in the repo and `dict.json` in the config folder. It's just an implementation note but I want to make sure you're aware.
- The new highest-priority/bottom of list dictionary is `user.json`, with the contents `{}` (empty dictionary) to let users create their own dictionary without even realizing it.

Thawts?


plover/main.py
```diff
@@ -48,14 +48,19 @@ def init_config_dir():
...
-        dict_filenames = glob.glob(os.path.join(ASSETS_DIR, '*.json'))
-        for dict_filename in dict_filenames:
-            unified_dict.update(json.load(open(dict_filename, 'rb')))
-        ordered = OrderedDict(sorted(unified_dict.iteritems(), key=lambda x: x[1]))
-        outfile = open(DEFAULT_DICTIONARY_FILE, 'wb')
-        json.dump(ordered, outfile, indent=0, separators=(',', ': '))
+    def copy_dictionary_to_config(name):
+        source_path = os.path.join(ASSETS_DIR, name)
+        out_path = os.path.join(CONFIG_DIR, name)
+        if not os.path.exists(out_path):
```
morinted: Should probably rename unified dict to something more meaningful.

stenoknight: This all looks good. If you want to rename main.json, what about defaultdict.json or plovertheory.json or something along those lines?

morinted: I'm okay with main.json, if you are. It's a real word, it's short and to the point. These dictionaries will only be made if the user doesn't have any dictionaries.

In that comment I was just talking about renaming the variable because "unified_dict" was the thing I replaced. A better piece would be something "unsorted_dict = json.load([...]" for lines 55 and 56

stenoknight: Ah, sorry. Reading comprehension fail. Variable naming is definitely not my strong suit, so I leave it entirely to your discretion there.

Yeah, I have no problem with main.json if you're cool with it.

Issue #133, opened 2013-11-07
config file gets bogus values


balshetzer:
There have been several reports that the config file can have bogus values be written to it. Of course, it's also possible for people to write bogus values. Plover should make sure not to write bad values to the config file but also be resilient to them.

port = None. This causes an OSError instead of the expected SerialException and so isn't caught.

[Main Frame]
x = -32000
y = -32000

This causes the window to not be visible.

morinted:
I think this requires better checking around all loads and maybe even writes. The window problem may be addressed when #272 gets solved.

PR #289, opened 2015-10-27
 make sure main window is placed within screen boundaries

benoit-pierre:
+ one additional minor cleanup

### DIFF
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -18,11 +18,9 @@ class AddTranslationDialog(wx.Dialog):
     other_instances = []
     
     def __init__(self, parent, engine, config):
-        pos = (config.get_translation_frame_x(), 
+        pos = (config.get_translation_frame_x(),
                config.get_translation_frame_y())
-        wx.Dialog.__init__(self, parent, wx.ID_ANY, TITLE, 
-                           pos, wx.DefaultSize, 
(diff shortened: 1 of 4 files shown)


morinted:
Awesome work! Just a couple questions:

- Can I ask you to expand a little bit on how this fixes the problem?
- Is it true that other GUI windows are not protected (or are they since they will have config values loaded unlike the main frame?)
- Does this need testing on OSX and Windows?

benoit-pierre:
The call to `self.SetRect(AdjustRectToScreen(...))` is already present for other windows. The problem with the main one was the call to `self.SetPosition(...)` after (plus the fact that the configuration is loaded during the init, so we can't use it right away).

I tested on Linux, and the Windows version with Wine.

benoit-pierre:
To be clear, the goal is to fix the case when config values are out of bounds (e.g. you send you config to someone using a smaller screen size).

morinted:
Thank you for testing it. I'll build this and try to break it on OS X.
Expect a merge tomorrow if all goes well.

One case that I see fixing is bogus values getting read in. I'll try with
the config set with -32000 X and Y as that happened sometimes. See #133

benoit-pierre:
Yes, this fix this case too.

morinted:
Thanks again Ben, another great commit. I appreciate it a lot.

PR #293, opened 2015-11-02
Fix paper tape width for alignment and no bottom scrollbar

KarlHegbloom:
This makes it so that all of the letters in the "paper tape" display are aligned with the letters in the label at the top, with no bottom scrollbar.

### DIFF
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -54,15 +54,16 @@ class StrokeDisplayDialog(wx.Dialog):
         sizer.Add(box, flag=wx.LEFT | wx.RIGHT | wx.BOTTOM | wx.EXPAND, 
                   border=UI_BORDER)
         
-        self.header = MyStaticText(self, label=ALL_KEYS)
+        self.header = MyStaticText(self, label=ALL_KEYS+"  ")
         font = self.header.GetFont()
         font.SetFaceName("Courier")
         self.header.SetFont(font)
         sizer.Add(self.header, flag=wx.LEFT | wx.RIGHT | wx.BOTTOM, 
-                  border=UI_BORDER)
+                  border=UI_BORDER + 4)
         sizer.Add(wx.StaticLine(self), flag=wx.EXPAND)
 
-        self.listbox = wx.ListBox(self, size=wx.Size(210, 500))
+        listboxwidth = self.header.DoGetBestSize().GetWidth() + 3
+        self.listbox = wx.ListBox(self, size=wx.Size(listboxwidth, 500))
         font = self.listbox.GetFont()
         font.SetFaceName("Courier")
         self.listbox.SetFont(font)


benoit-pierre:
This really depend on what font/theme is used. Hard-coding a few extra pixels here and there is not going to work. In fact, on my system, the font used is not even fixed width...

The following patch takes care of this:

```diff
diff --git c/plover/gui/paper_tape.py w/plover/gui/paper_tape.py
index 3dddbcb..e0b3e11 100644
--- c/plover/gui/paper_tape.py
+++ w/plover/gui/paper_tape.py
@@ -55,17 +55,16 @@ class StrokeDisplayDialog(wx.Dialog):
                   border=UI_BORDER)
         
         self.header = MyStaticText(self, label=ALL_KEYS)
-        font = self.header.GetFont()
-        font.SetFaceName("Courier")
-        self.header.SetFont(font)
+        fixed_font = self.header.GetFont()
+        fixed_font.SetFamily(wx.FONTFAMILY_MODERN)
+        self.header.SetFont(fixed_font)
         sizer.Add(self.header, flag=wx.LEFT | wx.RIGHT | wx.BOTTOM, 
                   border=UI_BORDER)
         sizer.Add(wx.StaticLine(self), flag=wx.EXPAND)
 
         self.listbox = wx.ListBox(self, size=wx.Size(210, 500))
-        font = self.listbox.GetFont()
-        font.SetFaceName("Courier")
-        self.listbox.SetFont(font)
+        font.SetFamily(wx.FONTFAMILY_MODERN)
+        self.listbox.SetFont(fixed_font)
 
         sizer.Add(self.listbox,
                   flag=wx.ALL | wx.FIXED_MINSIZE,
```

KarlHegbloom:
Courier was not fixed width? I think it was meant to emulate the appearance
of an old steno paper tape. Is the FONTFAMILY_MODERN guaranteed to be a
fixed width typeface?

I think the "magic numbers" are just the necessary pixel padding to match
the borders etc in the other elements to get vertical alignment, no
horizontal scroll bar,  and the Z not hidden after a vertical scrollbar
appears. The spaces at the end of ALL_KEYS seemed better than a magic
number since their width depends on the font size.

KarlHegbloom:
I am busy at work now.  This evening I will try that and also look at the
key suppression problem on other thread.

morinted:
I do prefer calculating the size rather than any static value because sizes
vary greatly between systems and DPIs.

benoit-pierre:
I don't have "Courier" on my system (I have "Courier New" available in the gvim font picker). According to the documentation, using FONTFAMILY_MODERN/ FONTFAMILY_TELETYPE will result in a fixed width font.

morinted:
@KarlHegbloom could you apply Ben's diff in your pull request? I'd like to work on this PR :smiling_imp:

benoit-pierre:
How about something [along those lines](https://github.com/openstenoproject/plover/compare/master...benoit-pierre:improve-paper-tape) instead?

KarlHegbloom:
It does not look right. See the attached screenshot. Before, it would put
the Z way over to the right. Now it does not.

On Sun, Nov 8, 2015 at 6:08 AM, Benoit Pierre <notifications@github.com>
wrote:

> How about something along those lines
> <https://github.com/openstenoproject/plover/compare/master...benoit-pierre:improve-paper-tape>
> instead?
>
> —
> <https://github.com/openstenoproject/plover/pull/293#issuecomment-154825758>
> .
>



-- 
Karl.Hegbloom@gmail.com
http://karlhegbloom.blogspot.com

KarlHegbloom:
Here is what the stroke display looks like in my master branch.

On Sun, Nov 8, 2015 at 12:12 PM, Karl Hegbloom <karl.hegbloom@gmail.com>
wrote:

> It does not look right. See the attached screenshot. Before, it would put
> the Z way over to the right. Now it does not.
>
> On Sun, Nov 8, 2015 at 6:08 AM, Benoit Pierre <notifications@github.com>
> wrote:
>
>> How about something along those lines
>> <https://github.com/openstenoproject/plover/compare/master...benoit-pierre:improve-paper-tape>
>> instead?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/openstenoproject/plover/pull/293#issuecomment-154825758>
>> .
>>
>
>
>
> --
> Karl.Hegbloom@gmail.com
> http://karlhegbloom.blogspot.com
>
>


-- 
Karl.Hegbloom@gmail.com
http://karlhegbloom.blogspot.com

benoit-pierre:
I'm still not seeing a screenshot.

For reference this is how it looks on my side:

 plover paper tape

It's missing a little space for the vertical scroll bar.

Note that we cannot center or right align the header and the list if we want them to still be aligned when the vertical scrollbar is displayed, we need to left align.

KarlHegbloom:
plover stroke display_001
plover stroke display_002

KarlHegbloom:
I have Benoit Pierre's fork of Plover as a remote, and did "git checkout improve-paper-tape", and ran it using this script:

#!/usr/bin/env python

import sys
from os.path import abspath, dirname

from plover.main import main
main()

benoit-pierre:
So the problem is again with the font. We need to find a reliable and portable way to select a fixed font.

morinted:
We could also let the user change the font.

benoit-pierre:
> We could also let the user change the font.

Several crashes latter.. this is starting to look simpler...

Last try: 0cfdff9130106b792e733136d553c250d11ac864

morinted:
@KarlHegbloom are you okay with closing this and moving our efforts to #296 ?

KarlHegbloom:
You decide.

morinted:
Thank you for the PR :D feel free to give feedback on #296

PR #295, opened 2015-11-09
Improve keyboard config

benoit-pierre:
This add a capture dialog when editing a keymap entry:

plover keyboard config

### DIFF
--- a/plover/gui/keyboard_config.py
+++ b/plover/gui/keyboard_config.py
@@ -16,7 +16,7 @@ UI_BORDER = 4
 
 class EditKeysDialog(wx.Dialog):
 
-    def __init__(self, parent, keys):
+    def __init__(self, parent, action, keys):
         super(EditKeysDialog, self).__init__(parent)
         self.sizer = wx.BoxSizer(wx.VERTICAL)
         instructions = wx.StaticText(self, label='Press on the key you want to add/remove.')
@@ -27,7 +27,9 @@ class EditKeysDialog(wx.Dialog):
         self.sizer.Add(buttons, border=UI_BORDER, flag=wx.ALL|wx.ALIGN_CENTER_HORIZONTAL)
         self.SetSizer(self.sizer)
         self.sizer.Fit(self)
+        self.action = action
         self.keys = set(keys)
+        self.original_keys = self.keys.copy()
         self.capture = KeyboardCapture(KeyboardCapture.SUPPORTED_KEYS)
         self.capture.key_down = lambda key: wx.CallAfter(self.on_capture_key, key)
 
@@ -43,7 +45,17 @@ class EditKeysDialog(wx.Dialog):
         return code
 
     def update_message(self):
-        message = ' '.join(sorted(self.keys))
+        message = '\nKeys for %s: ' % self.action
+        message += ' '.join(sorted(self.keys)) if self.keys else 'None'
+        message += '\n\nChanges: '
+        changes = []
+        for key in sorted(self.keys.union(self.original_keys)):
+            if key not in self.original_keys:
+                changes.append('+' + key)
+            elif key not in self.keys:
+                changes.append('-' + key)
+        message += ' '.join(changes) if changes else 'None'
+        message += '\n'
         self.message.SetLabelText(message)
         self.sizer.Fit(self)
         self.sizer.Layout()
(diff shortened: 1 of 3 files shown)


benoit-pierre:
I forgot to mention that I developed/tested on Linux, and when I tried to test the Windows version with Wine the capture dialog would not work the first time. Hopefully this is a problem with Wine and not the code.

morinted:
I'll try on both OS X and Windows, then.

morinted:
How much trouble would it be to make this show a diff compared to the current settings? For example, new letters could be bolded and removed letters could be added to a string at the bottom that says "Removed: a b c". Alternatively, you could have "Added: d c e" instead of the bold if that isn't possible.

benoit-pierre:
> How much trouble would it be to make this show a diff compared to the current settings? For example, new letters could be bolded and removed letters could be added to a string at the bottom that says "Removed: a b c". Alternatively, you could have "Added: d c e" instead of the bold if that isn't possible.

Where? In the main dialog, or in the capture dialog? IMHO looks complicated for not much pay off.

> And F-keys don't work on Mac, but that may be expected if we haven't updated the keycodes in OSX KBD control

I did add them (see [here](https://github.com/openstenoproject/plover/commit/89fe6c5e9246a06ad55a1db7cb75b7dae9a67361#diff-ac6ad4451f565448d405029169ef74cf)) based on documentation I found on the web, looks like I got it wrong...

Achim63:
F-Keys on OSX should work if they are set to behave as standard F-keys in System Preferences/Keyboard (otherwise you need to also hit the "fn" key to use them).

morinted:
I was using the fn key in my testing.

benoit-pierre:
OK, so is it the behavior we want? What are the keycodes when pressing those keys without `fn`?

benoit-pierre:
Mmm, from looking at this:
apple keyboard

this would not be an ideal user experience (forcing the user to set them to behave as standard). Unless... what does `fn+key` do when they are set to behave as standard F-keys in System Preferences/Keyboard? Does it gives back the multimedia/display keycodes?

benoit-pierre:
Also, does it behaves like this with any keyboard? Because unless the Apple keyboard has NKRO support, then we don't really care about it, no?

morinted:
I will test with a USB keyboard tomorrow and get back to you!

morinted:
Same effect on my real Windows machine -- doesn't work the first time after Plover has launched.

The reason that I like the diff view is that people know if they've made changes. For new users who are modifying the keymap, the probably won't know exactly what they are doing and the status could really help.

Maybe

```
Changes: None
```

```
Changes: +a +b -c -d
```

Achim63:
@benoit-pierre Yes, the fn-key gives the multimedia keys when the setting in system preferences is changed. I'd put it in the Readme and be done with it - Mac users usually know how to get back their F-keys.

morinted:
Oh, that might mean that we have to share the capture instance, doesn't it?

benoit-pierre:
Does it works on Mac OS X? Not just the capture dialog, but does Plover keeps on registering key presses after leaving the config dialog? (It does on Windows)

benoit-pierre:
> Oh, that might mean that we have to share the capture instance, doesn't it?

That would mean having to save the list of suppressed keys before using the instance, to restore it when the dialog is closed. Also, no capture instance already exists when another machine type is configured and you switch to NKRO and configure it.

If Mac OS X does work with multiple instance, might be easier to fix the Windows code to support it too.

morinted:
The behavior with Mac is when the capture dialogue is open, all key strokes
are caught and affect the capture dialogue, even when it is not focused.
When I close the dialogue, I can resume Ploverring without delay. After I
click save in the configuration dialogue, the effects are immediate and all
my changes are applied—meaning suppressed keys are suppressed and
unsuppressed keys aren't. I'd say, then, that it works perfectly on Mac. I
also do like the changes line—helps in case you forget the defaults (A and
Q are arbitrary on my set up so it helps to see).

benoit-pierre:
@morinted: I changed `winkeyboardcontrol.KeyboardCapture` implementation to support multiple instances. It works fine with Wine, can you test with a real Windows machine?

morinted:
Hm, not working well for me. The keyboard does not read on the configuration window only if Plover is *not* running. I don't believe that's what you intended :)

You can see a screen of the in console logging catching the keystrokes while the key capture window has not changed. 

screenshot

benoit-pierre:
> Hm, not working well for me. The keyboard does not read on the configuration window only if Plover is not running. I don't believe that's what you intended :)

No, I did not! Fixed it and rebased on master; the problem was the partial evaluation on this line:

```diff
-    allow = allow and callback(event)
+    allow = callback(event) and allow
```

morinted:
Looks all good over here. If you want to clean up history (or not), I can thereafter.

benoit-pierre:
I already cleaned up the history while rebasing, so it's all good to go.

Issue #294, opened 2015-11-07
Lookup window won't scroll horizontally and won't resize


morinted:
Using some longer strokes--noticed that the Lookup windows doesn't let me scroll except by clicking and dragging or using horizontal scroll wheel hardware.

benoit-pierre:
Is that still an issue with the new UI?

morinted:
Nope, looks like the new UI has word wrap.

Issue #256, opened 2015-09-23
Plover in OS X has strange rendering problems


morinted:
1. The main Plover window and all of its children don't hold focus in the system upon summoning. The Plover icon in the dock must be used to "wake" the app and get access to its menus.

2. There are no entries in the menu bar.

morinted:
Good news, bad news. Good news: I don't think this a problem with Plover. Bad news: I haven't been able to successfully build Plover with working menu bar on any commit, no matter how old. This means that the build deps for me are no good. Got to figure out what causes the problem and document it.

morinted:
So I think this partially has to do with wxPython updating. For one, I get retina-compatible builds, whereas it would seem that 2.5.8 was not retina. I'm going to look into customizing the menu bar, because at the moment it is not populated and Plover doesn't respond to Cmd+Q, and doesn't pull its windows to the front.

PR #296, opened 2015-11-09
Improve paper tape and suggestions

benoit-pierre:
paper tape:

- switch to a fixed width font
- rework controls:  change strokes list and header controls to text entries, and align them
- improve width/height calculations
- fix flickering when toggling "on top"
- do not mess with background color

suggestions:

- switch to a fixed width font
- fix flickering when toggling "on top"
- do not mess with background color
- get rid of MyStaticText hack

### DIFF
--- a/plover/dictionary_editor_store.py
+++ b/plover/dictionary_editor_store.py
@@ -1,4 +1,5 @@
 from plover.steno import normalize_steno
+from plover.gui.util import shorten_unicode
 
 STROKE = "STROKE"
 TRANSLATION = "TRANSLATION"
@@ -79,29 +80,7 @@ class DictionaryEditorStore():
         if col is COL_STROKE:
             result = item.stroke
         elif col is COL_TRANSLATION:
-            s = item.translation
-            # Detect and shorten unicode to prevent crashes
-            if isinstance(s, unicode):
-                def shorten_unicode(s):
-                    # Turn into 4 byte chars
-                    encoded = s.encode('utf-32-be')
-                    word = ""
-                    for i in xrange(len(encoded)/4):
-                        start = i * 4
-                        end = start + 4
-                        character = encoded[start:end].decode('utf-32-be')
-                        # Get 1 unicode char at a time
-                        character = character.encode('utf-8')
-                        # Within range?
-                        if (len(character) <= 3):
-                            word += character
-                        else:
-                            # Replace with white box
-                            word += unichr(9634)
-                    return word
-                s = shorten_unicode(s)
-            result = s
-
+            result = shorten_unicode(item.translation)
         elif col is COL_DICTIONARY:
             result = item.dictionary
         return result
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -59,6 +59,8 @@ class StrokeDisplayDialog(wx.Dialog):
         dc = wx.ScreenDC()
         dc.SetFont(fixed_font)
         text_width, text_height = dc.GetTextExtent(ALL_KEYS)
+        scroll_width = wx.SystemSettings.GetMetric(wx.SYS_VSCROLL_X)
+        scroll_height = wx.SystemSettings.GetMetric(wx.SYS_VSCROLL_X)
 
         self.header = wx.StaticText(self, label=ALL_KEYS)
         self.header.SetFont(fixed_font)
@@ -68,8 +70,9 @@ class StrokeDisplayDialog(wx.Dialog):
 
         self.listbox = wx.TextCtrl(self,
                                    style=wx.TE_MULTILINE|wx.TE_READONLY|wx.TE_DONTWRAP|wx.BORDER_NONE|wx.HSCROLL,
-                                   # Will show MAX_STROKE_LINES lines (+1 for inter-line spacings...).
-                                   size=wx.Size(text_width, text_height * (MAX_STROKE_LINES + 1)))
+                                   # Will show MAX_STROKE_LINES lines.
+                                   size=wx.Size(scroll_width + text_width,
+                                                scroll_height + text_height * MAX_STROKE_LINES))
         self.listbox.SetFont(fixed_font)
 
         sizer.Add(self.listbox,
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -7,6 +7,7 @@ import wx
 from wx.lib.utils import AdjustRectToScreen
 from collections import deque
 from plover.steno import STENO_KEY_ORDER, STENO_KEY_NUMBERS
+from plover.gui.util import find_fixed_width_font
 
 TITLE = 'Plover: Stroke Display'
 ON_TOP_TEXT = "Always on top"
@@ -53,19 +54,17 @@ class StrokeDisplayDialog(wx.Dialog):
         box.Add(self.choice, proportion=1)
         sizer.Add(box, flag=wx.LEFT | wx.RIGHT | wx.BOTTOM | wx.EXPAND, 
                   border=UI_BORDER)
-        
+
+        fixed_font = find_fixed_width_font()
+
         self.header = MyStaticText(self, label=ALL_KEYS)
-        font = self.header.GetFont()
-        font.SetFaceName("Courier")
-        self.header.SetFont(font)
+        self.header.SetFont(fixed_font)
         sizer.Add(self.header, flag=wx.LEFT | wx.RIGHT | wx.BOTTOM, 
                   border=UI_BORDER)
         sizer.Add(wx.StaticLine(self), flag=wx.EXPAND)
 
         self.listbox = wx.ListBox(self, size=wx.Size(210, 500))
-        font = self.listbox.GetFont()
-        font.SetFaceName("Courier")
-        self.listbox.SetFont(font)
+        self.listbox.SetFont(fixed_font)
 
         sizer.Add(self.listbox,
                   flag=wx.ALL | wx.FIXED_MINSIZE,
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -120,8 +120,12 @@ class StrokeDisplayDialog(wx.Dialog):
         self.show_text(self.formatter(stroke))
 
     def handle_on_top(self, event):
-        self.config.set_stroke_display_on_top(event.IsChecked())
-        self.display(self.GetParent(), self.config)
+        on_top = event.IsChecked()
+        self.config.set_stroke_display_on_top(on_top)
+        style = wx.DEFAULT_DIALOG_STYLE
+        if on_top:
+            style |= wx.STAY_ON_TOP
+        self.SetWindowStyleFlag(style)
 
     def on_style(self, event=None):
         format = self.choice.GetStringSelection()
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -15,7 +15,7 @@ UI_BORDER = 4
 ALL_KEYS = ''.join(x[0].strip('-') for x in 
                    sorted(STENO_KEY_ORDER.items(), key=lambda x: x[1]))
 REVERSE_NUMBERS = {v: k for k, v in STENO_KEY_NUMBERS.items()}
-MAX_STROKE_LINES = 38
+MAX_STROKE_LINES = 30
 STYLE_TEXT = 'Style:'
 STYLE_PAPER = 'Paper'
 STYLE_RAW = 'Raw'
@@ -56,11 +56,10 @@ class StrokeDisplayDialog(wx.Dialog):
         fixed_font = find_fixed_width_font()
 
         # Calculate required width and height.
-        # Extra spaces added to compensate for
-        # left margin and scroll bar width...
+        # Extra space to compensate for left margin...
         dc = wx.MemoryDC()
         dc.SetFont(fixed_font)
-        text_width, text_height = dc.GetTextExtent(ALL_KEYS + 3 * ' ')
+        text_width, text_height = dc.GetTextExtent(ALL_KEYS + ' ')
 
         # Default text style.
         text_style = wx.TextAttr()
@@ -77,8 +76,8 @@ class StrokeDisplayDialog(wx.Dialog):
 
         self.listbox = wx.TextCtrl(self,
                                    style=wx.TE_MULTILINE|wx.TE_READONLY|wx.TE_DONTWRAP|wx.BORDER_NONE,
-                                   # Will show 30 lines (+1 for inter-line spacings...).
-                                   size=wx.Size(text_width, text_height * 31))
+                                   # Will show MAX_STROKE_LINES lines (+1 for inter-line spacings...).
+                                   size=wx.Size(text_width, text_height * (MAX_STROKE_LINES + 1)))
         self.listbox.SetDefaultStyle(text_style)
         self.listbox.SetFont(fixed_font)
 
@@ -112,7 +111,13 @@ class StrokeDisplayDialog(wx.Dialog):
         event.Skip()
         
     def show_text(self, text):
-        self.listbox.AppendText(text + '\n')
+        if len(self.line_lengths) == MAX_STROKE_LINES:
+            self.listbox.Remove(0, self.line_lengths.pop(0))
+        if self.listbox.IsEmpty():
+            self.listbox.AppendText(text)
+        else:
+            self.listbox.AppendText('\n' + text)
+        self.line_lengths.append(len(text) + 1)
         
     def show_stroke(self, stroke):
         self.show_text(self.formatter(stroke))
@@ -129,6 +134,7 @@ class StrokeDisplayDialog(wx.Dialog):
         format = self.choice.GetStringSelection()
         self.formatter = getattr(self, format.lower() + '_format')
         self.listbox.Clear()
+        self.line_lengths = []
         for stroke in self.strokes:
             self.show_stroke(stroke)
         if STYLE_PAPER == format:
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -57,19 +57,29 @@ class StrokeDisplayDialog(wx.Dialog):
 
         fixed_font = find_fixed_width_font()
 
-        self.header = MyStaticText(self, label=ALL_KEYS)
+        # Default text style.
+        text_style = wx.TextAttr()
+        # Add a left margin.
+        text_style.SetLeftIndent(10)
+
+        self.header = wx.TextCtrl(self, style=wx.TE_READONLY|wx.TE_DONTWRAP|wx.BORDER_NONE)
+        self.header.SetDefaultStyle(text_style)
         self.header.SetFont(fixed_font)
-        sizer.Add(self.header, flag=wx.LEFT | wx.RIGHT | wx.BOTTOM, 
-                  border=UI_BORDER)
+        self.header.AppendText(ALL_KEYS)
+        sizer.Add(self.header, flag=wx.ALL|wx.EXPAND|wx.FIXED_MINSIZE, border=UI_BORDER)
+
         sizer.Add(wx.StaticLine(self), flag=wx.EXPAND)
 
-        self.listbox = wx.ListBox(self, size=wx.Size(210, 500))
+        self.listbox = wx.TextCtrl(self,
+                                   style=wx.TE_MULTILINE|wx.TE_READONLY|wx.TE_DONTWRAP|wx.BORDER_NONE,
+                                   size=wx.Size(210, 500))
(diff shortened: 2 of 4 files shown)


benoit-pierre:
Looking over the code, it seems the intent was to limit the number of strokes displayed:

```python
MAX_STROKE_LINES = 30
STYLE_TEXT = 'Style:'
STYLE_PAPER = 'Paper'
STYLE_RAW = 'Raw'
STYLES = [STYLE_PAPER, STYLE_RAW]

class StrokeDisplayDialog(wx.Dialog):
    
    other_instances = []
    strokes = deque(maxlen=MAX_STROKE_LINES)
```

So  – with the original code  – when closing and re-opening the paper tape window (manually, or triggered by toggling "on top"), only the last `MAX_STROKE_LINES` are displayed.

By honoring `MAX_STROKE_LINES` for sizing the list at creation and limiting the maximum number of lines, the code can be simplified a little: we don't have to worry about a vertical scrollbar messing with our text alignment and can just center the header and the list lines to align them with each other.

morinted:
If we implemented [Chris' feature suggestion](https://github.com/openstenoproject/plover/issues/297) then we would need to scroll. What are your thoughts on that suggestion? (PS. again offline tonight, should be working on some PRs tomorrow, I hope)

benoit-pierre:
Only for the suggestions, not for the paper tape, no?

Anyway, after the changes to paper tape, Chris' feature suggestion is pretty easy to implement, so I took a pass at it: https://github.com/benoit-pierre/plover/compare/improve-paper-tape-and-suggestions...benoit-pierre:improve-suggestions

It looks like this:

plover suggestions

I'm not sure the strokes right alignment is great for readability. Maybe swap sides with the word?

morinted:
I wouldn't mind keeping the current layout where strokes are indented.

benoit-pierre:
Easy enough: 9a829f0782e0da04c079b4e2a2cdec9183e9da04

plover suggestions

morinted:
Wow, looks cool. Do you think it makes more sense to append to the bottom
or add to the top? I'd probably default mentally to adding on top, because
then the last title would always be in the same place.

benoit-pierre:
Either is fine to me.

morinted:
Would you want to add that feature to this PR? I'd like it (adding to the top would be my preference). Any idea of how many strokes you'd want to keep? I'd suggest probably the same as Plover uses (max_buffer / the max length of a dictionary entry)

benoit-pierre:
Done.

Note: history size is set to the same number as 'minimum undo length' for the engine translation, but still a hard-coded number, since no constant exists.

morinted:
Thanks, Ben. I'll see if I can get my wxPython installed. Another thing is on my machine, #256 was never resolved. Right now I'm following along [these instruction](http://stackoverflow.com/questions/33134896/install-wxpython-2-8-for-ride-on-osx-el-capitan) despite this being for an older version of wxPython (new version is crashing on build). This is a big inconvenience :(

morinted:
Hm, font seems really small on my system (retina on Mac). I also don't get any tabbing for the strokes. I'd ask if you could get rid of the first entry being bold, too, as in all likelihood it could be a misstroke definition anyway.

Is it possible to grab the max-undo-length that is determined from the dictionary having an entry longer than 10 strokes? I have an entry that is "A/" a hundred times so that I get a long history.

screen shot 2015-11-12 at 7 16 51 pm

screen shot 2015-11-12 at 7 15 55 pm

benoit-pierre:
> Hm, font seems really small on my system (retina on Mac).

The size is normally set to be the same as the default GUI font, but yeah, it does look smaller...

> I also don't get any tabbing for the strokes.

Weird. So what version of wxpython are you using/did you manage to install? Can you check what value is calculated for said margins? (Maybe print the font size too while you're at it)

> I'd ask if you could get rid of the first entry being bold, too, as in all likelihood it could be a misstroke definition anyway.

Sure.

> Is it possible to grab the max-undo-length that is determined from the dictionary having an entry longer than 10 strokes? I have an entry that is "A/" a hundred times so that I get a long history.

Better: remove the "no-wrap" flag, so we get a horizontal scrollbar. Agreed?

morinted:
I am using wxPython 3.0.2.0 installed with brew. I'll check values tomorrow (working on my day job tonight).

> Better: remove the "no-wrap" flag, so we get a horizontal scrollbar. Agreed?

I was talking about the undo history length for the number of entries kept. Plover reads in the dictionary and then ignores the "minimum undo length" of 10 if the longest stroke length is > 10.

But yea, a horizontal scroll bar would be nice. If you are feeling generous, #294 Lookup window needs a horizontal scrollbar, too.

benoit-pierre:
> I was talking about the undo history length for the number of entries kept. Plover reads in the dictionary and then ignores the "minimum undo length" of 10 if the longest stroke length is > 10.

IMHO, it would be better to add a config entry, so you don't need that hack of defining a really long entry, and to use both at init and to determine the history size for the suggestions. 
Because, from what I understand, adding this magical entry to your dictionary results in many unnecessary lookups in said dictionary.

morinted:
You are correct on both fronts.

benoit-pierre:
I created a issue for this: #300

benoit-pierre:
And you just created one too :P.

morinted:
Yup, all good :D I closed mine, yours had a checklist. I'm off for tonight, I'll be back tomorrow or Saturday.

benoit-pierre:
I removed the bold attribute for the first stroke. Kept the "no-wrap" since that's what we want to get a horizontal scrollbar when a line is too long.

Note: I'll rewrite the history once the changes are given the okay.

KarlHegbloom:
Maybe it would be good to see what dictionary the stroke is defined by? I think that would mean adding that information to the dictionary entry... is that correct?

Instead of only the stroke suggestions for the last word or phrase... what about showing some look-ahead, like, if I hold down at least two keys, it matches anything that contains those, narrowing the completion dynamically as I hold down other keys, trying to find the right stoke? That would mean reading the set of pressed keys, rather than the log file.

Did you see what I said in the thread on the Google Group, about autokey and ibus-table?

morinted:
Okey, built again, this is what it's looking like now. I probably won't dive into the runtime stuff until tomorrow. DZ is wrapping and there are some width issues with the list boxes.

screen shot 2015-11-13 at 5 06 10 pm

morinted:
### Font size

Fixed font point size is 11, standard (listbox) is 13.

I managed to fix the font sizing issue by normalizing them; `fixed_font = find_fixed_width_font(point_size=standard_font.GetPointSize())` **or** by setting the default against `SYS_ANSI_FIXED_FONT` instead of the default GUI font. Maybe you could try this constant on your system and see if that font's size is best?

### Sidebar

Looks like the sidebar space that makes my DZ wrap is because of the vertical scrollbar that appears.

How does this code affect your system?

```diff
diff --git a/plover/gui/paper_tape.py b/plover/gui/paper_tape.py
index 0453e0a..0c5dc05 100644
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -60,10 +60,12 @@ class StrokeDisplayDialog(wx.Dialog):
         dc = wx.ScreenDC()
         dc.SetFont(fixed_font)
         text_width, text_height = dc.GetTextExtent(ALL_KEYS + 2 * ' ')
+        text_width += wx.SystemSettings.GetMetric(wx.SYS_VSCROLL_X)

         self.header = wx.StaticText(self, label=ALL_KEYS, style=wx.ALIGN_CENTRE_HORIZONTAL)
         self.header.SetFont(fixed_font)
-        sizer.Add(self.header, flag=wx.ALL|wx.FIXED_MINSIZE|wx.ALIGN_CENTER_HORIZONTAL, border=UI_BORDER)
+        sizer.Add(self.header, flag=wx.ALL|wx.FIXED_MINSIZE|wx.ALIGN_LEFT, border=UI_BORDER)

         sizer.Add(wx.StaticLine(self), flag=wx.EXPAND)
```

### Indent

Margin was set to `80`. I tried giving it `100` (on all three styles) but still there's no change. Looking at http://docs.wxwidgets.org/trunk/classwx_list_box.html, looks like TAB isn't supported -- maybe that's related. I think it might just be best to stick with spaces, shame that that is; `self.listbox.WriteText('  ' + '/'.join(strokes) + '\n')`. I realize that this won't indent on wrap.

The closest I got was using RichTextCtrl instead of TextCtrl. It only worked for the first suggestion... Maybe you can see what's wrong here https://gist.github.com/morinted/780f603942c2a894d909

benoit-pierre:
Suggestions:
- use spaces to indent strokes
- fix fixed font size

Paper tape:
- simplify code a little more
- fix fixed font size

Regarding your issues on Mac OS X, it seems there is a (6 years old) bug regarding `wx.TE_DONTWRAP` support: http://trac.wxwidgets.org/ticket/11321

In which case we're screwed... Sure, we can fiddle around with the text width calculations to try to avoid wrapping with the paper tape, but what about the suggestions?

Also, the Windows version is a complete disaster with Wine, but I'm hopping it's a problem with Wine and not wxWidgets...

morinted:
Ah, I didn't even catch on the lack of horizontal scrolling in OS X. Herp, derp. That's a bad bug, man. Um,  well, for the paper tape my code at least added enough space to prevent wrapping. If that code isn't needed on other platforms, we could include it conditionally, as always. For the suggestions, we should probably just settle on wrapping. We could manually insert line breaks to fake nested indentation such as:

```
???????
    STPH/STPH/STPH/
       STPH/STPH/
       STPH/STPH
    STPH/-7
    H-F/-7
    H-F/H-F/H-F/H-F/H-F/
        H-F/H-F
```

Basically manually providing nested indent by doing nearest to x characters, wrapping at the / where possible, then indenting twice as much on wrapped lines. Your opinion welcome, as this may be overly complicated.

morinted:
Yeah, here's what I see on Windows 10:

capture

morinted:
I see a cursor on the Windows build. I used this on my rich text version that worked for that: ` self.listbox.GetCaret().Hide()`

benoit-pierre:
- paper tape: tweak size to take scrollbars into account
- suggestions: manually wrap overlong strokes

Note:
- `RichTextCtrl` does not seem to be supported on Linux
- `self.listbox.GetCaret().Hide()` does not work with `TextCtrl`

benoit-pierre:
Added: reflow the suggestions text on resize:

plover suggestions big plover suggestions medium plover suggestions small

morinted:
Looks really nice, this is really getting there.

- I noticed that the alignment of the title label is off -- in my gist I just set the alignment to left (because of the scrollbar width increase):
Scrollbar alignment

- The suggestions window crashes Plover if I try to use an emoji. You can see how I handled this in the dictionary editor: https://github.com/openstenoproject/plover/blob/master/plover/dictionary_editor_store.py#L83
    We could pull that out into the util and call it something like sanitize_wx_string, using it when the output might have extended unicode.

morinted:
Seems to also not totally handle the scroll bar width as you can see here:

screen shot 2015-11-18 at 8 36 14 am

morinted:
Ugh I hate OSX sometimes. When I enable scroll bars to be always on, restart Plover, and now it looks like this:

screen shot 2015-11-18 at 9 20 04 am

- S doesn't have a space until I use the number bar during execution.
- Both D and Z wrap after the number bar is taking its space.

benoit-pierre:
*sigh*

morinted:
My sentiments exactly.

benoit-pierre:
So for the suggestions, let's just assume the vertical scroll bar is twice its size, because... I tried to see if `GetMargins` would explain why the calculation are wrong, but it returns `(-1, -1)`. Since the documentation mention for some platform `SetMargins` must be called first, I tried it:
```
(plover:22258): GLib-GObject-WARNING **: invalid cast from 'GtkTextView' to 'GtkEntry'

(plover:22258): Gtk-CRITICAL **: IA__gtk_entry_get_inner_border: assertion 'GTK_IS_ENTRY (entry)' failed

(plover:22258): Gtk-CRITICAL **: IA__gtk_entry_set_inner_border: assertion 'GTK_IS_ENTRY (entry)' failed
```

Of course a following call to `GetMargins` would still return `(-1, -1)`...

benoit-pierre:
Of course, twice the size may not work on other platforms, and we may have to take other factors, such has the phase of the moon, ambient temperature, etc... So please test, if that magic multiplier is not enough, will increase it...

P.S.: I'm definitively adding wxWidgets to the list of GUI toolkit to avoid.

benoit-pierre:
I'm not seeing a crash with those emojis: 😰😀🙏

Is that another Mac OS X specific bug? Does it crash on Windows too?

Also, in your code to handle the issue, why encode to 'utf-32-be' first? You can directly iterate on the characters if it's an Unicode string.

morinted:
Well, lots of emoji are length 2-4 (🇷🇺, 🇨🇦). However, they should be
read as one character and utf-32 groups the large strings into groups of
1-4 (if you get what I mean…?) If my logic is wrong, though, of course I'd
love the simpler code.

Basically, looping over with Unicode gives me pieces of long characters
(i.e. emoji are stored as \u1234\u1234 and each of the escape groups gets
iterated when I'm only interested in the whole character)

I haven't tested on Windows, I'll do that tonight, most probably. If this
is another OS X specific bug, I'm going to be so bummed out. There's no
reason that OS X shouldn't handle emoji. Should we consider migrating away
from wx? It's not like Plover's GUI is fancy or heavy weight.

benoit-pierre:
```python
>>>  s = u'😰😀🙏'
>>> len(s)
3
>>> for c in s:
...     print c, len(c.encode('utf-8'))
...
...
😰 4
😀 4
🙏 4
```

morinted:
>>> s = u'😰😀🙏'
>>> len(s)
6
>>> for c in s:
...     print c, len(c.encode('utf-8'))
...
� 3
� 3
� 3
� 3
� 3
� 3
>>>

benoit-pierre:
Which version of Python? I have the same behavior (`len(s) == 3`) with 2.7.10 and 3.5.0.

benoit-pierre:
Regarding the GUI:

I think the way wxWidgets work is a recipe for bugs: one implementation per platform, each using a completely different GUI toolkit?

Migrating to another toolkit would be some work, but it could also be used as an opportunity to clean up the code.

I personally don't think having a native look on each platform is more important than being able to use a GUI toolkit that work more consistently on all those platforms.

morinted:
That behavior happens in 2.7.10, in 3.5.0 it's working properly like yours.

My bias is towards the modern web-based frameworks, like electron. Then we
can still make something that **looks** good, though I'd understand if
people aren't comfortable with the weight of such a GUI and moving so far
away from Python. Is the only other popular solution QT?

morinted:
Kivy and PyQT look most viable, just from a quick Google search. How do you feel about things?

benoit-pierre:
OK, so depending on the way Python2 is built, Unicode strings are either UCS2 or UCS4 encoded.

Python 3 however work consistently.

benoit-pierre:
Not a fan of (modern or not) web-based frameworks, to put it mildly.

QT is mature, I've not used it a lot, but in my experience the documentation is good (certainly much better than GTK, Clutter, ...). I have never used PyQT.

morinted:
PyQT and Kivy both seem good, though Kivy is definitely less mature. I'd be
cool with migrating towards PyQT.

morinted:
Uh, that being said, if we were changing GUI frameworks, I'd probably want
to revamp Plover's GUI entirely. I have some gripes with how it is laid out
now—the tiny window, the configuration dialogue as it is. Definitely at the
very least some reorganization is in store. I'd have to draw out all of
Plover's features and try to group them and make something appealing.

benoit-pierre:
Agreed.

morinted:
Hm, did the label font get separated from the text box font? This is what it looks like for me:

screen shot 2015-11-21 at 11 45 42 am

morinted:
Couple things on the wide Python thing: https://github.com/Homebrew/homebrew/pull/32368, and how to handle it with narrow Unicode https://gist.github.com/tdsmith/e8b46f5a7ce49b63fb7f

Seems like most people will have narrow unicode by default -- should we just accept that and use workarounds for long characters?

benoit-pierre:
> Hm, did the label font get separated from the text box font?

No, no. So how many characters should be added to prevent wrapping? Is 2 enough?
```python
        text_width, text_height = dc.GetTextExtent(ALL_KEYS + 2 * ' ')
```

> Seems like most people will have narrow unicode by default -- should we just accept that and use workarounds for long characters?

OK, but what about the fact that it does not crash on Linux? Did you test on Windows?

morinted:
2 should be good. Haven't tested on Windows; I'm just very busy at work
right now. Deadline Monday... I'll check it out soon though.

benoit-pierre:
OK, no hurry.

morinted:
Looks like on Windows it just gives me boxes. May be the font not handling the icon.
Emoji

No crash, though, so that's nice.

morinted:
@benoit-pierre Would you like to make the change by adding width for the Mac sidebar?

Don't worry about the Mac crash, I'll fix that in another PR.

benoit-pierre:
> Looks like on Windows it just gives me boxes. May be the font not handling the icon.

Yes, normal missing glyph behavior.

> Would you like to make the change by adding width for the Mac sidebar?

I added 2 extra spaces, please test if it's enough.

> Don't worry about the Mac crash, I'll fix that in another PR.

Move `shorten_unicode` to `gui/util`, and make it a no-op for Linux/Windows?

morinted:
I'll test the extra spaces on Mac soon (been out for the holidays), but yes, move the shorten unicode code to util. I was considering also moving it away from the osxkeyboard handling, but iirc that's not the same function, just similar. I also wanted to change the unknown character to the [Unicode replacement character](http://www.fileformat.info/info/unicode/char/0fffd/index.htm): �

morinted:
All right, the 2 extra spaces give enough room for things to work. Very close to being done with this commit.

Here's the alignment issue I have with the title:

Label unaligned slightly

And not necessary, but if you were willing you could move shorten_unicode to gui/util and make it OS X only, if you'd like. It would apply for both suggestions and dictionary editor. Also, I'd like to change the character to �, (`\ufffd`)

morinted:
Oh, that alignment is so minor that I wouldn't *really* worry about it. There are bigger fish to fry.

benoit-pierre:
OK, I did the `shorten_unicode` change, tested on Linux and Windows 10 VM. Please check on Mac.

As for the alignment... I'm not sure there is a way to really fix it in all cases (platforms, fonts, themes...)

morinted:
Okay, the dictionary editor works great with that change.

The suggestions window:

```
2016-02-03 4:25:55.983 PM Plover[31746]: Traceback (most recent call last):
2016-02-03 4:25:55.983 PM Plover[31746]:   File "wx/_core.pyc", line 16766, in <lambda>
2016-02-03 4:25:55.983 PM Plover[31746]:   File "plover/gui/suggestions.pyc", line 238, in show_stroke
2016-02-03 4:25:55.983 PM Plover[31746]:   File "plover/gui/suggestions.pyc", line 180, in show_suggestions
2016-02-03 4:25:55.983 PM Plover[31746]:   File "wx/_core.pyc", line 13092, in WriteText
2016-02-03 4:25:55.983 PM Plover[31746]: UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 2: ordinal not in range(128)
```

benoit-pierre:
And now?

morinted:
```python
        # util.py:107
        if len(character) <= 2:
```
Then it all works, except that most of the suggestions says "no suggestions" for my invalid characters "emoji". Just to illustrate, I enter a cat via stroke in the lookup dialog with stroke suggestions open:

screen shot 2016-02-03 at 4 51 01 pm

benoit-pierre:
For reference, the cat problem is a bug in Python 2 narrow builds when a string contains surrogate pairs:

```python
>>> import sys, re
>>> sys.platform, sys.maxunicode
('linux', 1114111)
>>> re.findall(r'[-\'"\w]+|[^\w\s]', 'account 😻')
['account', '😻']
>>> re.findall(r'[-\'"\w]+|[^\w\s]', 'account 😻', re.U)
['account', '😻']
```

```python
>>> import sys, re
>>> sys.platform, sys.maxunicode
('win32', 65535)
>>> re.findall(r'[-\'"\w]+|[^\w\s]', 'account 😻')
['account', '😻']
>>> re.findall(r'[-\'"\w]+|[^\w\s]', 'account 😻', re.U)
['account', '\ud83d', '\ude3b']
```

Since the issue was already present, how about merging and we'll deal with it latter?

morinted:
All right, I can agree with that. Do you want to squash the commits anymore or keep all 23?

benoit-pierre:
No, no more rewrites, you can merge.

PR #301, opened 2015-11-13
machine/sidewinder: fix arpeggiate key suppression

benoit-pierre:
Don't suppress the key if arpeggiate mode is not enabled.


morinted:
Thanks Ben, I realize this seems trivial but I think it's important :+1:  Really appreciaTed.

PR #302, opened 2015-11-14
Update and fix Mac configuration and build instructions

### LINKED ISSUE #243: Version number for OS X
The info window in the Finder gives 0.0.0 as version for Plover. Would be nice to have the actual version show up, so we can see if we have the latest one (i.e. edit the Info.plist accordingly).

### LINKED ISSUE #256: Plover in OS X has strange rendering problems

morinted:
- Argv_emulation is now set to False, fixes #256
- Add CFBundle information so that .app has a valid PList, fixes #243
- Add menu bar to app, Plover now responds to Cmd+Q
- Update developer readme for OSX

@benoit-pierre Can you make sure this doesn't break anything on other platforms? I'm concerned about the MenuBar creation. If it affects other platforms, then this will need to be conditional.


setup.py
```diff
@@ -67,7 +76,7 @@
             'pyobjc-core>=3.0.3',
             'pyobjc-framework-Cocoa>=3.0.3',
             'pyobjc-framework-Quartz>=3.0.3',
-            'wxPython>=2.8',
+            'wxPython>=3.0',
```
benoit-pierre: If we increase the version requirements here (which I'm fine with), than we should do it in `plover/main.py` too, no?

morinted: Yep... is it okay to get this update on all platforms? Are you also using 3+ on Linux?

benoit-pierre: I am on Arch Linux. Looks like it's available on Ubuntu starting with Vivid (15.04), so I think we're good.

morinted:
Updated.

morinted:
All right, I've tested this on Windows. I'm going to merge now, because this thing makes my life on Mac so much easier. If there are issues on Linux, let me know and I'll make a fix @benoit-pierre

PR #303, opened 2015-11-15
oslayer/config: improve check for pkg_resources support

benoit-pierre:


morinted:
Is your build.py script going to come into play in another PR? This one still gives the file exists warning.

morinted:
To be clear, I don't think this PR fixed what you wanted it to. This is what I get on launch:

Plover Error Screenshot

benoit-pierre:
That's with pyinstaller 2.1, it works with 3.0.

morinted:
Ah, of course it is. Maybe we should write in to the dev instructions that it's `pip install pyinstaller==3.0` for explicitness. Thanks for fixing this though.

PR #304, opened 2015-11-16
Improve log + errors handling

### LINKED ISSUE #48: Direct errors to a log
Errors are currently directed to the void by xwidgets. Directing it to a file in a known location will help with bug reports.

benoit-pierre:
Change the logging mechanism, with default output to `stderr` (warnings and above) and a hard-coded `"plover.log"` (info and above) in the standard configuration directory.

It's used like this:

```python
from plover import log

log.info('public service announcement')
log.warning('warning!')
log.error('steno machine is on fire')
```

The default strokes/translations log is now `"strokes.log"` (using `log.stroke(...)` and `log.translation(...)`. The format of said log and of the related configuration entries have been left unchanged for backward compatibility.

Note:
- strokes/translations are not sent to the regular handlers (stderr and `"plover.log"`).
- the code checks and show an error if `"plover.log"` is used

**Rational** 

Having a hard-coded log make it easier to debug issues:
- it's available earlier in the startup process since it does not depend on the configuration
- when a user is having an issue with Plover, you can ask him for the log without worrying about it's placement and whether it's enabled or not

**GUI handlers**

2 GUI handlers are available, to display visual notifications for warnings and above:
- a default fallback platform independent wxWidgets based based implementation
- a Linux specific implementation using the "Desktop Notifications Specification" through DBus

**Additional related changes**

- exception: remove unused SerialPortException
- dictionary: more informative exceptions (show dictionary name, JSON error)
- machine: improved logging (use `log.error` instead of `print` on error opening the serial port)
- machine/keymap: add some sanity checks (key bound to multiple actions)

This fixes #48.


benoit-pierre:
I noticed that dictionary tests are failing on master:

```python
self = <plover.dictionary.test_default_dict.TestCase testMethod=test_no_duplicates_unsorted_file>

    def test_no_duplicates_unsorted_file(self):

>       d = json.load(open(DICT_PATH), object_pairs_hook=read_key_pairs)
E       IOError: [Errno 2] No such file or directory: 'plover/assets/dict.json'

plover/dictionary/test_default_dict.py:27: IOError
```

morinted:
Yes, I believe #286 removes that test, it was there because we used to
compose dictionaries. You would have to copy your dict.json that Plover
generated to test it.

benoit-pierre:
Indeed, are you going to merge it?

morinted:
Done.

benoit-pierre:
:+1:

morinted:
I am very excited about this change, too. I don't have time to review it all tonight, but I look forward to tomorrow evening. Should help debugging problems with end users, big time.

morinted:
Hm, can you automatically check if the logging file is plover.log, change it to strokes.log, and notify the user? Also having some issues with displaying the warnings in wx on the Mac, but I'll fix it this time.

benoit-pierre:
More Mac problems, yeah! What's happening?

morinted:
screen shot 2015-11-19 at 10 57 10 pm

Right now I'm working on getting my build to recognize the Treal, so I'm a little distracted...

benoit-pierre:
Is that not how other desktop notifications are displayed?

morinted:
Hm, no, they are usually thrown into the notification center. I'd probably prefer just to default to the wx fall back.

benoit-pierre:
This is the wx fallback (the DBus log is only enabled on Linux).

morinted:
Right. It's a weird widget. Can't interact with it, disappears on lost focus, doesn't show up in list of windows. Was a dialogue not favorable?

benoit-pierre:
You're not supposed to interact with it, it's just a notification. It disappears after a timeout (higher timeout for higher severity), mimicking the way desktop notifications work. On Linux with the DBus implementation, the notification does not have focus, does not appear in the list of windows. I tested with Wine, and it seems to use the standard Windows mechanism, which behaves similarly. Also note that you could have several notifications at the same time, and they don't block which ever code spawned them.

morinted:
Okay, it's not pretty but it's functional. I do like having the logging.
Thank you, again, for a job well done.

Issue #214, opened 2014-11-01
portable plover breaks when drive names change


balshetzer:
Plover stores absolute paths in the config file, this breaks when the external drive gets a new name. This can be fixed by storing paths relative to the config file.

morinted:
Was this done? If not, it seems important to do because then Plover would be fully useable off of cloud storage.

benoit-pierre:
I took a pass at this: https://github.com/openstenoproject/plover/compare/master...benoit-pierre:config-relative-paths

When a path in the config (log, dictionaries) is relative, it's assumed to be relative to `CONFIG_DIR`.

When setting a config path, it's stored as relative when bellow `CONFIG_DIR`.

morinted:
I think that that's the right way to go. It would need to be documented, but it makes sense. The current behavior is that a config next to the Plover executable make Plover portable, right? You're saying that if I added `../dictionaries/dict.json` that that would be stored as `/home/ted/steno/dictionaries/dict.json`, right? What if it stored relative if, 1) Plover is next to its config and 2) the dictionary is on the same drive?

I can definitely see someone having a dictionary directory not under Plover's folder, however I *don't* think its necessary to code around it since we can just declare that dictionaries have to be in the same folder as the Plover executable.

benoit-pierre:
Another case that is handled by this change (beside portable plover) is simply to be able to version control `~/.local/share/plover` and use it on multiple machines without 
worrying about my home directory being different:

```
[Dictionary Configuration]
dictionary_file1 = main.json
dictionary_file2 = commands.json
dictionary_file3 = user.json
```

morinted:
That's true. I'm very excited to be able to store Plover.dmg and Plover.exe in a folder on Google Drive next to my dictionaries. Less fuss than it is right now.

benoit-pierre:
And yes, if you put `../dictionaries/dict.json`, the next time the config is saved it will be changed back to an absolute path. `dictionaries/dict.json` would work (any path bellow the configuration directory).

morinted:
Is it below or below and next to? i.e. does `./dict.json` work? Err, also do you store as `./dict.json` or `dict.json`?

benoit-pierre:
`./dict.json` works, and is converted to `dict.json` the next time the config is saved.

morinted:
Good to know. Well, I am excited for this change too. Thanks for all the work you've done on Plover lately. You are rocking the steno-world :D

[...]

PR #305, opened 2015-11-17
config: store paths below CONFIG_DIR as relative

benoit-pierre:
Fix for #214


morinted:
Can you update this given the new stroke logging file? By the way, the notifications look really good in Windows 7, 8, and 10.

benoit-pierre:
Done.

PR #311, opened 2015-11-26
Improve machine logging, clarify Mac documentation

morinted:
Using the Treal, I had to make some changes and these helped me through it.

### DIFF
--- a/plover/machine/stentura.py
+++ b/plover/machine/stentura.py
@@ -150,6 +150,7 @@ import array
 import itertools
 import struct
 import time
+from plover import log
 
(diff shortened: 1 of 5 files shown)


plover/machine/stentura.py
```diff
@@ -658,5 +659,6 @@ def run(self):
             _loop(self.serial_port, self.finished, self._notify, self._ready)
         except _StopException:
             pass
-        except _ConnectionLostException, _ProtocolViolationException:
+        except Exception as e:
```
morinted: Should annotate this!

Issue #316, opened 2015-12-14
Update Linux install instructions to include workaround to get wxPython 3.0 under Ubuntu LTS and similar


tschulte:
I am having some issues compiling plover on my linuxmint box. The wxPython version shipped with linuxmint (and Debian/Ubuntu) is 2.8. But with commit 0eb8fb7514695d10b105d3c1720ce4f2cae177f9 you raised the required version of wxPython to 3.0. I had some issues trying to compile wxPython. I finally managed to compile and install it, but still I get 

```
/usr/local/lib/python2.7/dist-packages/wxPython_common-3.0.3.0-py2.7.egg/wx/_core.py:16633: UserWarning: wxPython/wxWidgets release number mismatch
  warnings.warn("wxPython/wxWidgets release number mismatch")
```

during the startup of plover.

Do you really need version 3.0? I tried resetting the changes made in that commit and it seems to work? If you don't need wxPython 3.0 it might be better to leave the required version at 2.8.

PaulBeaudet:
Had same issue, Agreed. I'm not sure why this was changed but it's only two lines of code it looks like. So it can be easily changed if 3.0 is not part of the future plans.

PaulBeaudet:
Also, if you haven't already @tschulte  I just change those two lines I think in setup.py and main.py.  everything works as of current.

morinted:
3.0 was required for OSX and it was pushed up on other systems to keep
everything inline. I don't mind reverting the version requirement, though
it might be harder for people who have 3.0 to know that >=2.8 applies to
them. I'd welcome a PR.

PaulBeaudet:
This seems to be coded into the source as opposed to just being required by the setup file

    WXVER = '3.0'
    if not hasattr(sys, 'frozen'):
        import wxversion
        wxversion.ensureMinimal(WXVER)

Now if the setup file already ensures a minimal for the separate OSes I'm not sure what this layer is for. But again I'm not super familiar, I just understand it is a single point where the OS does't matter. Unless of course the constant WXVER was set by OS type. Which I guess would be fine.

Willing to bet it will take a super unnecessarily long time for WX 3 to be included in debian repos. Not sure how setup-tools satisfies its dependencies but if it's through the repos the issue will likely persist into an actual release time frame.

morinted:
[...]

PR #318, opened 2015-12-18
Change wx version requirement for older Linux distros issue #316

PaulBeaudet:
This is really all it took to get things working in Mint 17. #316 I mentioned checking OS to only do this in Linux. Should unnecessary if it is true that setup-tools ensures 3.0 for Windows and OSX. If either manage to have 2.8 installed before the check everything should presumably still function as intended.

### DIFF
--- a/plover/main.py
+++ b/plover/main.py
@@ -8,7 +8,7 @@ import shutil
 import sys
 import traceback
 
-WXVER = '3.0'
+WXVER = '2.8'
 if not hasattr(sys, 'frozen'):
     import wxversion
     wxversion.ensureMinimal(WXVER)
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setuptools.setup(
         ],
         ':"linux" in sys_platform': [
             'python-xlib>=0.14',
-            'wxPython>=3.0',
+            'wxPython>=2.8',
         ],
         ':"darwin" in sys_platform': [
             'pyobjc-core>=3.0.3',


benoit-pierre:
That's not right: if 3.0 is required on Windows/OSX, than the code in `plover/main.py
` should ask for 3.0 on those platforms.

PaulBeaudet:
Sorry got caught up in other things. So we really need the "if Linux" statement in main huh? This is messy obviously, for the reasons you mention in the issue thread. Would rather have the discussion around the specifics anyhow.

Couple things I'm confused about, line 12
     if not hasattr(sys, 'frozen'):
What the consequence of sys having the attribute frozen? Is the test for minimal wx version skipped?

The other thing is line 14
Is ensure minimal version a test for plover to run or a test for the OS to run Plover. From what your saying I'm assuming its the second. So, in this case does that mean WX 2.8 is incompatible with the latest versions of OSX and Windows? Even if someone managed to have that version after set-up tools asked for 3.0? "wxversion.ensureMinimal('2.8')" would be that 3.0 would still be excepted right?

balshetzer:
I think there is some misunderstanding here.

On OSX and Windows, Plover ships as a fully packaged binary with all its
dependencies. There is no such thing as a user installing whichever version
of wx they want. Plover uses the one that it is shipped with. sys.frozen
means that the program is running from such a packaged bundle. There is
absolutely no use of setup.py for Plover other than on linux.

setuptools is for python libraries. It is not ideal for application
installation. In fact, it couldn't even fully install Plover in the past.
The only reason we use it on linux is because we never got around to
building the right thing for each platform like an ubuntu package. We are
not using it in the way you imagine we are. Therefore, we can't have
reasoning like "setuptools requires something, therefore its available"
because no such guarantees exist.

Finally, the ensureMinimal is just a good way to make sure that we have the
right version of wx that Plover needs. Otherwise, if the developer has the
wrong version in their dev environment (or a linux user hasn't installed
the right version) then they get a quick understandable error rather than
cryptic errors or weird problems while running the app.

benoit-pierre:
> On OSX and Windows, Plover ships as a fully packaged binary with all its dependencies. There is no such thing as a user installing whichever version of wx they want. Plover uses the one that it is shipped with. sys.frozen means that the program is running from such a packaged bundle. There is absolutely no use of setup.py for Plover other than on linux.

That's not entirely true anymore: 'setup.py' has been updated, and you can use it on OSX/Windows (even if most users will indeed want to use the pre-built binaries). It's incidentally also used on Max OS X to build said binary.

PaulBeaudet:
@benoit-pierre So then would it be better to create a legacy Linux .deb package base on a fork instead of messing with the source in trunk, that is going to "just work" in the case of most dev builds? Would rather spend time in a higher value area like packaging. As opposed to sticking in "if old Linux lines" and experimenting whether they work or not with other OSes. I really can only practically test in Linux which puts me at a slight disadvantage from a testing perspective here.

benoit-pierre:
Maybe. I did build some deb packages for Ubuntu at some point:
- https://launchpad.net/~benoit.pierre/+archive/ubuntu/ppa
- https://github.com/benoit-pierre/plover/tree/debian

PR #319, opened 2016-01-11
[RFC] theory customization

benoit-pierre:
There have been a number of posts on the Plover mailing list highlighting an interest in custom theories.

Unfortunately, Plover does not allow changing any aspect (letters/keys, undo stroke, orthographic rules, ...) of the theory beyond the dictionaries without patching the source.

This goal of this merge request is to get the discussion started on how we can address this issue.

Changes:
* all theory related settings are available through  `plover.theory`
* a new section has been added to the configuration

2 uses cases are supported:

* a theory directly supported by Plover is used, for example [Plover's default theory](https://github.com/benoit-pierre/plover/blob/8db2615f35ebe6b903509153ea5d3042d963ee89/plover/theory/default.py):
```
[Theory Configuration]
name = Plover
```
* a custom theory is used, e.g. Kinglet:
```
[Theory Configuration]
name = custom
file = kinglet.py
```
`file` must be the path to a valid Python file (which will be loaded relative to the configuration directory if not absolute).

An example that build on this for supporting the Kinglet theory is available [here](https://github.com/benoit-pierre/plover/compare/theory-customization...benoit-pierre:theory-customization-example).

GUI support can be added latter (note that support for changing the theory without restarting will necessitate some work).

Right now, I'd like first to get the name of all the needed theory settings right, before doing anything else. Also, even though the theory is a Python file, the goal is to have something as simple as possible, so users with limited knowledge can start from `plover.theory.default` as a template (without the need to know Python programming): the module does not need any imports, only defines some constants (relatively simple ones, except for the orthographic rules regular expressions).

Thoughts?


morinted:
I'm very happy that you've taken the first stab at this. I like your term, "theory". Though it is used when people debate Magnum and StenED, I do think it well maps what you are encompassing in your theory file; orthography, layout, key names. Some questions that have mulled over in my head since thinking about multi-lingual steno:

- The number bar in Plover feels like a workaround. Why not define strokes as "#ST-" instead of "12-" ? It's effectively just another "key" except that many keys apply to it. The only advantage that the current system offers is no need for dictionary entries for all number combinations (which in total would only be 1023 anyway... hardly anything). Have you given thought to this?
- We would need to go back to the serial machines and add support for other keys, and figure out how to map them. Stenoboard can use extra keys, the Treal has extra keys to the left and in between the thumb keys. Right now, Plover doesn't consume them, but in the interest of varied steno layouts, we should provide a way to map them for users. Kind of like you do with the qwerty map in this commit, but with the steno machine. E.g., the Treal could have a diagram showing key ID's, something like 0, 1, 2 from top-left to bottom-right, going in "steno-ish order" up to down, left to right. Then in the theory file you could provide a mapping for the various steno protocols.
    - We probably can't standardize between protocols because a steno machine could potentially have *n* keys to the left, right, bottom, and top of the layout. The Korean layout posted to the group had many rows, for example. The Chinese boards also have a different layout.
- Layout switching is a concern I have, for people who are multi-lingual and potentially multi-theoried. Take Achim, who will probably be porting over his German dictionaries to the German theory (same layout, different keynames), he would also like to switch between English and German while writing, maybe even midway through a sentence. Definitely hot-swapping of theories is a must, long term (though I appreciate the technical hurdles involved).
- Another thought I've had is pairing dictionaries to theories, so that either the dictionary defines its preferred theory or the user can pair certain theories to dictionaries.
- And what of space-handling? If we are really striving for multilingual, there are [languages that don't use spaces](http://linguistics.stackexchange.com/questions/6131/is-there-a-long-list-of-languages-whose-writing-systems-dont-use-spaces) at all the same as English. Heck, even programmers might want to turn off automatic spacing and leave it to their strokes to do the work.

All in all, lots to think about. Again, I'm really glad you took the first step toward making this happen. Of course, not all the questions above need an answer, as we could wait forever trying to build requirements without building anything, but at least now you know what's been going on inside my head.

benoit-pierre:
> - The number bar in Plover feels like a workaround. Why not define strokes as "#ST-" instead of "12-" ? It's effectively just another "key" except that many keys apply to it. The only advantage that the current system offers is no need for dictionary entries for all number combinations (which in total would only be 1023 anyway... hardly anything). Have you given thought to this?

I agree that it would be better to move it to a custom dictionary. I also think this is perfect use case for my [Python dictionaries](https://github.com/benoit-pierre/plover/commit/87349cda621aa9f3c39c7f2926ee1e32284694da).

> - We would need to go back to the serial machines and add support for other keys, and figure out how to map them. Stenoboard can use extra keys, the Treal has extra keys to the left and in between the thumb keys. Right now, Plover doesn't consume them, but in the interest of varied steno layouts, we should provide a way to map them for users. Kind of like you do with the qwerty map in this commit, but with the steno machine. E.g., the Treal could have a diagram showing key ID's, something like 0, 1, 2 from top-left to bottom-right, going in "steno-ish order" up to down, left to right. Then in the theory file you could provide a mapping for the various steno protocols.

Agreed.

> - We probably can't standardize between protocols because a steno machine could potentially have *n* keys to the left, right, bottom, and top of the layout. The Korean layout posted to the group had many rows, for example. The Chinese boards also have a different layout.
> - Layout switching is a concern I have, for people who are multi-lingual and potentially multi-theoried. Take Achim, who will probably be porting over his German dictionaries to the German theory (same layout, different keynames), he would also like to switch between English and German while writing, maybe even midway through a sentence. Definitely hot-swapping of theories is a must, long term (though I appreciate the technical hurdles involved).

Yes, a number of settings will need to be moved around, the configuration could look like this:

```
[Theory Configuration]
name = custom
file = kinglet.py

[Plover]
stentura_keymap = ...
keyboard_keymap = ...
dictionary_file1 = plover_main.json
dictionary_file2 = plover_commands.json
dictionary_file3 = plover_user.json
space_placement = Before Output
space_char = " "

[Kinglet]
stentura_keymap = ...
keyboard_keymap = ...
dictionary_file1 = kinglet.py
dictionary_file2 = kinglet_commands.json
dictionary_file3 = kinglet_user.json
space_placement = After Output
space_char = ""
```

So a theory should also provide:
* a name
* a set of defaults for those settings common to all theories (list of dictionaries, space character/placement, default keymaps for supported machines...)
* an optional set of custom settings

> - Another thought I've had is pairing dictionaries to theories, so that either the dictionary defines its preferred theory or the user can pair certain theories to dictionaries.

Yes. To be frank I'm also not a fan of the way the standard dictionaries are handled, with the copy to the user configuration. I wonder if it would not make more sense to load them directly from the Plover package (and mark them as read-only). Of course the user should have the option to make a copy to it's configuration directory, if needed. This way one can get an updated dictionary automatically. And since it's always possible to override an entry in a custom dictionary with precedence, are writable theory dictionaries really needed? I.e. is there a use case where there is a need to delete an entry from such a dictionary? Maybe for word boundary conflicts?

Support for marking some dictionaries read-only is also something I'm going to need for Python dictionaries.

> - And what of space-handling? If we are really striving for multilingual, there are [languages that don't use spaces](http://linguistics.stackexchange.com/questions/6131/is-there-a-long-list-of-languages-whose-writing-systems-dont-use-spaces) at all the same as English. Heck, even programmers might want to turn off automatic spacing and leave it to their strokes to do the work.

It's always possible to use `{^bla}` for all entries. We can also do something similar to your Caps/Title/Lowercase/Snake merge request and allow setting the space character to the empty string.

4n08o:
Would this also cover different languages (meaning software, dictionaries, and language specific steno machines)?

There is a first draft with regard to German mentioned in the plover group:
https://groups.google.com/forum/m/#!topic/ploversteno/MYNlSMD68Qc
Please refer to the 5th entry from Mike.

morinted:
Yes, @4n08o this is inspired by these and similar needs by many people.

@benoit-pierre:

- I'm very interested in the Python dictionaries. Do you have any examples of how you would use them? Have you made some prototypes?
- As for the read-only dictionaries, I'm all right with that, and I'd like to discourage people changing the default dictionaries, however I definitely think that you should be able to disable these dictionaries or remove them (maybe remove, then when you add there's a list of built-in dictionaries?) I guess we'll have to work through the user interface so that when you add a dictionary, you select the theory from a list of installed/configured theories, or add a new one right then.

I wonder if the barrage of new settings will confuse users who just want to use the default dictionary with their own custom English dictionary.

Hm, I'm struggling with the terminology of "theories" versus what this pull request is suggesting. Maybe we could rename them to "engines" or something of the like. Engine, system, chording system, layout (which doesn't well encompass a whole system like Kinglet or Mecatype). That all being said, I don't think we should call it "Plover" theory as in this pull request. I believe a better name would be "English stenography" or something of the like. Plover theory is just the default dictionary and what Learn Plover! teaches, similar to StenEd. I imagine that Phoenix users would be confused if when adding their dictionary they had to select "Plover" theory. "English stenography" or just "Stenography" is much more clearly "STKPWHRAOEUFRPBLGTSDZ" in my opinion. Then Kinglet makes sense as its own entity. Would Kinglet be a py dictionary and a *system/theory* file, or not need a dictionary because it's all contained by the theory/system/engine/layout/driver/core file?

benreynwar:
At the moment the flow in Plover is 

-> Machine -> Translator -> Formatter -> OutputHelper ->

The Machine -> Translator link is done with Strokes
The Translator -> Formatter link is done with (undo, do, state) structure containing translations.
The Formatter -> OutputHelper is one with strings or combos or commands

Plover could be made considerably more flexible by allowing one to switch out these different components and by making the method of communication between them theory and machine agnostic.

You could then have plugins for plover, implemented as independent python packages, that make additional machines, translators, formatters or output helpers available.  Ideally these plugins would also hook into the plover GUI to allows the configuration of this additional functionality.

benoit-pierre:
> - I'm very interested in the Python dictionaries. Do you have any examples of how you would use them? Have you made some prototypes?

Some examples (orthographic theories): [Kinglet (English)](https://github.com/benoit-pierre/plover/blob/kinglet/kinglet.py), [Mecatipia (Spanish)](https://github.com/benoit-pierre/plover/blob/mecatipia/mecatipia.py), [Melani (Italian)](https://github.com/benoit-pierre/plover/blob/melani/melani.py), [Shelton Syllabic / Jackdaw (English)](https://github.com/benoit-pierre/plover/blob/shelton-syllabic/shelton_syllabic.py).

Here is another, very simple, example:

```python
MAXIMUM_KEY_LENGTH = 2

SHOW_STROKE_STENO = 'STR*'

def lookup_translation(key):
    assert len(key) <= MAXIMUM_KEY_LENGTH
    if SHOW_STROKE_STENO != key[0]:
        raise KeyError
    if 1 == len(key):
        return ' '
    return key[1]
```

If you stroke `STR*`, then the next stroke will be shown verbatim (untranslated), e.g. `-T STROEBG TP-R KW-GS STROEBG KR-GS S STR* STROEBG` outputs: ` the stroke for "stroke" is STROEBG`

> - As for the read-only dictionaries, I'm all right with that, and I'd like to discourage people changing the default dictionaries, however I definitely think that you should be able to disable these dictionaries or remove them (maybe remove, then when you add there's a list of built-in dictionaries?) I guess we'll have to work through the user interface so that when you add a dictionary, you select the theory from a list of installed/configured theories, or add a new one right then.

> I wonder if the barrage of new settings will confuse users who just want to use the default dictionary with their own custom English dictionary.

> Hm, I'm struggling with the terminology of "theories" versus what this pull request is suggesting. Maybe we could rename them to "engines" or something of the like. Engine, system, chording system, layout (which doesn't well encompass a whole system like Kinglet or Mecatype). That all being said, I don't think we should call it "Plover" theory as in this pull request. I believe a better name would be "English stenography" or something of the like. Plover theory is just the default dictionary and what Learn Plover! teaches, similar to StenEd. I imagine that Phoenix users would be confused if when adding their dictionary they had to select "Plover" theory. "English stenography" or just "Stenography" is much more clearly "STKPWHRAOEUFRPBLGTSDZ" in my opinion.

OK, so `s/theory/system`. And the default system is `English Stenotype`.

> Then Kinglet makes sense as its own entity. Would Kinglet be a py dictionary and a *system/theory* file, or not need a dictionary because it's all contained by the theory/system/engine/layout/driver/core file?

Both, a system and a (python) dictionary.

I rewrote the branch completely: the implementation now uses `setuptools/pkg_resources` support for entry points to handle plugins.

A plugin is an egg file, placed in the plugins directory (which must be located in the same directory as the configuration), with some specific entry points: systems and python dictionaries.

You can see an example for Kinglet [here](https://github.com/benoit-pierre/plover-kinglet). If you look at [setup.py](https://github.com/benoit-pierre/plover-kinglet/blob/master/setup.py), you'll see that both a system and a dictionary are exposed by the plugin. Build the egg with `python2 setup.py bdist_egg`, stick it in `~/.local/share/plover/plugins`, and you can use Kinglet with the following entries in the configuration:

```
[Dictionary Configuration]
dictionary_file1 = plugins:Kinglet

[System Configuration]
name = Kinglet
```

I even managed to make it work with PyInstaller... (Note: you need at least version 3.1 to be able to package Plover with WxWidgets 3.0).

jeremy-w:
Allowing binding additional keys could be useful even in the context of the standard Plover theory when used with an nkro keyboard. It would allow basically switching theory "layers" without worrying so much about clobbering existing dictionary entries, so for example, you could throw a group of app-specific commands as chords with the Z key, and not have any worry of accidentally invoking them or needing to work around existing outlines in the rest of the theory.

benoit-pierre:
Closing this, initial support for plugins is coming with #615.

PR #324, opened 2016-01-15
Allow engine to be passed to GUI on initialization.

benreynwar:
This is so that the engine can be customized from outside the GUI.
Using this one can make use of the plover GUI stuff, but use a modified engine, without creating a fork of the project.
I'm using it at https://github.com/benreynwar/stenoprog/blob/73e8c4423dcf228d039ee0bee73f1ac8e044d4ab/launch.py#L142

There's quite likely a more elegant way of doing this, but this one works.

### DIFF
--- a/plover/gui/main.py
+++ b/plover/gui/main.py
@@ -37,8 +37,9 @@ from plover import __license__
 class PloverGUI(wx.App):
     """The main entry point for the Plover application."""
 
-    def __init__(self, config):
+    def __init__(self, config, engine=None):
         self.config = config
+        self.engine = engine
         # Override sys.argv[0] so X11 windows class is correctly set.
         argv = sys.argv
         try:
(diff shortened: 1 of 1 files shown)


morinted:
@benoit-pierre I see nothing wrong with this code and it isn't breaking any tests. Do you agree that we could merge it?

benoit-pierre:
I think it would be better if the 'engine' argument was mandatory, as right know this make the code a little worse.

benreynwar:
I think the work that @benoit-pierre is doing to make plover play more nicely with alternative theories will make this unnecessary, so I think that probably this should not be merged.

Better to wait until a nicer solution exists.

benoit-pierre:
I think if the argument is made mandatory, it's a step in the right direction: reduce some of the inter-dependencies between modules.

IMHO, the GUI should not depend on the `plover.app` module, and the `init_engine/reset_machine/update_engine` functions should be made methods of the StenoEngine class.

morinted:
Closing for now, you can reopen if you want to further decouple the GUI as benoit-pierre mentioned.

PR #326, opened 2016-01-19
Add PLOVER:PAUSE:XX command.  Useful for adding brief pauses in seque…

benreynwar:
…nces of commands.

### DIFF
--- a/plover/gui/main.py
+++ b/plover/gui/main.py
@@ -10,6 +10,8 @@ resumes stenotype translation and allows for application configuration.
 
 import sys
 import os
+import re
+import time
 import wx
 import wx.animate
 from wx.lib.utils import AdjustRectToScreen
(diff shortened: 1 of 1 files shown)


morinted:
The problem with this is that the pause code blocks Plover from receiving input.

If I have a pause of two seconds, I can't continue writing in between those two seconds. I think that would need to be supported for us to merge. When you included the pause as a key to be pressed, was this a non-issue?

Further, I'm sort of tempted to ask you to make the pause a key, because then we could "pause" after hitting the shift key and such. Let me explain, if I had a stroke that mapped to "{#Shift_L(pause:1000)}", then I could use that stroke, then click within a second, and get a shifted-click, something currently not possible in Plover.

Thanks for submitting the PR so quickly!

morinted:
I'm still interested in this PR, but am just decluttering the list. Please, feel free to reopen if you can address my concerns. Mainly:

- Pausing shouldn't block input
- Wishlist feature: allow users to "hold down" keys with pauses. I was thinking that you could express {#a(hold10000)} and that would hold down the "a" key for 10 seconds.

Let me know if you have any questions or need help with the code.

benreynwar:
Sounds good.  I'm not allowing myself to do plover-related programming,
using non-chorded typing so it'll be a while before I'm fast enough to get
back to this.

PR #330, opened 2016-01-26
Fix stentura tests

benoit-pierre:
Fix the tests that got broken with the recent changes to timeout handling.

I don't have a Stentura, but only added some sanity check to the read code, so hopefully it still works with a real machine.

### DIFF
--- a/plover/machine/stentura.py
+++ b/plover/machine/stentura.py
@@ -443,9 +443,11 @@ def _read_data(port, stop, buf, offset, num_bytes):
     _TimeoutException: If the timeout is reached with no data read.
 
     """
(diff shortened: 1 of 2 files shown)


morinted:
I'll create a build and see if @stanographer would try it out for us.

morinted:
Build is available on [my Google Drive](https://drive.google.com/file/d/0B_4v0_ER8UlwZTVsSTlPN0VnQm8/view?usp=sharing) for a little while.

morinted:
Okay, apparently Stanley only has TX Bolt available. Does anyone have a Stentura (Stenograph) machine to try out this OS X build with?

PR #340, opened 2016-02-08
various GUI related fixes

benoit-pierre:
The most important being the first one: `gui/keyboard_config: fix mapping editing`

For some reason on Windows, after changing a mapping, the corresponding row would not update...


morinted:
Tested on Mac as well, works well. If you could just update the comment language then merge is go.

benoit-pierre:
Done.

PR #341, opened 2016-02-08
Rework keyboard suppression

benoit-pierre:
Cleanup the keyboard machine code, now that all `KeyboardCapture` implementations support suppression (which also mean we can get rid of the dependency on `KeyboadEmulation`).

Rework the way machines handle suppression to:
- avoid the need for testing for a particular class instance to know if suppression is needed
- suppress all the key down events when post-suppressing a stroke: previously, when toggling Plover on with a stroke, key auto-repeat events were not suppressed (those can happen when being a little heavy handed and/or configuring X11 with a low delay for triggering auto-repeating)

Also change the `KeyboardCapture` implementation so list of suppressed keys is not set once and for all at initialization, which will make it easier to share capture instances (e.g. in the keyboard config dialog) and will also be used with the upcoming rework of machine keymaps.

Note: tested on Linux and with a Windows 10 virtual machine.

### DIFF
--- a/plover/app.py
+++ b/plover/app.py
@@ -22,7 +22,6 @@ import plover.config as conf
 import plover.formatting as formatting
 import plover.steno as steno
 import plover.machine.base
-import plover.machine.sidewinder
 import plover.translation as translation
 from plover.exception import InvalidConfigurationError,DictionaryLoaderException
 from plover.machine.registry import machine_registry, NoSuchMachineException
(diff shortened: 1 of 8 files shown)


morinted:
I tried this out and went through most of the commits, all is good and I'd merge now, but I just need to take a closer look at 43a36a9 to internalize it.

benoit-pierre:
This can still be improved a little if we want to handle keys that are not mapped to a steno key but suppressed to prevent miss-strokes, see new commit.

benoit-pierre:
Note: let me know when you want to merge and I'll rewrite the fixups.

morinted:
Oh, that's neat. So my stroke to toggle in NKRO is "T-LG", or "wol" in Qwerty. This means that if I write "wol;" (T-LGS), there is no toggle. BUT if I write "wol." (T-LG + the period key), you actually backspace that period as well. Very interesting. I think I like it.

Yeah, I think I'm ready for a clean up and merge.

benoit-pierre:
Done.

PR #342, opened 2016-02-09
Simplify OS X dev setup

jeremy-w:
- Replace manual instructions to `pip install` a list of packages with requirements.txt
- Replace manual setup instructions with a bootstrap script
- Rewrite the README to point people at the bootstrap script first and supply additional information as a "For the More Curious" section.

The largest change in the README is to discourage people from using a virtual environment for developing Plover. It ends up not being all that virtual in the end anyway thanks to wxPython.

This is an improvement, but it's not sufficent; I still don't have a functional dev environment myself. I can build an app bundle, but it fails during launch. Running from the commandline with `arch -32 /usr/local/bin/python -m plover.app` runs afoul of the `objc` package not having a 32-bit slice.

32-bit apps under OS X are going the way of the dodo, and the whole toolchain has moved to reflect that, unfortunately. :\

### DIFF
--- a/osx/requirements.txt
+++ b/osx/requirements.txt
@@ -1,0 +1,16 @@
+# General requirements
+appdirs  # locating platform-varying libraries like Documents
+pyserial  # connecting to steno machines that use a serial protocol
+mock  # unittest.mock for Python 2.6 through 3.2; in stdlib with 3.3+
+
+
+# Darwin-specific requirements
+hidapi; sys_platform == 'darwin'  # supporting the Treal's USB protocol
+pyobjc-core; sys_platform == 'darwin'  # Notification Center
+pyobjc-framework-quartz; sys_platform == 'darwin'  # accessing the CGEvent API
+py2app; sys_platform == 'darwin'  # building the Plover.app bundle
+
+# Later versions of setuptools appear to lead to `py2app` not bundling
+# the required `six` package with the app. Pinning to 19.2 fixes this.
+# We should revisit this periodically by unpinning and checking in Travis.
+setuptools == 19.2; sys_platform == 'darwin'
--- a/osx/README.md
+++ b/osx/README.md
@@ -1,34 +1,50 @@
 # Plover Development on OS X
 
+## Getting Started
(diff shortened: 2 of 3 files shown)


morinted:
I have to try going through this again... I only started OS X dev setup 6 months ago, it's odd that you are having all these difficulties.

jeremy-w:
Yeah, I'm surprised at all the problems, as well. I think something might be wonky with the pip being pulled in by various invocations based on weird output, but I'm unsure how to wipe the slate clean. :\

Example weird output: "It's up to date, but it's out of date, and you should update it."

```
> /usr/local/bin/pip install --upgrade pip
Requirement already up-to-date: pip in /usr/local/lib/python2.7/site-packages
You are using pip version 1.4.1, however version 8.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

jeremy-w:
My issues aside, the bootstrap script and requirements.txt approach should make things easier for people who don't have totally messed up Python environments. :)

morinted:
Yup, you'll have to wait for me to set up a VM and try it out from scratch to test it proper.

jeremy-w:
Sounds great! Thanks. :)

osx/requirements.txt
```diff
@@ -0,0 +1,13 @@
+# General requirements
+appdirs  # locating platform-varying libraries like Documents
+pyserial  # connecting to steno machines that use a serial protocol
+mock  # unittest.mock for Python 2.6 through 3.2
+
+
+# Darwin-specific requirements
+# Use Ted's fork of hidapi so that we get a Universal extension, not just x86-64.
+git+https://github.com/morinted/cython-hidapi.git#egg=hidapi; sys_platform == 'darwin'
```
jeremy-w: We can switch back to normal hidapi after #356 gets merged.

jeremy-w:
Updating in light of having merged #356 and switched to 64-bit only builds.

jeremy-w:
All-right, updated for the new, 64-bit–only world!

jeremy-w:
Now very much squashed.

PR #344, opened 2016-02-10
Add Mac Notification Center support

morinted:
Use Notification Center if available on Mac.

Instead of the wx notifications, which are pretty awful looking on OSX, I changed to notification center, which acts like other applications. The biggest limitation is that we can't expire notifications, but OSX users should be used to that. Here is the hover view and the history view for the notification:

screen shot 2016-02-10 at 12 37 00 pm

screen shot 2016-02-10 at 12 47 54 pm

### DIFF
--- a/plover/gui/log.py
+++ b/plover/gui/log.py
@@ -1,4 +1,3 @@
-
 import sys
 from plover import log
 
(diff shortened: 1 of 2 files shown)


morinted:
@benoit-pierre if I could get your feedback on this implementation, I understand you can't test it, but feel free to merge it if it's good.

benoit-pierre:
What happens if you get multiple notifications?

morinted:
The most recent one replaces the older one in that foreground notification (the floating one), then they stack up in the actual sidebar --> same thing happens with Slack and Discord messages and stuff.

Noticed that you are limited to 256 Bytes, though. Anything longer gets truncated. In most cases this still seems okay.

screen shot 2016-02-10 at 1 36 39 pm

benoit-pierre:
OK.

PR #348, opened 2016-02-11
Reduce Tréal CPU usage

morinted:
The Tréal file was set to use non-blocking reads from the hidapi. But that code was in a while loop, and so it was called infinitely, causing very high CPU usage. I've switched over to a blocking version of the call, which actually gives Plover a break in the while loop.

The effects have only been positive so far. I feel a little silly for not noticing this high CPU usage before in the months using Plover with the Tréal 😕

### DIFF
--- a/plover/machine/treal.py
+++ b/plover/machine/treal.py
@@ -89,7 +89,7 @@ class Stenotype(StenotypeBase):
                     self._machine.open(VENDOR_ID, 1)
                 else:
                     self._machine = hid.device(VENDOR_ID, 1)
-                self._machine.set_nonblocking(1)
+                self._machine.set_nonblocking(0)
             except IOError as e:
                 log.info('Treal device not found: %s', str(e))
                 log.warning('Treal is not connected')


morinted:
For a comparison, the non-blocking did about 3.5 million calls in 6.5 seconds of Plover launching, this one does about 300 calls.

benoit-pierre:
What about when you stop the capture? Using non-blocking mode might have been a (ugly) way to avoid getting stuck in the read call and handle being able to stop the capture anytime.

morinted:
The loop still runs dozens of times a second, so any stop call would effectively stay the same.

I just tried running through the motions of the machine connection, going from on to off, disconnecting the machine while connected, reconnecting it, changing machine types back and forth. All seems to work the same as it did before (after a physical disconnect, Plover needs to be reset by either restarting Plover or switching between machines.) There are no exceptions being thrown (I'd almost like to detect when the Tréal is disconnected…)

But, barring new features, everything seems fine.

PR #349, opened 2016-02-11
Add command to run Plover in development on OS X

morinted:
Frustrated with having to build Plover between code changes, I finally took the time to figure it out. @jeremy-w, you pay be interested in this change.


jeremy-w:
Oh yes, I've been using sudo to circumvent that. But my 32-bit build dies with a 64-bit only pyobjc library. :\

Issue #261, opened 2015-09-25
Lack of documentation


morinted:
- Tasks
  - [ ] Lint the codebase using `flake8`
- Contributor documentation
  - [ ] Editing guide for the dictionary for dictionary contributors
  - [x] Add a Contributing.md (via PR #351)
  - [ ] Point contributors to our lint script in README / CONTRIBUTING
  - [ ] A couple diagrams explaining the hierarchy of the system would be useful for new developers getting started.
    - There could be a global diagram, and also a section for users who just want to work on the dictionary (which is less technical and doesn't involve Python)
- User documentation
  - [x] Converting the Plover Guide to Markdown (via PR #352)
  - [ ] Updating the Plover Guide:
    - Remove outdated information
    - Add a header structure to make documentation easier to reference. Link to material in Learn Plover! where appropriate.
    - Section ideas: dictionary hacking, special commands, guides on various Plover utilities (will be required with dictionary manager and suggestions window)

jeremy-w:
@morinted I can tackle these.

I'm not clear on the first item, though:  "Editing guide for the dictionary for contributors" Could you elaborate on this?

morinted:
Just what we do and don't want for the default dictionary → for example, adding animal names would be nice, but adding weird, special commands that are platform specific, not so much. Maybe link out to Stenodict now that it exists. Also, we have a commands.json where all commands should go. Finally, preserve the alphabetical order of the dictionary (I think it's ordered by translation)

jeremy-w:
Perfect, thanks!

morinted:
Here's my work in progress changelog for the next release:


Features
====================

- **Full Unicode support** on Windows, OS X, and Linux. Now you can use a non-Latin alphabet, including extended symbols such as emoji 💥
- **New lookup window**, to look up strokes for a word. Use {PLOVER:LOOKUP}
- **Configure whether space is outputted at the beginning or end** of the current stroke (default/current is like " add", new option outputs "add ")
- **New Dictionary Editor**! Lets you view and edit strokes in your dictionaries. You can filter by stroke or definition, and add, edit, and delete entries in your JSON and RTF dictionaries.
- **New Suggestions window**. The suggestions window is a training tool that watches what you are writing, and gives other strokes that have the same output. It's great for discovering briefs.
- **New commands:**
    + Repeat last stroke: repeats the last stroke exactly, useful for things like arrow keys to avoid having to do a complex stroke multiple times: `{*+}`
[...]

PR #352, opened 2016-02-12
Add Markdown version of Plover Guide

jeremy-w:
This was just an automatic conversion via
`pandoc -f latex -t markdown plover_guide.tex -o plover_guide.md`,
but it appears to have got the job done.

[#261]

### DIFF
--- a/doc/plover_guide.md
+++ b/doc/plover_guide.md
@@ -1,0 +1,286 @@
+Quick Start
+===========
+
+Installation
+------------
+
+The latest release of <span>*Plover*</span> can be downloaded from
+<https://launchpad.net/plover> as a compressed <span>`.tar.gz`</span>
+archive that must be extracted. Among the extracted files will be a
+README.txt file. Follow the directions in this file for installing and
+running <span>*Plover*</span>.
+
+Starting the Program
+--------------------
+
+Once installed, <span>*Plover*</span> can be run from either the
+application menu or from a <span>[terminal
+window](https://help.ubuntu.com/community/UsingTheTerminal)</span> (<span>`plover`</span>
+and then <span>`ENTER`</span>). If a stenotype machine other than the
+default is used, set the machine type in the configuration dialog.
+
+Configuration {#config}
+-------------
+
+<span>*Plover*</span> comes with a default configuration that should
+work for most setups that use a keyboard as the stenotype machine.
+<span>*Plover*</span> can be configured by clicking the
+<span>`Configure...`</span> button. Within the configuration dialog, the
+user can select which stenotype machine to use, which dictionary file to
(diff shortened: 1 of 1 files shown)


jeremy-w:
Wow, the guide looks pretty out of date.

Is there any value in having this, either tex or md, vs just a pointer to *Learn Plover!* ?

morinted:
Fair point, it might be worth referencing Learn Plover instead. I was in
the process of rewriting the Plover Guide but there is a LOT of content
added.

jeremy-w:
Yeah, #261's section on user docs makes clear that the Plover Guide is basically going to get completely rewritten. I think there probably is value in having something that explains the basic windows and features of Plover, but for things like the how and why of dictionary hacking, we should probably just point to Learn Plover.

stenoknight:
Main use of the Plover Guide as opposed to Learn Plover is the extensive
list of names for the complete character set, though that's obviated
somewhat by the new unreleased build of Plover, which allows for pasting
symbols and characters directly into the dictionary update window rather
than being forced to identify them by their name.

On Fri, Feb 12, 2016 at 12:22 AM, Jeremy W. Sherman <
notifications@github.com> wrote:

> Yeah, #261 <https://github.com/openstenoproject/plover/issues/261>'s
> section on user docs makes clear that the Plover Guide is basically going
> to get completely rewritten. I think there probably is value in having
> something that explains the basic windows and features of Plover, but for
> things like the how and why of dictionary hacking, we should probably just
> point to Learn Plover.
>
> —
> <https://github.com/openstenoproject/plover/pull/352#issuecomment-183184406>
> .
>



-- 
Mirabai Knight, CRC, RDR
StenoKnight CART Services
917 576 4989
mkk@stenoknight.com
http://stenoknight.com

morinted:
Maybe we out to turn the Plover Guide into a section in the GitHub wiki? A series of pages instead of one giant document? Learn Plover is great for theory, but wouldn't encapsulate all the information we have on program specifics and technical limitations with current releases and such.

jeremy-w:
@stenoknight That's useful information. I haven't gotten to where I need that yet, so I didn't realize that was the key reference material in there for expert users.

@morinted I think tracking the docs in the repo is good vs a wiki. Nothing's stopping us from splitting it up into a collection of Markdown documents that link to each other.

I have a much better idea of the role the guide is supposed to play now. I think we should merge this as a starting point; at the least, it makes it easy for someone to link to and look up that symbol chart, without having to point at raw TeX or a PDF, so it's an incremental improvement over what we have now. We can grow it from here in later PRs.

jeremy-w:
Merging. Further improvements to the Markdown content will happen in later PRs.

PR #353, opened 2016-02-12
Extend suppressed keys support

### LINKED ISSUE #343: Cannot configure keyboard to use non-alphanumeric keys
Currently am trying to map Plover to the Kinesis Advantage. The Advantage changes the layout of a standard keyboard for ergonomic reasons, moving most of the pinkie keys to the thumbs. For this reason, it's not possible to hit the 'c', 'v', 'n', and 'm' keys with the thumbs. The most natural keys to remap them to would be 'backspace', 'delete', 'enter', and 'space', but I am not able to map the first 3 of these keys using the Keyboard Configuration GUI. It will not register 'backspace' or 'delete', and 'enter' will simply confirm the existing selection.

benoit-pierre:
Fix #343. Tested on Linux and Windows 10 VM.

Note the missing `Insert` on Mac: it's not supported in the `KEYNAME_TO_KEYCODE` dict above. Because Mac keyboard don't have an insert key?


morinted:
Correct, Mac has no Insert key. A big painbody when remote desktoping, I'll add. Looking forward to testing this out.

jeremy-w:
I have a keyboard with an Insert key hooked up to my Mac. Karabiner's EventViewer app reports that pressing down and then releasing the Insert key does this:

```
eventType:KeyDown         code:0x72       name:Help            flags:Fn                                 misc:KeyCode::HELP	characters:    
eventType:KeyUp           code:0x72       name:Help            flags:Fn                                 misc:KeyCode::HELP	characters:    
```

morinted:
If you add `0x72`, I can try it tomorrow with an external USB keyboard.

benoit-pierre:
There is already an entry for `0x72` in `KEYNAME_TO_KEYCODE`: `'Help': [114]`.

morinted:
Erp, I ran into an issue on OS X. Let me git bisect to see the cause.

Apparently c320d9ee caused crashes on OSX that I hadn't encountered due to limited testing. The issue, I think, comes from wx trying to handle a suppressed key, because we modify the event to something that it doesn't understand.

To reproduce my exceptions/crashes:

## Case 1: Exception

- Run plover
- Open "Configuration" Dialog
- Hit the "a" key (keycode 0, translation 'is')
- [Exception is logged to console](https://gist.github.com/morinted/928f1f08b4a4807887b1#file-exception-txt)

## Case 2: Crash

- Run plover
- Open "Configuration >> Machine >> NKRO Configure... >> Double-click "a q"
- Hit the "a" key
- The dialog updates, showing "-a" at the bottom, the app immediately crashes.
- [Exceptions are logged to console](https://gist.github.com/morinted/928f1f08b4a4807887b1#file-crash-txt)
- Seems like the output is affecting the # of exceptions, "a", output "is" gives less exceptions than "d", output "about".

morinted:
Okay, barring the exceptions (we might just leave it there for now since this bug has existed for a little while), the Mac didn't let me suppress the extended keys. I needed to modify the keyboard suppression a little bit:

```diff
diff --git a/plover/oslayer/osxkeyboardcontrol.py b/plover/oslayer/osxkeyboardcontrol.py
index 47ec470..d6005b9 100644
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -245,7 +245,12 @@ class KeyboardCapture(threading.Thread):
                 MY_EVENT_SOURCE_ID):
                 return event
             # Don't intercept the event if it has modifiers.
-            if CGEventGetFlags(event) & ~kCGEventFlagMaskNonCoalesced:
+            flags = CGEventGetFlags(event)
+            if (flags & kCGEventFlagMaskShift or
+                    flags & kCGEventFlagMaskControl or
+                    flags & kCGEventFlagMaskAlternate or
+                    flags & kCGEventFlagMaskCommand or not
+                    flags & kCGEventFlagMaskNonCoalesced is kCGEventFlagMaskNonCoalesced):
                 return event
             keycode = CGEventGetIntegerValueField(event, kCGKeyboardEventKeycode)
             key = KEYCODE_TO_KEY.get(keycode)
```

If you could add that change, then this should be good to merge. I never got the "Insert/Help" key working, I think Mac reserves it for "Help" functionality (the question mark cursor).

benoit-pierre:
So this is the documentation I could find on the flags:

```c
 enum CGEventFlags : UInt64 {
    case MaskAlphaShift      // Indicates that the Caps Lock key is down for a keyboard, mouse, or flag-changed event.
    case MaskShift           // Indicates that the Shift key is down for a keyboard, mouse, or flag-changed event.
    case MaskControl         // Indicates that the Control key is down for a keyboard, mouse, or flag-changed event.
    case MaskAlternate       // Indicates that the Alt or Option key is down for a keyboard, mouse, or flag-changed event.
    case MaskCommand         // Indicates that the Command key is down for a keyboard, mouse, or flag-changed event.
    case MaskHelp            // Indicates that the Help modifier key is down for a keyboard, mouse, or flag-changed event. This key is not present on most keyboards, and is different than the Help key found in the same row as Home and Page Up.
    case MaskSecondaryFn     // Indicates that the Fn (Function) key is down for a keyboard, mouse, or flag-changed event. This key is found primarily on laptop keyboards.
    case MaskNumericPad      // Identifies key events from the numeric keypad area on extended keyboards.
    case MaskNonCoalesced    // Indicates that mouse and pen movement events are not being coalesced.
} 
```

Except for `MaskNonCoalesced`, I would think we still don't want to suppress any event with one of the flags set. Which one is it we need now? Which keys are considered extended? And does that mean some of  the key events can only be obtained with a combo? Home/End maybe? And it it different depending on the type of keyboard (mac vs regular) used?

morinted:
The settings I gave worked pretty well for most cases. In particular, the `flags & kCGEventFlagMaskNonCoalesced is kCGEventFlagMaskNonCoalesced` line prevents Plover from an infinite loop where it processes its own key presses.

As for all the others, home and end and the arrows key worked on the keyboard I was trying. The internal keyboard only really has keys for backspace and up down left and right. I guess we could add add the others without any issue.

morinted:
[More information on the various flags](https://developer.apple.com/library/mac/documentation/Carbon/Reference/QuartzEventServicesRef/index.html#//apple_ref/swift/enumelt/CGEventFlags/c:@E@CGEventFlags@kCGEventFlagMaskNumericPad).

So for the arrow keys, for example, on the Mac, both SecondaryFn and Numeric flags are set. This is the set that gave me the best results:

```
            flags = CGEventGetFlags(event)
            if (flags & kCGEventFlagMaskCommand or
                    flags & kCGEventFlagMaskShift or
                    flags & kCGEventFlagMaskAlternate or
                    flags & kCGEventFlagMaskControl or
                    flags & kCGEventFlagMaskHelp or
                    flags & kCGEventFlagMaskNonCoalesced is not kCGEventFlagMaskNonCoalesced):
                return event
```

If you can add those, I think we are good to merge.

benoit-pierre:
> [More information on the various flags](https://developer.apple.com/library/mac/documentation/Carbon/Reference/QuartzEventServicesRef/index.html#//apple_ref/swift/enumelt/CGEventFlags/c:@E@CGEventFlags@kCGEventFlagMaskNumericPad).

Yeah, it's the same info I found.

> So for the arrow keys, for example, on the Mac, both SecondaryFn and Numeric flags are set.

Which does not make sense for `SecondaryFn` according to the documentation, as you're not pressing the Fn key to get those right? My concern is that since the rest of the code completely ignore the modifier, could that get us in trouble with some Fn+key combination being handled as if only the key was pressed (without the Fn modifier combo)?

The bit about `kCGEventFlagMaskNonCoalesced` also worries me: so when we generate events with the `KeyboardEmulation` class, the flag is not set, but with standard events it is? Seems like a kludge, are we sure we'll never get valid system events with the flag not set? Shouldn't this already be handled by the custom event source (`MY_EVENT_SOURCE`) ?

And let's make the code more explicit and document the non-obvious behavior:

```diff
diff --git i/plover/oslayer/osxkeyboardcontrol.py w/plover/oslayer/osxkeyboardcontrol.py
index 731e0f9..fe4deb0 100644
--- i/plover/oslayer/osxkeyboardcontrol.py
+++ w/plover/oslayer/osxkeyboardcontrol.py
@@ -21,6 +21,8 @@ from Quartz import (
     kCGEventFlagMaskControl,
     kCGEventFlagMaskNonCoalesced,
     kCGEventFlagMaskShift,
+    kCGEventFlagMaskSecondaryFn,
+    kCGEventFlagMaskNumericPad,
     kCGEventKeyDown,
     kCGEventKeyUp,
     kCGEventSourceStateID,
@@ -240,8 +242,9 @@ class KeyboardCapture(threading.Thread):
             if (CGEventGetIntegerValueField(event, kCGEventSourceStateID) ==
                 MY_EVENT_SOURCE_ID):
                 return event
-            # Don't intercept the event if it has modifiers.
-            if CGEventGetFlags(event) & ~kCGEventFlagMaskNonCoalesced:
+            # Don't intercept the event if it has modifiers, allow
+            # Fn and Numeric flags so we can suppress the arrow keys.
+            if CGEventGetFlags(event) & ~(kCGEventFlagMaskNumericPad | kCGEventFlagMaskSecondaryFn | kCGEventFlagMaskNonCoalesced):
                 return event
             keycode = CGEventGetIntegerValueField(event, kCGKeyboardEventKeycode)
             key = KEYCODE_TO_KEY.get(keycode)
```

morinted:
As usual, your code is best. I'd add to the comment and say "the arrow and extended (home, end, etc.) keys"

benoit-pierre:
OK, done.

morinted:
I'll just do one last test through then merge. Likely tonight as I'm out
for the day.

PR #354, opened 2016-02-12
Add Appveyor support

benoit-pierre:
The Appveyor build consist of the following steps:
-  install all the required packages
-  run the testsuite
- package a version with PyInstaller.

The version number will be derived from `git describe --tag`, e.g. `v2.5.8-301-g6284391` -> `Plover-2.5.8+302.g315e5ee.exe`.

Most of the work is done by the `windows/helper.py` script, which can also be used on Linux to setup a Wine prefix to test and package the Windows version (see `--help` for usage).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ nosetests.xml
 # deleted files
 \#*#
 .#*
+
+/windows/.wine/
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,51 @@
+environment:
+
+  global:
+    PIP_ACCEL_LOG_FORMAT: "%(name)-18s %(levelname)s %(message)s"
+
+  matrix:
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7"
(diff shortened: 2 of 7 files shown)


benoit-pierre:
Note that you still need to test the executable generated by PyInstaller, as it will not error out if a module cannot be found. Something I found the hard way when wxPython would not install itself in the right location...

morinted:
So, why use AppVeyor instead of Travis for the Windows build?

Also, would you like me to set up AppVeyor for the repository to try this out?

benoit-pierre:
Travis CI environments are Linux or OS X only: https://docs.travis-ci.com/user/ci-environment/#Virtualization-environments

We can use both.

Sure, go ahead and setup the repository (I tested the build with mine).

morinted:
Okay, I've set up AppVeyor, I think it will only run for new PR submissions and commits to master, so I'll merge this to see how it works. We can revert if required. I'm very excited for getting CI set up, thanks so much for making this.

morinted:
Seems good! Very happy to have this for Windows. Only two beasts left.

PR #355, opened 2016-02-12
oslayer/xkeyboardcontrol: add Unicode support

benoit-pierre:
This is done by dynamically changing the keymap when a character we want
to send is not already mapped. A rolling list of unused keycodes is used
to ensure we don't reuse a keycode before the target application has
time to process each sequence of keymap change + key press/release
events. The keymap entries used by Plover are marked with a special
keysym, so if Plover is killed/restarted, those entries are still
considered unused, and consequently reusable.

### DIFF
--- a/plover/oslayer/xkeyboardcontrol.py
+++ b/plover/oslayer/xkeyboardcontrol.py
@@ -1162,9 +1162,12 @@ class KeyboardEmulation(object):
         number_of_backspace -- The number of backspaces to emulate.
 
         """
+        target_window = self.display.get_input_focus().focus
         for x in xrange(number_of_backspaces):
-            self._send_keycode(self.backspace_mapping.keycode,
+            self._send_keycode(target_window,
+                               self.backspace_mapping.keycode,
                                self.backspace_mapping.modifiers)
+        self.display.sync()
 
     def send_string(self, s):
         """Emulate the given string.
@@ -1177,12 +1180,16 @@ class KeyboardEmulation(object):
 
         """
         assert isinstance(s, unicode)
+        target_window = self.display.get_input_focus().focus
         for char in s:
             keysym = uchr_to_keysym(char)
             mapping = self._get_mapping(keysym)
             if mapping is None:
                 continue
-            self._send_keycode(mapping.keycode, mapping.modifiers)
+            self._send_keycode(target_window,
+                               mapping.keycode,
+                               mapping.modifiers)
+        self.display.sync()
 
     def send_key_combination(self, combo_string):
         """Emulate a sequence of key combinations.
@@ -1259,10 +1266,12 @@ class KeyboardEmulation(object):
             xtest.fake_input(self.display, event_type, keycode)
         self.display.sync()
 
-    def _send_keycode(self, keycode, modifiers=0):
+    def _send_keycode(self, target_window, keycode, modifiers=0):
         """Emulate a key press and release.
 
         Arguments:
+
+        target_window -- The window to send the event to.
 
         keycode -- An integer in the inclusive range [8-255].
 
@@ -1271,15 +1280,17 @@ class KeyboardEmulation(object):
         and Alt.
 
         """
-        self._send_key_event(keycode, modifiers, event.KeyPress)
-        self._send_key_event(keycode, modifiers, event.KeyRelease)
+        self._send_key_event(target_window, keycode, modifiers, event.KeyPress)
+        self._send_key_event(target_window, keycode, modifiers, event.KeyRelease)
 
-    def _send_key_event(self, keycode, modifiers, event_class):
+    def _send_key_event(self, target_window, keycode, modifiers, event_class):
         """Simulate a key press or release.
 
         These events are not detected by KeyboardCapture.
 
         Arguments:
+
+        target_window -- The window to send the event to.
 
         keycode -- An integer in the inclusive range [8-255].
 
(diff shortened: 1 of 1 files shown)


morinted:
What flavors of Linux have you tried this on?

benoit-pierre:
Arch Linux.

morinted:
We should try on Ubuntu (don't care about the version) as I imagine that's what most people are using, just to see if there are any gotchyas.

benoit-pierre:
Only standard X11 calls, so I don't expect any issue on another distribution. I did however think of some use case that could be problematic. Will try to test it and amend the code accordingly.

benoit-pierre:
OK, so I've completely rewritten the code, and tested with a Ubuntu VM too.

plover/oslayer/xkeyboardcontrol.py
```diff
@@ -246,18 +252,905 @@ def ignore_key_events(self, key_events):
...
+
+    def _update_keymap(self):
+        '''Analyse keymap, build a mapping of keysym to (keycode + modifiers),
+        and find unused keycodes that can be used for unmapped keysyms.
+        '''
+        self.keymap = {}
+        self.custom_mappings_queue = []
+        # Analyse X11 keymap.
+        keycode = self.display.display.info.min_keycode
+        nb_keycodes = self.display.display.info.max_keycode - keycode + 1
```
jeremy-w: It took me a while to work out that `nb` was "number of". Maybe `keycode_count` would avoid the need to abbreviate?

benoit-pierre: That's something I use a lot, `nb_something`, but I now realize that might be a mistake... How about: `num_keycodes`?

jeremy-w: That would normally be clear, but this case is special, because we're dealing with keycodes. What's confusing with any identifier name using just "num" here is that it could be abbreviating "numeric". Presumably there are keycodes that correspond to the number keys, and those would be either "number keycodes" or "numeric keycodes", and then I'm confused again. The code clears it up, but it's a confusion we can avoid, either by spelling out `number_of_keycodes` or flipping it around to `keycode_count`.

benoit-pierre: OK, will amend.

benoit-pierre:
I'll rewrite the history once we're ready to merge.

morinted:
Sounds good. I'm feeling as if you should merge when you think it's ready, as no one else can really test it.

If I were to test this, I suppose that I'd try:

- A normal sentence.
- A long word (supercalifragilisticexpialidocious).
- A command sequence (control c, control v, Alt tab).
- Arrow keys through a document.

Or, more simply:

```json
{
"TEFT": "The quick brown fox jumps over the lazy dog. Supercalifragilisticexpialidocious. {#Shift_L(Left Left Left) Control_L(c v v v v v). You should see 5 's. 's. All is well? All is well. 🐱 😆 💩 👪 ⁉." 
}
```

benoit-pierre:
You're missing a `}`. It works fine. It's also unlikely to catch some of the corner cases. My test dictionaries is:

```
{
"-F":"♜♞♝♛♚♝♞♜",
"*F":"ǐ",
"-R":"♖♘♗♕♔♗♘♖",
"*R":"ħĦ7ðÐ8*þÞ×÷äÄåÅãÃøØđĐłŁúÚüÜöÖ«‹»",
"-P": "По оживлённым берегам",
"*P": "ПО ОЖИВЛЁННЫМ БЕРЕГАМ",
"-B": "Ἰοὺ ἰού· τὰ πάντʼ ἂν ἐξήκοι σαφῆ",
"*B": "ἸΟῪ ἸΟΎ· ΤᾺ ΠΆΝΤʼ ἊΝ ἘΞΉΚΟΙ ΣΑΦῆ",
"-L": "पशुपतिरपि तान्यहानि कृच्छ्राद् ",
"-G": "子曰：「學而時習之，不亦說乎？有朋自遠方來，不亦樂乎",
"-T": "ஸ்றீனிவாஸ ராமானுஜன் ஐயங்கார்",
"-S": "إِيَّاكَ نَعْبُدُ وَإِيَّاكَ نَسْتَعِينُ",
"-D": "♈♉♊♋♌♍♎♏♐♑♒♓",
"-Z": "àáâãäāăąȧǎȁȃèéêẽëēĕęėǎěȅȇìíîĩïīĭįiìǐȉȋòóôõöōŏǫȯőǒȍȏùúûũüūŭųǎěűǔȕȗỳýŷỹÿȳǎěűẏǎěűẏǎěűẏ",
"*Z": "ÀÁÂÃÄĀĂĄȦǍȀȂÈÉÊẼËĒĔĘĖǍĚȄȆÌÍÎĨÏĪĬĮIÌǏȈȊÒÓÔÕÖŌŎǪȮŐǑȌȎÙÚÛŨÜŪŬŲǍĚŰǓȔȖỲÝŶỸŸȲǍĚŰẎǍĚŰẎǍĚŰẎ"
}
```

But the problem I had thought about was with this:

```python
#!/usr/bin/env python2
# -*- coding: utf-8 -*-

import time

from plover.oslayer.xkeyboardcontrol import KeyboardEmulation
from plover import log


log.set_level(log.DEBUG)
kbd = KeyboardEmulation()

time; time.sleep(3)

kbd.send_string(u'ⅰⅱⅲⅳⅴⅵⅶⅷⅰ♜')
```

With only 8 free mappings, with the previous version of the code, the 9th character "ⅰ" would be wrong because for sending "♜" we reuse the same mapping.

The thing is, if you don't have enough free mappings and/or the application you send output is too slow, you can still run into issues because you'll reuse a mapping too soon... Like with only 4 free mappings and GMail in Firefox. (But for reference, I get 28 unused keycodes, so 56 free mappings with the standard colemak layout)

morinted:
Well, some support is better than no support. I'm in favor of merging this now, same as the Windows rewrite, then at least we'll get early feedback when we release the first pre-release.

benoit-pierre:
Agreed.

PR #356, opened 2016-02-12
Switch to x64 builds on Mac

morinted:
We were using 32-bit builds due to an old version of wxPython.
Additionally, PyObjc had an update that allowed us to use
"CGEventKeyboardSetUnicodeString" without a conversion through
PyObjc.

The application could still technically be built with 32-bit
support in "universal" fashion, but that value is super marginal,
because it's more complex to get setup and this would only affect
users on 32-bit OSX machines, which are very few. If they insist
on 32-bit, they can use the older builds.

The other reason to stick with a stripped application is the
filesize. A universal OSX application weighs in at about 90MB,
the 64/32-bit-only versions are a lighter 50MB (less in a .dmg).

@jeremy-w Can you please try this out and see if it resolves your
issues with Plover crashing after running it as `sudo`? If this works,
it will greatly simplify setting up Travis CI, which I'm very excited about.

I also removed the venv instructions, I get the sense they have no
value at this point, especially if we implement your bootstrap (or,
perhaps just the requirements.txt as it may no longer be necessary).

### DIFF
--- a/osx/README.md
+++ b/osx/README.md
@@ -4,7 +4,7 @@
 
 ### Python
 
-You need Python 2.7, with 32-bit support. `brew install python --universal --framework`
+You need Python 2.7. We build with brewed Python, as Apple's Python will cause you a few unexpected issues. `brew install python --framework`
 
 ### XCode Tools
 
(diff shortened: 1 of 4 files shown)


jeremy-w:
The bootstrap bits that ensure wxPython is installed via Homebrew and that handle the pip installation are probably still worthwhile. It'll get a lot simpler by not having to test for 32-bit stuff, though!

I poked around to see what we'd be dropping by dropping 32-bit support. Apple switched to 64-bit machines in August 2007. Chrome went 64-bit–only with version 39 in November 2014 (http://www.computerworld.com/article/2849225/chrome-for-os-x-turns-64-bit-forsakes-early-intel-macs.html). For perspective, every MacBook Air ever has been able to run x86-64 binaries.

Sounds like a clear win!

I'll test this PR on my machine when I can. Should be in a few hours.
--
Jeremy W. Sherman
http://jeremywsherman.com/

morinted:
Yeah, I looked over the bootstrap and it seems useful, I'm also all for a requirements.txt

I'm trying to get Travis building on my local branch. He seems to not be cooperating for one reason or another. I'll keep hashing it out with him and await your updates with excitement.

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -294,13 +295,11 @@ def characters(s):
...
         yield character
 
-CGEventKeyboardSetUnicodeString = ctypes.cdll.LoadLibrary(ctypes.util.find_library('ApplicationServices')).CGEventKeyboardSetUnicodeString
-CGEventKeyboardSetUnicodeString.restype = None
 native_utf16 = 'utf-16-le' if sys.byteorder == 'little' else 'utf-16-be'
 
 def set_string(event, s):
     buf = s.encode(native_utf16)
-    CGEventKeyboardSetUnicodeString(objc.pyobjc_id(event), len(buf) / 2, buf)
+    CGEventKeyboardSetUnicodeString(event, len(buf) / 2, buf)
```
jeremy-w: A-hah! I thought that pyobjc_id thing looked suspicious, but didn't think to just try removing it. I was going to scour the PyObjC docs first. Failure to bias towards action on my part there!

osx/README.md
```diff
@@ -14,53 +14,21 @@ You need XCode commandline tools. Check by running `gcc` in the terminal. You wi
...
-### cython-hidapi
-
-Installing hidapi through pip only creates a 64-bit binary on my system, so when it is stripped to 32-bit for Plover, you cannot use the module. The workaround for this is to install `hidapi` from another source. I've forked `cython-hidapi` for this purpose. Please [build and install the fork](https://github.com/morinted/cython-hidapi).
-
 ### Pip
 
 Pip dependencies can be grabbed with:
 
-`pip install appdirs pyserial simplejson py2app Cython mock pyobjc-core pyobjc-framework-cocoa pyobjc-framework-quartz`
+`pip install appdirs pyserial simplejson py2app Cython mock hidapi pyobjc-core pyobjc-framework-cocoa pyobjc-framework-quartz`
```
jeremy-w: I didn't have hidapi in my requirements.txt in https://github.com/openstenoproject/plover/pull/342. I probably didn't run into it as an issue since I don't have any non-keyboard steno devices around.

jeremy-w: Oh wait, yes I did, but I was using your custom branch. Guess we no longer need that!

morinted: You were using my fork, which isn't needed anymore with the 64-bit-onlyness.

jeremy-w:
Successfully used to steno with `sudo python launch.py`. :shipit:

PR #357, opened 2016-02-13
Add ability to add multiple dictionaries at a time

### LINKED ISSUE #167: Add dictionary > non-existent filename, hangs Plover
Trying to add a dictionary with a non-existent filename freezes Plover. The expected behavior (for me) would be to create an empty dictionary with that filename.

Tested on Windows 7, Plover 2.5.7

morinted:
Also, provide some useful feedback in case the file you are adding doesn't
exist, should fix #167. Also lets you know if the dictionary is already
present.

### DIFF
--- a/plover/gui/config.py
+++ b/plover/gui/config.py
@@ -15,6 +15,7 @@ from plover.gui.serial_config import SerialConfigDialog
 import plover.gui.add_translation
 import plover.gui.lookup
 import plover.gui.dictionary_editor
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Works fine on Linux and Windows 10 VM, so I'll go ahead and merge (please don't forget to follow the contributing guidelines and mention the platforms you tested on! :P).

PR #335, opened 2016-02-03
travis ci integration test

benoit-pierre:

PR #359, opened 2016-02-13
Add OS X Travis build

morinted:
Add a Travis build for OSX on Travis.

Travis doesn't support Python on OS X, so we need to use generic language.
The commands install the Python we need, then get wx, then run the
bootstrap.sh for the OSX setup to make sure everything is good and get pip
dependencies. The deployment will publish builds to GitHub tags, but
otherwise will just discard the build. Maybe later we can add another
deployment for non-tagged builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+language: python
+matrix:
+  include:
+    - os: osx
+      language: generic
+before_install:
+ - >
+  if [[ "$TRAVIS_OS_NAME" == "osx" ]];
+  then
+    brew update &&
+    brew install python --framework &&
+    brew link --overwrite python &&
+    brew install wxpython &&
+    (cd osx && ./bootstrap.sh); fi
+
+install: true
+script:
+  - python setup.py patch_version
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then (cd osx && make); fi
+deploy:
+  provider: releases
+  skip_cleanup: true
+  api_key:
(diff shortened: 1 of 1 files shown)


morinted:
@benoit-pierre If we use Travis for both OS X .dmg and Linux .deb, should we merge one, then the other, or try to get them both going at once? Is the .deb building going to be a Travis task that then uploads and updates a PPA on release-tag like this one does? I wish Travis had built in artifact support like AppVeyor :/

As an aside, I'm really glad that the OSX situation is less flaky than it was. Having a reproducible environment setup is bringing calm to my soul. The Travis builds are also small in size (< 20MB as a .dmg, 38MB as a .app, better than my system makes).

jeremy-w:
I'm in favor of merging this now, since it's ready to go. #335 needs conflicts resolved before it can go through, so it'll take some work anyway. Getting a working Linux + OS X build matrix can be part of that work.

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+  include:
+    - os: osx
+      language: generic
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install python --framework; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew link --overwrite python; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install wxpython; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cd osx && ./bootstrap.sh && cd ..; fi
+
```
benoit-pierre: About formatting it like this instead:
```
before_install:
 - |
  if [[ "$TRAVIS_OS_NAME" == "osx" ]
  then
    brew update &&
    brew install python --framework &&
    brew link --overwrite python &&
    brew install wxpython &&
    (cd osx && ./bootstrap.sh)
  fi
```

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+install: true
+script:
+  - python run_tests.py
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then cd osx && make && cd ..; fi
```
benoit-pierre: Same as above.

morinted:
@benoit-pierre If we are not going to be making Linux builds with Travis (AppVeyor makes the egg), then maybe we should remove the matrix completely and just make `.travis.yml` for Mac only?

benoit-pierre:
The egg can be built on Windows, but not the wheel file. So I'll probably move both to Travis CI.

morinted:
Okay, I reformatted, not exactly the same but it should suffice. Also included the helper version patch.

PR #361, opened 2016-02-14
log: don't set the print handler level

benoit-pierre:
Otherwise, it acts as a second (possibly more restrictive) filter, after
the logger level. Leaving it unset make it possible to change the level
at the logger level.


benoit-pierre:
Same for the file handler.

PR #362, opened 2016-02-15
Rename "NKRO Keyboard" and "Sidewinder" to "Keyboard"

morinted:
In the code, we still refer to the Sidewinder even though it is no longer relevant for most users. As such, it's a little confusing when you are looking for the "keyboard" machine and you need to learn that it's called the Sidewinder internally. Also, it's a little misleading to say "NKRO Keyboard: connected", because users who don't have an NKRO keyboard will open Plover and see this message, and may be confused.

These issues affected me when I very first started using Plover (no NKRO keyboard, looking to find the plover/machine/keyboard).

@stenoknight, can I get your comment on this? Particularly, the change from "NKRO Keyboard: connected" to "Keyboard: connected"? 
screen shot 2016-02-14 at 11 42 54 pm
screen shot 2016-02-14 at 11 42 44 pm

### DIFF
--- a/plover/config.py
+++ b/plover/config.py
@@ -19,7 +19,7 @@ CONFIG_FILE = os.path.join(CONFIG_DIR, 'plover.cfg')
 # General configuration sections, options and defaults.
 MACHINE_CONFIG_SECTION = 'Machine Configuration'
 MACHINE_TYPE_OPTION = 'machine_type'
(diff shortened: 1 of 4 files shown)


stenoknight:
Agree 100%.

jeremy-w:
Writing this using this branch. Works great!

PR #364, opened 2016-02-15
Add Translation Transparency

jeremy-w:
This is a continuation of #252, rebased against master.

It has been reworked to set alpha/opacity rather than transparency, so that 100% means fully opaque not fully invisible.

captura de pantalla 2016-02-19 a las 08 37 49

## TODO

- [x] Label both ends, so we have a fixed Invisible label at the left edge and a dynamic Opaque: N% on the right edge of the slider.
    - Done. captura de pantalla 2016-02-24 a las 16 52 23
- [x] Figure out how to stop the main window from activating alongside the dictionary update window. Only the update window should come to fore when invoked by `{PLOVER:ADD_TRANSLATION}`.
    - User-fixable: Minimize the window.
- [ ] Address [ragged left alignment.](https://github.com/openstenoproject/plover/pull/364#issuecomment-189733306)
  raggedleft

### DIFF
--- a/plover/config.py
+++ b/plover/config.py
@@ -89,6 +89,8 @@ TRANSLATION_FRAME_X_OPTION = 'x'
 DEFAULT_TRANSLATION_FRAME_X = -1
 TRANSLATION_FRAME_Y_OPTION = 'y'
 DEFAULT_TRANSLATION_FRAME_Y = -1
+TRANSLATION_FRAME_OPACITY_OPTION = 'opacity'
+DEFAULT_TRANSLATION_FRAME_OPACITY = 100
 
 LOOKUP_FRAME_SECTION = 'Lookup Frame'
 LOOKUP_FRAME_X_OPTION = 'x'
@@ -120,6 +122,21 @@ RTF_EXTENSION = '.rtf'
 
 # Logging constants.
 LOG_EXTENSION = '.log'
+
+
+MIN_FRAME_OPACITY = 0
+MAX_FRAME_OPACITY = 100
+
+
+def raise_if_invalid_opacity(opacity):
+    """
+    Raises ValueError if opacity is out of range defined by
+    [MIN_FRAME_OPACITY, MAX_FRAME_OPACITY].
+    """
+    if not (MIN_FRAME_OPACITY <= opacity <= MAX_FRAME_OPACITY):
+        message = "opacity %u out of range [%u, %u]" \
+            % (opacity, MIN_FRAME_OPACITY, MAX_FRAME_OPACITY)
+        raise ValueError(message)
 
 # TODO: Unit test this class
 
@@ -401,7 +418,25 @@ class Config(object):
         return self._get_int(TRANSLATION_FRAME_SECTION, 
                              TRANSLATION_FRAME_Y_OPTION,
                              DEFAULT_TRANSLATION_FRAME_Y)
-                             
+
+    def set_translation_frame_opacity(self, opacity):
+        raise_if_invalid_opacity(opacity)
+        self._set(TRANSLATION_FRAME_SECTION,
+                  TRANSLATION_FRAME_OPACITY_OPTION,
+                  opacity)
+
+    def get_translation_frame_opacity(self):
+        opacity = self._get_int(TRANSLATION_FRAME_SECTION,
+                                TRANSLATION_FRAME_OPACITY_OPTION,
+                                DEFAULT_TRANSLATION_FRAME_OPACITY)
+        try:
+            raise_if_invalid_opacity(opacity)
+        except ValueError as e:
+            log.error("translation %s, reset to %u",
+                      e, DEFAULT_TRANSLATION_FRAME_OPACITY)
+            opacity = DEFAULT_TRANSLATION_FRAME_OPACITY
+        return opacity
+
     def set_lookup_frame_x(self, x):
         self._set(LOOKUP_FRAME_SECTION, LOOKUP_FRAME_X_OPTION, x)
     
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -21,6 +21,9 @@ class AddTranslationDialog(wx.Dialog):
                config.get_translation_frame_y())
         wx.Dialog.__init__(self, parent, title=TITLE, pos=pos)
 
+        opacity = config.get_translation_frame_opacity()
+        self._apply_opacity(opacity)
+
         self.config = config
 
         # components
@@ -45,7 +48,7 @@ class AddTranslationDialog(wx.Dialog):
         sizer.Add(label, 
                   flag=wx.TOP | wx.RIGHT | wx.BOTTOM | wx.ALIGN_CENTER_VERTICAL, 
                   border=self.BORDER)
-        sizer.Add(self.translation_text          , 
+        sizer.Add(self.translation_text,
                   flag=wx.TOP | wx.RIGHT | wx.BOTTOM | wx.ALIGN_CENTER_VERTICAL, 
                   border=self.BORDER)
         sizer.Add(button, 
@@ -203,6 +206,22 @@ class AddTranslationDialog(wx.Dialog):
         self.config.set_translation_frame_x(pos[0]) 
         self.config.set_translation_frame_y(pos[1])
         event.Skip()
+
+    def _apply_opacity(self, opacity):
+        """
+        Given a opacity in range 0 through 100, makes the window that
+        percent opaque. Handles conversion to range 0 through 0xFF.
+        """
+        if not self.CanSetTransparent():
+            return
+
+        # opacity ranges from 0 to 100,
+        # but SetTransparent takes a byte in range 0 through 255.
+        # NOTE: wxWidgets calls opacity transparency,
+        # but 0 still means invisible.
+        fractional_alpha = opacity / 100.0
+        alpha_byte = int(round(0xFF * fractional_alpha)) % 0x100
+        self.SetTransparent(alpha_byte)
 
     def _normalized_strokes(self):
         strokes = self.strokes_text.GetValue().upper().replace('/', ' ').split()
--- a/plover/gui/config.py
+++ b/plover/gui/config.py
@@ -49,6 +49,7 @@ SPACE_PLACEMENTS = [SPACE_PLACEMENT_BEFORE, SPACE_PLACEMENT_AFTER]
 FIRST_STROKE_LABEL = "First Stroke:"
 START_CAPITALIZED_LABEL = "Start Capitalized"
 START_ATTACHED_LABEL = "Suppress Space"
+TRANSLATION_OPACITY_LABEL = "Translation Dialog Opacity"
 
 UI_BORDER = 4
 COMPONENT_SPACE = 3
@@ -56,6 +57,7 @@ UP_IMAGE_FILE = os.path.join(conf.ASSETS_DIR, 'up.png')
 DOWN_IMAGE_FILE = os.path.join(conf.ASSETS_DIR, 'down.png')
 REMOVE_IMAGE_FILE = os.path.join(conf.ASSETS_DIR, 'remove.png')
 TITLE = "Plover Configuration"
+
 
 class ConfigurationDialog(wx.Dialog):
     """A GUI for viewing and editing Plover configuration files.
@@ -65,10 +67,10 @@ class ConfigurationDialog(wx.Dialog):
     application, which is typically after an application restart.
 
     """
-    
+
     # Keep track of other instances of ConfigurationDialog.
     other_instances = []
-    
+
     def __init__(self, engine, config, parent):
         """Create a configuration GUI based on the given config file.
 
@@ -79,9 +81,9 @@ class ConfigurationDialog(wx.Dialog):
         won't tell the user that Plover needs to be restarted.
         """
         pos = (config.get_config_frame_x(), config.get_config_frame_y())
-        size = wx.Size(config.get_config_frame_width(), 
+        size = wx.Size(config.get_config_frame_width(),
                        config.get_config_frame_height())
-        wx.Dialog.__init__(self, parent, title=TITLE, pos=pos, size=size, 
+        wx.Dialog.__init__(self, parent, title=TITLE, pos=pos, size=size,
                            style=wx.DEFAULT_DIALOG_STYLE | wx.RESIZE_BORDER)
         self.engine = engine
         self.config = config
@@ -100,7 +102,7 @@ class ConfigurationDialog(wx.Dialog):
 
         # Configuring each tab
         self.machine_config = MachineConfig(self.config, notebook)
-        self.dictionary_config = DictionaryConfig(self.engine, self.config, 
+        self.dictionary_config = DictionaryConfig(self.engine, self.config,
                                                   notebook)
         self.logging_config = LoggingConfig(self.config, notebook)
         self.display_config = DisplayConfig(self.config, notebook, self.engine)
@@ -129,20 +131,20 @@ class ConfigurationDialog(wx.Dialog):
         button_sizer.Realize()
 
         sizer.Add(button_sizer, flag=wx.ALL | wx.ALIGN_RIGHT, border=UI_BORDER)
-        
+
         self.SetSizerAndFit(sizer)
         self.SetRect(AdjustRectToScreen(self.GetRect()))
-        
+
         # Binding the save button to the self._save callback
         self.Bind(wx.EVT_BUTTON, self._save, save_button)
-        
+
         self.Bind(wx.EVT_MOVE, self.on_move)
         self.Bind(wx.EVT_SIZE, self.on_size)
         self.Bind(wx.EVT_CLOSE, self.on_close)
-        
+
     def on_move(self, event):
         pos = self.GetScreenPositionTuple()
-        self.config.set_config_frame_x(pos[0]) 
+        self.config.set_config_frame_x(pos[0])
         self.config.set_config_frame_y(pos[1])
         event.Skip()
 
@@ -151,14 +153,14 @@ class ConfigurationDialog(wx.Dialog):
         self.config.set_config_frame_width(size.GetWidth())
         self.config.set_config_frame_height(size.GetHeight())
         event.Skip()
-        
+
     def on_close(self, event):
         self.other_instances.remove(self)
         event.Skip()
 
     def _save(self, event):
         old_config = self.config.clone()
-        
+
         self.machine_config.save()
         self.dictionary_config.save()
         self.logging_config.save()
@@ -204,20 +206,30 @@ class MachineConfig(wx.Panel):
         sizer = wx.FlexGridSizer(2, 3)
         sizer.AddGrowableCol(1)
 
-        sizer_flags = wx.SizerFlags(1).Align(wx.ALIGN_CENTER_VERTICAL).Border(wx.ALL, UI_BORDER)
+        sizer_flags = wx.SizerFlags(1) \
+            .Align(wx.ALIGN_CENTER_VERTICAL) \
+            .Border(wx.ALL, UI_BORDER)
 
-        sizer.AddF(wx.StaticText(self, label=MACHINE_LABEL), sizer_flags.Left())
+        sizer.AddF(
+            wx.StaticText(self, label=MACHINE_LABEL),
+            sizer_flags.Left())
         machines = machine_registry.get_all_names()
         current_machine = self.config.get_machine_type()
         self.choice = wx.Choice(self, choices=sorted(machines))
-        self.choice.SetStringSelection(machine_registry.resolve_alias(current_machine))
+        selected_machine = machine_registry.resolve_alias(current_machine)
+        self.choice.SetStringSelection(selected_machine)
         sizer.AddF(self.choice, sizer_flags.Expand())
         self.Bind(wx.EVT_CHOICE, self._update, self.choice)
-        self.config_button = wx.Button(self, id=wx.ID_PREFERENCES, label=CONFIG_BUTTON_NAME)
+
+        self.config_button = wx.Button(
+            self,
+            id=wx.ID_PREFERENCES,
+            label=CONFIG_BUTTON_NAME)
         sizer.AddF(self.config_button, sizer_flags.Right())
         self.Bind(wx.EVT_BUTTON, self._advanced_config, self.config_button)
 
-        self.auto_start_checkbox = wx.CheckBox(self, label=MACHINE_AUTO_START_LABEL)
+        self.auto_start_checkbox = wx.CheckBox(
+            self, label=MACHINE_AUTO_START_LABEL)
         auto_start = config.get_auto_start()
         self.auto_start_checkbox.SetValue(auto_start)
         sizer.AddF(self.auto_start_checkbox, sizer_flags.Left())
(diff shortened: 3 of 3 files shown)


benoit-pierre:
Wouldn't a `SpinCtrl` be a more logical choice instead of a `Slider`?

jeremy-w:
@benoit-pierre It would, but I just want to get this cleaned up for now. That's an ER we can do later. Incremental progress. :)

jeremy-w:
I had time to flake8-ify gui/config.py. Got to run now, let's see what Appveyor says. :)

morinted:
So, should we offer one transparency setting for all the helper windows at once (add translation, look up, suggestions, ticker tape), or one for each (somehow)

stenoknight:
All or nothing works for me. Can't think of a case where I would want some
windows transparent and others not.

jeremy-w:
For me, I'd prefer windows that pin on top to have variable transparency separately from transient windows like dictionary update. When I invoke dictionary update, I'm slow, and want to see what I'm doing. When I have suggestions pinned on top, I'm on a laptop screen, so I'm constantly micro-managing on top / not on top and where the window is. (Having it only pin on top when translation is enabled would fix a lot of this.)

benoit-pierre:
Please let's not turn this into feature creep. If the only use case is a completely transparent "Add Translation" window, then let's just implement that.

benoit-pierre:
Also, how about keeping this on point, and avoid all those cleanups. Create a separate pull request, and we can merge it right now.

stenoknight:
That's also fine by me.

morinted:
Yeah, in terms of a pull request, it's probably best to keep this as just
transparency for the add translation window. It also doesn't make sense to
include the character changes and stuff when it doesn't affect this
feature, feel free to submit another PR with clean ups that are
OS-specific, though.

I'd argue that there's value in transparency settings for the helper
windows, because they are set to stay on top, and it would be nice if they
didn't block my view all the time.

But you're probably right that the add translation window is only one that
needs to be completely transparent.

As an aside, it might be necessary to add a way to set the transparency
through a stroke, because sometimes you might want an invisible dictionary
update, sometimes a visible one. {PLOVER:ADD_TRANSLATION:75} for 75%
transparency?

All this would best be reflected in separate issues, but I'm brainstorming.

benoit-pierre:
If I don't make a separate pull request, I usually try to put these kind cleanups first: if the least controversial commits are first, then it's possible to merge up to a point, leaving the rest for latter.

jeremy-w:
Interesting idea. I can split out the coding stuff, sure. It's been centralized into a single commit anyway.

benoit-pierre:
The only problem with that approach is that GitHub reorder issues by time, when displaying them... Which is annoying when reordering commits.

morinted:
You can modify the dates if the order concerns you, I've done that before
when logical.

jeremy-w:
@benoit-pierre It's been split out as #366 now. I did just add a couple new commits for the clamping behavior I mentioned, but other than that, the rest is what you saw yesterday.

morinted:
So for this PR, we need the default transparency set to 100%, and verification that it works on Linux and Windows.

jeremy-w:
You're having the same confusion I had on first contact with this code and UI. Right now, 0% transparency means fully visible (aka not at all transparent), and that also happens to be the default.

Maybe we should switch to alpha or opacity, where 100% means fully visible?
--
Jeremy W. Sherman
http://jeremywsherman.com/

morinted:
Yes. I've never seen it done this way.

jeremy-w:
@morinted Ready to test.

morinted:
@jeremy-w Works well for me. Let's discuss. Feel free to push back on my feedback below if you disagree.

Functionality-wise, how about labeling either end of the slider to be explicit (after redesigning that darned main UI)?

    Invisible |-----o-------------| Opaque: 28%

Is it possible/viable to have the translation live-update and only write to config on save and cancel on cancel, like you mentioned earlier?

I haven't looked at the code yet, this is just me trying it out. Very cool to have, by the way.

Hm, I was going to complain about the aesthetic of the "title" label, but all the rest of Plover's config dialog is on the same level, so never mind *shrug*.

plover/gui/add_translation.py
```diff
@@ -202,6 +205,20 @@ def on_move(self, event):
         self.config.set_translation_frame_y(pos[1])
         event.Skip()
 
+    def applyTransparency(self, transparency):
+        """
+        Given a transparency in range 0 through 100, makes the window that
+        percent transparent. Handles conversion to range 0 through 0xFF.
+        """
+        if not self.CanSetTransparent():
```
jeremy-w: We should probably either disable or not show the transparency UI in Config when we `not CanSetTransparent`.

jeremy-w:
## Labeling the Ends
That's probably doable. I'll have to actual read the sizer code rather than glazing over it, though. ;)

I also just noticed that Show Strokes UI uses different plurality: "Open stroke display" button, but "Open stroke**s** display on startup" checkbox.

## Live Update
I think this should be handled as a separate enhancement request.

Discoverability of live update would be a problem. We'd want a button or something to point out, "Hey, you should open this transient window for a bit so you can see what this'll look like."

Implementation-wise, we'd want slider movement to update the config object immediately, and have the update window watch the config object for changes to transparency somehow. On cancel, we'd undo the changes to config; on save, we'd leave them applied. It's not till the save button is clicked that the config gets flushed to disk, after all.

stenoknight:
This is perhaps a silly question, but currently when I bring up the "add
translation" window, it brings Plover's main window to front as well. Will
this new transparency setting allow me to make dictionary updates without
the (presumably not transparent) Plover window appearing?

On Fri, Feb 19, 2016 at 8:52 AM, Jeremy W. Sherman <notifications@github.com
> wrote:

> Labeling the Ends
>
> That's probably doable. I'll have to actual read the sizer code rather
> than glazing over it, though. ;)
>
> I also just noticed that Show Strokes UI uses different plurality: "Open
> stroke display" button, but "Open stroke*s* display on startup" checkbox.
> Live Update
>
> I think this should be handled as a separate enhancement request.
>
> Discoverability of live update would be a problem. We'd want a button or
> something to point out, "Hey, you should open this transient window for a
> bit so you can see what this'll look like."
>
> Implementation-wise, we'd want slider movement to update the config object
> immediately, and have the update window watch the config object for changes
> to transparency somehow. On cancel, we'd undo the changes to config; on
> save, we'd leave them applied. It's not till the save button is clicked
> that the config gets flushed to disk, after all.
>
> —
> <https://github.com/openstenoproject/plover/pull/364#issuecomment-186221489>
> .
>



-- 
Mirabai Knight, CRC, RDR
StenoKnight CART Services
917 576 4989
mkk@stenoknight.com
http://stenoknight.com

balshetzer:
I think this happens because the translation window is a child of the main
frame.

jeremy-w:
That's not silly at all. I hadn't noticed that, but now that you call it out:

- The main window comes to fore along with the update window.
- If I close the update window, the main window goes back to being hidden (maybe it's reactivating the last-active app?).

Other windows aren't affected this way - the Suggestions window for example stays put through all this.

benoit-pierre:
I don't have this issue on Linux: if I have hidden the main window, it stays that way when the "Add Translation" window is shown with `TKUPT`. It only pops back to front if it was not hidden.

benoit-pierre:
And the behavior may be completely different with another window manager...

morinted:
OSX: As long as I have Plover minimized, the dictionary update window will be the only one that pops up. But yes, if Plover is not minimized, the dictionary update window raises Plover.

stenoknight:
Ah, this is true for Windows as well. Didn't know that. Awesome!

jeremy-w:
By "hidden" I meant "occluded". I wasn't actually minimizing/hiding anything, it was just behind 10 other app windows. When I summon the add translation window, the main window comes to fore along with it, when I really just want the add translation window to become foreground.

plover/gui/add_translation.py
```diff
@@ -21,6 +21,9 @@ def __init__(self, parent, engine, config):
                config.get_translation_frame_y())
         wx.Dialog.__init__(self, parent, title=TITLE, pos=pos)
 
+        transparency = config.get_translation_frame_transparency()
+        self.applyTransparency(transparency)
```
morinted: Rename to `_apply_transparency`

plover/gui/config.py
```diff
@@ -46,6 +46,7 @@
 SPACE_PLACEMENT_BEFORE = "Before Output"
 SPACE_PLACEMENT_AFTER = "After Output"
 SPACE_PLACEMENTS = [SPACE_PLACEMENT_BEFORE, SPACE_PLACEMENT_AFTER]
+TRANSLATION_TRANSPARENCY_LABEL = "Translation Dialog Transparency"
```
morinted: Translation Dialog Opacity, now that we've flipped the slider?

plover/gui/config.py
```diff
@@ -200,20 +202,30 @@ def __init__(self, config, parent):
         sizer = wx.FlexGridSizer(2, 3)
         sizer.AddGrowableCol(1)
 
-        sizer_flags = wx.SizerFlags(1).Align(wx.ALIGN_CENTER_VERTICAL).Border(wx.ALL, UI_BORDER)
+        sizer_flags = wx.SizerFlags(1) \
```
morinted: I usually visually align the periods when linebreaking on a function call, does PEP8 have an opinion here? I'm not familiar with what most people do :)

plover/gui/config.py
```diff
@@ -200,20 +202,30 @@ def __init__(self, config, parent):
...
+            .Align(wx.ALIGN_CENTER_VERTICAL) \
+            .Border(wx.ALL, UI_BORDER)
 
-        sizer.AddF(wx.StaticText(self, label=MACHINE_LABEL), sizer_flags.Left())
+        sizer.AddF(
```
morinted: And here I'd just break after the comma,

```
name.call(parameter1,
          parameter2)
```

Again, just going off of intuition, not sure if that is written about.

plover/gui/config.py
```diff
@@ -200,20 +202,30 @@ def __init__(self, config, parent):
...
+
+        self.config_button = wx.Button(
+            self,
+            id=wx.ID_PREFERENCES,
+            label=CONFIG_BUTTON_NAME)
         sizer.AddF(self.config_button, sizer_flags.Right())
         self.Bind(wx.EVT_BUTTON, self._advanced_config, self.config_button)
 
-        self.auto_start_checkbox = wx.CheckBox(self, label=MACHINE_AUTO_START_LABEL)
+        self.auto_start_checkbox = wx.CheckBox(
```
morinted: Yeah, I guess when I see this, Python just seems to follow a convention of "visual ident" to group things, so to me it looks like the parameters should all indent after the parentheses.

plover/gui/config.py
jeremy-w: The awkward bit is that ultimately we pass this into wx SetTransparency. Which is actually SetOpacity. But I think we could add a comment at that call site acknowledging the confusion and be clearer through the rest of the code by adopting the opacity verbiage.

--
Jeremy W. Sherman
http://jeremywsherman.com/

plover/gui/config.py
jeremy-w: So flake8 insists that, if 1 param is on the same line as the open paren, then all must be at least that indented. But if you kick all the params off the line with the open paren, then any indentation is kosher by it. I mostly prefer to keep things near the left edge, so I preferred to evict all params from the line when wrapping was required due to line length.

plover/gui/config.py
jeremy-w: I tend to dislike indentation-as-alignment vs indentation-as-relative-level. Former really does not play well with tabs (since someone will indent to a mix of tabs and spaces for their specific tab=4 sp, then I'll view with tab=8 sp and it's all crazy), latter works great no matter how you're indenting.

plover/gui/config.py
jeremy-w: I ran all this by flake8, so it's pep8 legit.

plover/gui/add_translation.py
jeremy-w: :+1:

morinted:
For your focus problem: 1) I don't think it's a big issue, considering that you can minimize/hide the main Plover window if you don't want it pulling forward. So if you are finding it really hard to solve, just drop it and we can get this change in for release. 2) if you are looking for some place to start, maybe try catching the focus event and prevent it from propagating to its parent.

jeremy-w:
Thanks. I'm not stuck, I just haven't gotten much computer leisure time lately. I was able to read up on wxWidgets' sizing and layout approach in the intervening time at least, since that's doable on mobile. I hope to have a couple hours for this today.

jeremy-w:
@morinted All TODOs todone'd, should be good to go.

plover/config.py
```diff
@@ -85,6 +85,8 @@
 DEFAULT_TRANSLATION_FRAME_X = -1
 TRANSLATION_FRAME_Y_OPTION = 'y'
 DEFAULT_TRANSLATION_FRAME_Y = -1
+TRANSLATION_FRAME_TRANSPARENCY_OPTION = 'transparency'
```
benoit-pierre: This should really be renamed to "opacity" too.

benoit-pierre:
The GUI option does not display properly (Arch Linux / Ubuntu):
2016-02-24-235052_451x232_scrot

plover/config.py
benoit-pierre: In fact, transparency should be replaced by opacity pretty much everywhere.

benoit-pierre:
The display is fine with the following changes:
```diff
diff --git i/plover/gui/config.py w/plover/gui/config.py
index 31c5e97..807005b 100644
--- i/plover/gui/config.py
+++ w/plover/gui/config.py
@@ -480,6 +480,10 @@ class DisplayConfig(wx.Panel):
         self.engine = engine
         sizer = wx.BoxSizer(wx.VERTICAL)
 
+        sizer_flags = wx.SizerFlags(0) \
+            .Align(wx.ALIGN_CENTER_VERTICAL | wx.ALIGN_LEFT) \
+            .Border(wx.ALL, COMPONENT_SPACE)
+
         show_strokes_sizer = wx.BoxSizer(wx.VERTICAL)
 
         show_strokes_button = wx.Button(
@@ -517,22 +521,19 @@ class DisplayConfig(wx.Panel):
             border=UI_BORDER,
             flag=(wx.LEFT | wx.RIGHT | wx.BOTTOM))
 
-        sizer.Add(show_suggestions_sizer, border=UI_BORDER, flag=wx.BOTTOM)
+        sizer.AddF(show_suggestions_sizer, sizer_flags)
 
         translation_transparency_sizer = wx.BoxSizer(wx.VERTICAL)
 
         translation_transparency_label = wx.StaticText(
             self, label=TRANSLATION_OPACITY_LABEL)
-        translation_transparency_sizer.Add(
-            translation_transparency_label,
-            border=COMPONENT_SPACE,
-            flag=wx.ALL)
+        translation_transparency_sizer.AddF(
+            translation_transparency_label, sizer_flags)
 
         translation_transparency_slider_sizer = wx.BoxSizer(wx.HORIZONTAL)
 
         zero_percent_label = wx.StaticText(self, label="Invisible")
-        translation_transparency_slider_sizer.Add(
-            zero_percent_label, flag=wx.ALIGN_LEFT | wx.ALIGN_CENTER_VERTICAL)
+        translation_transparency_slider_sizer.AddF(zero_percent_label, sizer_flags)
 
         configured_transparency = self.config.\
             get_translation_frame_transparency()
@@ -545,24 +546,21 @@ class DisplayConfig(wx.Panel):
         self.translation_transparency_slider.\
             Bind(wx.EVT_SLIDER, self.on_transparency_slider_move)
 
-        translation_transparency_slider_sizer.Add(
-            self.translation_transparency_slider,
-            proportion=1,
-            flag=wx.ALIGN_CENTER_VERTICAL)
+        translation_transparency_slider_sizer.AddF(
+            self.translation_transparency_slider, sizer_flags.Proportion(2).Expand())
 
         self.translation_transparency_value = wx.StaticText(
             self, label=self.label_for_transparency(configured_transparency))
 
-        translation_transparency_slider_sizer.Add(
-            self.translation_transparency_value,
-            flag=wx.ALIGN_CENTER_VERTICAL | wx.ALIGN_RIGHT)
+        translation_transparency_slider_sizer.AddF(
+            self.translation_transparency_value, sizer_flags.Proportion(0))
 
-        translation_transparency_sizer.Add(
-            translation_transparency_slider_sizer)
+        translation_transparency_sizer.AddF(
+            translation_transparency_slider_sizer, sizer_flags.Expand())
 
-        sizer.Add(translation_transparency_sizer)
+        sizer.AddF(translation_transparency_sizer, sizer_flags.Expand().Border(0))
 
-        self.SetSizer(sizer)
+        self.SetSizerAndFit(sizer)
 
     def save(self):
         """Write all parameters to the config."""
```

plover/config.py
jeremy-w: Darn, I thought I got them all. I'll do an ag transparen pass over the source tree.

--
Jeremy W. Sherman
http://jeremywsherman.com/

jeremy-w:
@benoit-pierre The issues you reported should be fixed now.

I noticed the display config window can resize arbitrarily far vertically and horizontally, which can leave a ton of dead space around. Does wxWidgets provide any way to say "don't grow space here"? I saw how to position it when space grows with alignment, and how to direct it to distribute space between items using proportion, but nothing that would contribute to calculating a max window size.

(The two layout models I'm accustomed to are the old Cocoa springs-and-struts model and the modern Cocoa constraint-based layout ("AutoLayout") approach, so the layout system used by wxWidgets is new to me.)

morinted:
Works well on Windows, thanks. But the actual layout seems unaligned on the left, does this happen for you, too?

alignment

jeremy-w:
Now that you mention it, yeah, it's slightly out of left alignment. I see the same on OS X.

jeremy-w:
Looks to be trouble with sizers. :\

jeremy-w:
Also a mix of UI_BORDER and COMPONENT_SPACE, which differ by 1 pt.

jeremy-w:
I think I've fixed it:

raggedleftfixed

One of the things that threw me off was that the SizerFlags had a fluent interface (just keep chaining calls), but each call was not returning a new sizer, just mutating it in place. This meant that some of our calls were stacking in undesirable ways as we reused and (tried to) adjust the base sizer flags.

The [Widget Inspection Tool](http://wiki.wxpython.org/Widget%20Inspection%20Tool) came in handy. I hacked it in like so:

```diff
diff --git a/plover/gui/main.py b/plover/gui/main.py
index 79496e7..4f721c5 100644
--- a/plover/gui/main.py
+++ b/plover/gui/main.py
@@ -14,6 +14,7 @@ import os
 import wx
 import wx.animate
 from wx.lib.utils import AdjustRectToScreen
+from wx.lib.inspection import InspectionTool
 import plover.app as app
 from plover.config import ASSETS_DIR, SPINNER_FILE
 from plover.gui.config import ConfigurationDialog
@@ -55,6 +56,7 @@ class PloverGUI(wx.App):
         frame = MainFrame(self.config)
         self.SetTopWindow(frame)
         frame.Show()
+        wx.lib.inspection.InspectionTool().Show()
         return True
``

File: plover/config.py [hunk 8]
```diff
@@ -383,7 +396,18 @@ def get_translation_frame_y(self):
         return self._get_int(TRANSLATION_FRAME_SECTION, 
                              TRANSLATION_FRAME_Y_OPTION,
                              DEFAULT_TRANSLATION_FRAME_Y)
-                             
+
+    def set_translation_frame_opacity(self, opacity):
```
morinted: @benoit-pierre has suggested just asserting the function input instead of correcting it, especially because this is an internal-facing call.

morinted:
It's looking pretty good on Windows now, thanks. I think it's pretty much ready for a squash&merge, just left the one comment about auto-fixing the config values.

@benoit-pierre has suggested that on reading the config, if the value is out of bounds, we should throw an exception (which right now isn't viable, but will be hopefully in an upcoming PR), and instead of correcting the values on write, we use assertions to provide exceptions for function callers.

plover/config.py
jeremy-w: Makes sense for setting. I'll throw the assert in.

My thinking was along the lines "bogus opacity configuration would be a lame reason for a program someone relies on to do their job crashing", but if we're setting a bogus value, we should catch that during testing, and it's totally our fault.

For reading a value, that could be someone flubbing a manual edit to their config file, and I'd like to be forgiving of that.

jeremy-w:
Looks like there are now conflicts too, so I'll resolve those along with adding the assert on set, then squash.

jeremy-w:
This has been updated and squashed down to a single commit. The commit message now summarizes the entire chunk of work across the various files.

plover/config.py
```diff
@@ -401,7 +416,21 @@ def get_translation_frame_y(self):
...
+        self._set(TRANSLATION_FRAME_SECTION, TRANSLATION_FRAME_OPACITY_OPTION,
+                  clamped)
+
+    def get_translation_frame_opacity(self):
+        raw = self._get_int(TRANSLATION_FRAME_SECTION,
+                            TRANSLATION_FRAME_OPACITY_OPTION,
+                            DEFAULT_TRANSLATION_FRAME_OPACITY)
+        clamped_value = clamp(raw, 0, 100)
+        return clamped_value
+
```
benoit-pierre: I would change this too:

```python
MIN_FRAME_OPACITY = 0
MAX_FRAME_OPACITY = 100

def check_opacity(opacity):
    if MIN_FRAME_OPACITY <= opacity <= MAX_FRAME_OPACITY:
        return opacity
    raise ValueError("opacity %u out of range [%u, %u]" % (opacity,
                                                           MIN_FRAME_OPACITY,
                                                           MAX_FRAME_OPACITY))

[...]

    def set_translation_frame_opacity(self, opacity):
        check_opacity(opacity)
        self._set(TRANSLATION_FRAME_SECTION,
                  TRANSLATION_FRAME_OPACITY_OPTION,
                  opacity)

    def get_translation_frame_opacity(self):
        opacity = self._get_int(TRANSLATION_FRAME_SECTION,
                                TRANSLATION_FRAME_OPACITY_OPTION,
                                DEFAULT_TRANSLATION_FRAME_OPACITY)
        try:
            check_opacity(opacity)
        except ValueError as e:
            log.error("translation %s, reset to %u", e,
                      DEFAULT_TRANSLATION_FRAME_OPACITY)
            opacity = DEFAULT_TRANSLATION_FRAME_OPACITY
        return opacity
```

benoit-pierre:
The layout of the slider is wrong again.

jeremy-w:
@benoit-pierre I'm going to set up a Linux VM so I can see what this stuff looks like under that platform. It seems like something wonky with WX that it should always be wrong only on that platform. :\ Lacking a sane minimum size on the slider perhaps?

benoit-pierre:
This seem to take care of the problem:
```diff
diff --git i/plover/gui/config.py w/plover/gui/config.py
index 10c8404..d9d704c 100644
--- i/plover/gui/config.py
+++ w/plover/gui/config.py
@@ -579,7 +579,7 @@ class DisplayConfig(wx.Panel):
               .Border(wx.LEFT | wx.BOTTOM | wx.RIGHT, UI_BORDER)
               .Expand())

-        sizer.Add(opacity_column_sizer, border=UI_BORDER, flag=wx.BOTTOM)
+        sizer.Add(opacity_column_sizer, border=UI_BORDER, flag=wx.BOTTOM|wx.EXPAND)

         self.SetSizerAndFit(sizer)

```

jeremy-w:
It seems to fill the window width then as you resize. Looks good to me.

jeremy-w:
@benoit-pierre Thanks for your help. The slider should resize appropriately now, and I've adopted the `ValueError` approach, as well.

benoit-pierre:
Good to merge for me.

Issue #263, opened 2015-09-25
Plover could use more art


panathea:
Right now the red "P" doesn't mean anything and is used arbitrarily as an indicator for Plover being on and off. A couple things that I don't like about the P:
- It is low resolution which looks bad on high density displays
- Red/green colorblind users probably don't benefit from the color change indication
- The red and green are very visually intense!
- Having the icon and the indicator being the same logo means that you may assume that Plover is always "off" because in your task bar and on your desktop, it is the red P.

I think that the Plover art could:
- Have more than just color to indicate its status
- Have a separate application icon. Status != application
- Use some of the Plover branding (which is still up in the air at this point, apart from Delores)

I'll be talking a little bit with my friend Jeff to see if he can fit this into his free time, but any contributions and suggestions are _totally welcome_.

morinted:
Ideas for an icon? The icon doesn't have to be the same as the indicator for on/off. Because if it is, then you associate Plover's icon with "on" or "off", which doesn't make sense. Also, in OS X and Windows, the icon is the red P, while in Linux it's the green one...

All right, let's see. Eclipse has ![their logo](http://www.eclipsecat.com/sites/default/files/imagecache/product/TotalEclipse_Diamond_800x800.png)

Very abstract, kind of ugly and outdated. I wouldn't pursue this. What about ProCAT Winner? ![](http://www.acculaw.com/ProductImages/software/procat.jpg)

I'd say that's a little better. At least they have their brand. No idea if they use that on the desktop though or if it's just box art.

Plover can obviously use a Plover bird, or it can be more abstract and professional such as a steno machine. Anything from a cartoon rendition to a general flat-style shape. Ideally it would be simple but recognizable. Any ideas for a launcher icon that would make you happy each time you see it? :)

lifton:
How about a stylized single steno key?

morinted:
I'd also like to leave a callback to the old Plover logo drafts from the [blog post](http://plover.stenoknight.com/2010/04/plover-logo-prototypes.html) which are:

![Plover birds](http://stenoknight.com/plover%20ideas.png)

Then there's Dolores who you can see in all her one-pose glory:

![Dolores](http://stenoknight.com/plover/ploverlogo.png)

lifton:
Maybe Dolores with squinted eyes and lightning in the background.

morinted:
[...]

PR #365, opened 2016-02-16
Redesign main frame

morinted:
- Remove bitmaps for status indicator, as part of the effort for #263.
- Windows has a grey color for the main frame because that is Windows' default color. To fix this, you place a top-level panel in the frame.
- Add a disabled state for output status when a machine is disconnected.
- Remove status from title-bar.
- Section off the user interface for clarity, needed especially because of the status indicator.
- Show status with radio buttons.

Here is the redesign:

screen shot 2016-02-20 at 6 08 22 pm
screen shot 2016-02-20 at 6 08 25 pm
screen shot 2016-02-20 at 6 08 30 pm

### DIFF
--- a/plover/gui/main.py
+++ b/plover/gui/main.py
@@ -73,9 +73,7 @@ class MainFrame(wx.Frame):
     DISCONNECTED_IMAGE_FILE = os.path.join(ASSETS_DIR, 'disconnected.png')
     REFRESH_IMAGE_FILE = os.path.join(ASSETS_DIR, 'refresh.png')
     BORDER = 5
-    RUNNING_MESSAGE = "running"
-    STOPPED_MESSAGE = "stopped"
-    ERROR_MESSAGE = "error"
+    STATUS_DISCONNECTED, STATUS_OUTPUT_ENABLED, STATUS_OUTPUT_DISABLED = range(3)
     CONFIGURE_BUTTON_LABEL = u"Configure…"
     ABOUT_BUTTON_LABEL = u"About…"
     RECONNECT_BUTTON_LABEL = u"Reconnect…"
@@ -96,15 +94,16 @@ class MainFrame(wx.Frame):
                           style=wx.DEFAULT_FRAME_STYLE & ~(wx.RESIZE_BORDER |
                                                            wx.RESIZE_BOX |
                                                            wx.MAXIMIZE_BOX))
+        self.top_panel = wx.Panel(self, style=wx.DEFAULT_FRAME_STYLE)
 
         # Status button.
         self.on_bitmap = wx.Bitmap(self.ON_IMAGE_FILE, wx.BITMAP_TYPE_PNG)
         self.off_bitmap = wx.Bitmap(self.OFF_IMAGE_FILE, wx.BITMAP_TYPE_PNG)
-        self.status_button = wx.BitmapButton(self, bitmap=self.on_bitmap)
+        self.status_button = wx.BitmapButton(self.top_panel, bitmap=self.on_bitmap)
         self.status_button.Bind(wx.EVT_BUTTON, self._toggle_steno_engine)
 
         # Configure button.
-        self.configure_button = wx.Button(self,
+        self.configure_button = wx.Button(self.top_panel,
                                           label=self.CONFIGURE_BUTTON_LABEL)
         self.configure_button.Bind(wx.EVT_BUTTON, self._show_config_dialog)
 
@@ -113,20 +112,20 @@ class MainFrame(wx.Frame):
         self.SetMenuBar(MenuBar)
 
         # About button.
-        self.about_button = wx.Button(self, label=self.ABOUT_BUTTON_LABEL)
+        self.about_button = wx.Button(self.top_panel, label=self.ABOUT_BUTTON_LABEL)
         self.about_button.Bind(wx.EVT_BUTTON, self._show_about_dialog)
 
         # Machine status.
         # TODO: Figure out why spinner has darker gray background.
-        self.spinner = wx.animate.GIFAnimationCtrl(self, -1, SPINNER_FILE)
+        self.spinner = wx.animate.GIFAnimationCtrl(self.top_panel, -1, SPINNER_FILE)
         self.spinner.GetPlayer().UseBackgroundColour(True)
         self.spinner.Hide()
 
-        self.connected_bitmap = wx.Bitmap(self.CONNECTED_IMAGE_FILE, 
+        self.connected_bitmap = wx.Bitmap(self.CONNECTED_IMAGE_FILE,
                                           wx.BITMAP_TYPE_PNG)
         self.disconnected_bitmap = wx.Bitmap(self.DISCONNECTED_IMAGE_FILE, 
                                              wx.BITMAP_TYPE_PNG)
-        self.connection_ctrl = wx.StaticBitmap(self, 
+        self.connection_ctrl = wx.StaticBitmap(self.top_panel,
                                                bitmap=self.disconnected_bitmap)
 
         # Layout.
@@ -157,18 +156,19 @@ class MainFrame(wx.Frame):
         longest_state = max((STATE_ERROR, STATE_INITIALIZING, STATE_RUNNING), 
                             key=len)
         longest_status = '%s: %s' % (longest_machine, longest_state)
-        self.machine_status_text = wx.StaticText(self, label=longest_status)
-        sizer.Add(self.machine_status_text, 
+        self.machine_status_text = wx.StaticText(self.top_panel, label=longest_status)
+        sizer.Add(self.machine_status_text,
                   flag=wx.BOTTOM | wx.RIGHT | wx.ALIGN_CENTER_VERTICAL,
                   border=self.BORDER)
         refresh_bitmap = wx.Bitmap(self.REFRESH_IMAGE_FILE, wx.BITMAP_TYPE_PNG)          
-        self.reconnect_button = wx.BitmapButton(self, bitmap=refresh_bitmap)
+        self.reconnect_button = wx.BitmapButton(self.top_panel, bitmap=refresh_bitmap)
         sizer.Add(self.reconnect_button, 
                   flag=wx.BOTTOM | wx.RIGHT | wx.ALIGN_CENTER_VERTICAL, 
                   border=self.BORDER)
         self.machine_status_sizer = sizer
         global_sizer.Add(sizer)
         self.SetSizer(global_sizer)
+        global_sizer.Fit(self.top_panel)
         global_sizer.Fit(self)
 
         self.Bind(wx.EVT_CLOSE, self._quit)
@@ -217,7 +217,7 @@ class MainFrame(wx.Frame):
             SuggestionsDisplayDialog.display(self, self.config, self.steno_engine)
 
     def consume_command(self, command):
-        # The first commands can be used whether plover is active or not.
+        # The first commands can be used whether plover has output enabled or not.
         if command == self.COMMAND_RESUME:
             wx.CallAfter(self.steno_engine.set_is_running, True)
             return True
@@ -232,7 +232,7 @@ class MainFrame(wx.Frame):
         if not self.steno_engine.is_running:
             return False
 
-        # These commands can only be run when plover is active.
+        # These commands can only be run when plover has output enabled.
         if command == self.COMMAND_SUSPEND:
             wx.CallAfter(self.steno_engine.set_is_running, False)
             return True
@@ -273,18 +273,28 @@ class MainFrame(wx.Frame):
             elif state == STATE_ERROR:
                 self.connection_ctrl.SetBitmap(self.disconnected_bitmap)
             self.machine_status_sizer.Layout()
-        if self.steno_engine.machine:
-            self.status_button.Enable()
-            if self.steno_engine.is_running:
-                self.status_button.SetBitmapLabel(self.on_bitmap)
-                self.SetTitle("%s: %s" % (self.TITLE, self.RUNNING_MESSAGE))
-            else:
-                self.status_button.SetBitmapLabel(self.off_bitmap)
-                self.SetTitle("%s: %s" % (self.TITLE, self.STOPPED_MESSAGE))
+        if not self.steno_engine.machine or self.steno_engine.machine.state is not STATE_RUNNING:
+            status = self.STATUS_DISCONNECTED
+        elif self.steno_engine.is_running:
+            status = self.STATUS_OUTPUT_ENABLED
         else:
-            self.status_button.Disable()
-            self.status_button.SetBitmapLabel(self.off_bitmap)
-            self.SetTitle("%s: %s" % (self.TITLE, self.ERROR_MESSAGE))
+            status = self.STATUS_OUTPUT_DISABLED
+        self._show_status(status)
+
+    def _show_status(self, status):
+        if status is self.STATUS_DISCONNECTED:
+            enabled = False
+            bitmap = self.off_bitmap
+        elif status is self.STATUS_OUTPUT_DISABLED:
+            enabled = True
+            bitmap = self.off_bitmap
+        elif status is self.STATUS_OUTPUT_ENABLED:
+            enabled = True
+            bitmap = self.on_bitmap
+        else:
+            raise ValueError('invalid status: %s' % str(status))
+        self.status_button.SetBitmapLabel(bitmap)
+        self.status_button.Enable(enabled)
 
     def _quit(self, event=None):
         if self.steno_engine:
--- a/plover/gui/main.py
+++ b/plover/gui/main.py
@@ -73,7 +73,12 @@ class MainFrame(wx.Frame):
     DISCONNECTED_IMAGE_FILE = os.path.join(ASSETS_DIR, 'disconnected.png')
     REFRESH_IMAGE_FILE = os.path.join(ASSETS_DIR, 'refresh.png')
     BORDER = 5
-    STATUS_DISCONNECTED, STATUS_OUTPUT_ENABLED, STATUS_OUTPUT_DISABLED = range(3)
+    STATUS_DISCONNECTED, STATUS_OUTPUT_DISABLED, STATUS_OUTPUT_ENABLED = range(-1, 2)
+    ENABLE_OUTPUT_LABEL = "Enable"
+    DISABLE_OUTPUT_LABEL = "Disable"
+    HEADER_OUTPUT = "Output"
+    HEADER_MACHINE = "Machine"
+    HEADER_SETTINGS = "Settings"
     CONFIGURE_BUTTON_LABEL = u"Configure…"
     ABOUT_BUTTON_LABEL = u"About…"
     RECONNECT_BUTTON_LABEL = u"Reconnect…"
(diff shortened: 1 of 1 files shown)


morinted:
The branch check failed on AppVeyor while uploading the artifact, but the second build succeeded, so ignore the failed build please.

plover/gui/main.py
```diff
@@ -91,19 +92,20 @@ def __init__(self, config):
         self.config = config
 
         # Note: don't set position from config, since it's not yet loaded.
-        wx.Frame.__init__(self, None, title=self.TITLE,
+        mainframe = wx.Frame.__init__(self, None, title=self.TITLE,
```
benoit-pierre: No need for this.

plover/gui/main.py
```diff
@@ -273,18 +276,31 @@ def _update_status(self, state):
...
+            self.SetTitle("%s: %s" % (self.TITLE, self.STOPPED_MESSAGE))
+        elif status is self.STATUS_STOPPED:
+            self.status_button.Enable()
+            self.status_button.SetBitmapLabel(self.off_bitmap)
+            self.SetTitle("%s: %s" % (self.TITLE, self.STOPPED_MESSAGE))
+        elif status is self.STATUS_RUNNING:
+            self.status_button.Enable()
+            self.status_button.SetBitmapLabel(self.on_bitmap)
+            self.SetTitle("%s: %s" % (self.TITLE, self.RUNNING_MESSAGE))
+
```
benoit-pierre: This is confusing: why show the stopped message when the machine is not successfully connected? How about this instead:

```diff
diff --git i/plover/gui/main.py w/plover/gui/main.py
index 5013620..bef7878 100644
--- i/plover/gui/main.py
+++ w/plover/gui/main.py
@@ -277,29 +277,33 @@ class MainFrame(wx.Frame):
             self.machine_status_sizer.Layout()
         if self.steno_engine.machine:
             if self.steno_engine.machine.state is not STATE_RUNNING:
-                self._show_status(self.STATUS_ERROR)
+                status = self.STATUS_ERROR
             elif self.steno_engine.is_running:
-                self._show_status(self.STATUS_RUNNING)
+                status = self.STATUS_RUNNING
             else:
-                self._show_status(self.STATUS_STOPPED)
+                status = self.STATUS_STOPPED
         else:
-            self.status_button.Disable()
-            self.status_button.SetBitmapLabel(self.off_bitmap)
-            self.SetTitle("%s: %s" % (self.TITLE, self.ERROR_MESSAGE))
+            status = self.STATUS_ERROR
+        self._show_status(status)

     def _show_status(self, status):
         if status is self.STATUS_ERROR:
-            self.status_button.Disable()
-            self.status_button.SetBitmapLabel(self.off_bitmap)
-            self.SetTitle("%s: %s" % (self.TITLE, self.STOPPED_MESSAGE))
+            enabled = False
+            bitmap = self.off_bitmap
+            message = self.ERROR_MESSAGE
         elif status is self.STATUS_STOPPED:
-            self.status_button.Enable()
-            self.status_button.SetBitmapLabel(self.off_bitmap)
-            self.SetTitle("%s: %s" % (self.TITLE, self.STOPPED_MESSAGE))
+            enabled = True
+            bitmap = self.off_bitmap
+            message = self.STOPPED_MESSAGE
         elif status is self.STATUS_RUNNING:
-            self.status_button.Enable()
-            self.status_button.SetBitmapLabel(self.on_bitmap)
-            self.SetTitle("%s: %s" % (self.TITLE, self.RUNNING_MESSAGE))
+            enabled = True
+            bitmap = self.on_bitmap
+            message = self.RUNNING_MESSAGE
+        else:
+            raise ValueError('invalid status: %s' % str(status))
+        self.status_button.SetBitmapLabel(bitmap)
+        self.SetTitle("%s: %s" % (self.TITLE, message))
+        (self.status_button.Enable if enabled else self.status_button.Disable)()

     def _quit(self, event=None):
         if self.steno_engine:
```

Also, using "stopped" does not seem right, since Plover is still processing strokes in the background. We really need to find another term.

morinted:
Instead of "stopped" we could change to Mirabai's suggestion of "disabled". Alternatively, I've used "suspended" and "paused" to describe Plover's not-running state.

benoit-pierre:
The problem is: it is still running, listening to keys, if you use a machine, the port is still in use. I think: "output enabled"/"output disabled" is the better terminology.

morinted:
Yes, I can agree with that. I may also rename, in this PR, "error" to "disconnected", since that's when the error message is used.

benoit-pierre:
What about in case of initialization error?

morinted:
All right, I updated.

Removing the window title, it becomes clear that we'll need a secondary label. I'm all for folding in my checkbox change into this PR, with our updated terminology.

So, we could do like the checkbox in the art thread:

- [ ] Enable Output
- [*] Output Enabled

Or perhaps a label & button?

```
Output Enabled
[ ⏸Disable ]

----

Output Disabled
[ ▶️ Enable ]
```

Label and a button, short form?

```
Output Enabled [ ⏸ ]

----

Output Disabled [ ▶️ ]
```

For the icons, we should just render some bitmaps from some open icon repository, as Unicode support differs depending on the user's system font. The first two that come to mind are [Google's Material Icons](https://design.google.com/icons/) and [Ionic's Ionicons](http://ionicons.com/).

jeremy-w:
Google's Material icons don't seem to have a standard play/pause, just stuff for snoozing notifications and pausing phone calls. The Ionicons look great. (We should do the notice & terms requirement of MIT code in the About dialog somewhere, probably.)

stenoknight:
This might seem arbitrary, but I have an aesthetic dislike of using
play/pause icons for anything other than actual audio playback. I just feel
like something accepting input and producing output isn't best represented
by a schema like play/pause. Just my opinion.

benoit-pierre:
t

benoit-pierre:
t

morinted:
If we added a group, we'd have to group the other elements (machine status and configure/about buttons) as well. I'd like to try:

```
Output Enabled    [ Disable ]

Output Disabled   [ Enable ]
```

jeremy-w:
Could we split out the background color change from the "what buttons do we want" bit that's still under discussion? I'm all for getting the better background color in, that's an uncontroversial win AFAICT.

morinted:
@benoit-pierre @jeremy-w 

I folded in my status button work into this PR. If I could get your feedback on functionality and code style, then we can get this done in time for the release.

jeremy-w:
I like the screenshots in the description. Far clearer than what we have now.

Achim63:
I beg to differ - I don't like that enabled-disable disabled-enable button at all. Since it's a simple switch, a button is simply the wrong UI element here. It should either be two mutually exclusive radio buttons, a checkbox, or (my preference) an on/off slider - something like this (found in Google pictures, so maybe copyrighted): 
08-mobile-on-off-toggle-switch-psd

morinted:
@Achim63 you can read through the art exchange if you'd like. There are various reasons for why we ended up on this design and I'm afraid that ultimately not everyone is going to be happy.

- The toggle buttons like you are showing, I've looked at them for hours, and they don't make sense if you can't drag them → which you can't with the bitmap buttons that wxPython has. So, knowing that you can't swipe, what's another toggle button?
- The other issue with a bitmap is the need to manage retina display on a Mac, but let's ignore that hurdle for the sake of argument.
- So, I started to look at toggle buttons that don't need sliding. Lots of classical "switches", with on/off on each end. That's where I got my two-button switch, where the active button gets "pushed-in". That one had some issues with general aesthetic, and it took up a lot of vertical space.
- The other discussion was terminology. We discussed what sort of words can be used to describe Plover's output state. Previously it was "running" and "stopped", which isn't true because Plover is always running, listening for toggle strokes and such. Therefore, we found that "output enabled/disabled" most succinctly described what Plover was doing.
- The title bar with the status in it was annoying on Windows because no matter what it was always cut off, so that was removed along the way in favor of a more verbose user interface in the main configuration dialogue.
- That leaves us with the need to convey status as well as convey what clicking a button will do. While not super-pretty or revolutionary or what-have-you, the grouped solution allows us to clearly section off different pieces of Plover's main window. The label states Plover's output state, using unambiguous words (enabled, disabled, disconnected → all adjectives) and the button labels its own action (enable, disable → verbs).
- We also tried a check box, but the problem is indicating what clicking the check box will do.

Finally, for the toggle button, I really recommend reading through this stack overflow response: http://ux.stackexchange.com/a/40954

As always, though, the code is iterative. If the feedback is overwhelming when we release a preview build, we can always change the user interface later.

morinted:
I haven't tried radio buttons, though, that might be worth a try.

jeremy-w:
I was in favor of giving radio buttons a shot, but I like the visual symmetry with all normal buttons now that I see that.

@Achim63 A lot of UI stuff is hamstrung by using the lowest-common denominator of what wxWidgets has to offer via wxPython. It's frustrating.

plover/gui/main.py
```diff
@@ -73,9 +73,15 @@ class MainFrame(wx.Frame):
     DISCONNECTED_IMAGE_FILE = os.path.join(ASSETS_DIR, 'disconnected.png')
     REFRESH_IMAGE_FILE = os.path.join(ASSETS_DIR, 'refresh.png')
     BORDER = 5
-    RUNNING_MESSAGE = "running"
-    STOPPED_MESSAGE = "stopped"
-    ERROR_MESSAGE = "error"
+    STATUS_DISCONNECTED, STATUS_OUTPUT_ENABLED, STATUS_OUTPUT_DISABLED = range(3)
```
jeremy-w: I'd expect these to be disconnected => -1, disabled => 0, enabled => 1, but since this isn't exposed outside the app yet, no big deal.

plover/gui/main.py
```diff
@@ -96,80 +102,112 @@ def __init__(self, config):
...
-                  flag=(wx.LEFT | wx.BOTTOM | wx.RIGHT | 
-                        wx.ALIGN_CENTER_VERTICAL), 
+                  flag=(wx.LEFT | wx.RIGHT |
+                        wx.ALIGN_CENTER_VERTICAL),
                   border=self.BORDER)
         longest_machine = max(machine_registry.get_all_names(), key=len)
-        longest_state = max((STATE_ERROR, STATE_INITIALIZING, STATE_RUNNING), 
+        longest_state = max((STATE_ERROR, STATE_INITIALIZING, STATE_RUNNING),
                             key=len)
         longest_status = '%s: %s' % (longest_machine, longest_state)
```
jeremy-w: Could we stick with `state` for machine state by making this `longest_state`? There's a terminology mismatch here between `state` vs `status`.

We also already have a `longest_status` in scope, which made me wonder why we were recomputing it initially.

plover/gui/main.py
```diff
@@ -217,7 +255,7 @@ def __init__(self, config):
             SuggestionsDisplayDialog.display(self, self.config, self.steno_engine)
 
     def consume_command(self, command):
-        # The first commands can be used whether plover is active or not.
+        # The first commands can be used whether plover has output enabled or not.
```
jeremy-w: +:1+ on the active => output enabled change

jeremy-w:
:+1: Looks good to me!

jeremy-w:
Just used this branch while doing some drills. Working fine on Mac.

Achim63:
Too bad I'm not much of a programmer. We might switch to something like Kivy to get rid of the wx stuff. With the added bonus of being able to run on iOS and Android. I think a complete steno keyboard would fit easily on an iPad Pro (I tried that on a standard iPad and it's a bit too small). Just wishful thinking, maybe.

stenoknight:
Steno Keyboard for Android needs some bugfixing, but it's otherwise a
pretty tolerable little app. Works both with an on-screen keyboard
(Swype-style, though it would be a lot more usable if the keys could be
swiped off after being swipped on without breaking the stroke) and with a
hardware keyboard using USBOTG.

https://play.google.com/store/apps/details?id=com.brentandjody.stenoime&hl=en

On Thu, Feb 18, 2016 at 3:01 PM, Achim <notifications@github.com> wrote:

> Too bad I'm not much of a programmer. We might switch to something like
> Kivy to get rid of the wx stuff. With the added bonus of being able to run
> on iOS and Android. I think a complete steno keyboard would fit easily on
> an iPad Pro (I tried that on a standard iPad and it's a bit too small).
> Just wishful thinking, maybe.
>
> —
> <https://github.com/openstenoproject/plover/pull/365#issuecomment-185891715>
> .
>



-- 
Mirabai Knight, CRC, RDR
StenoKnight CART Services
917 576 4989
mkk@stenoknight.com
http://stenoknight.com

Achim63:
I tried this one: https://itunes.apple.com/de/app/istenopad/id399707498?mt=8
But as mentioned, it's a bit too small for me, but it doesn't need sliding, you can type like on a steno machine.

Regarding the Plover window, I'd much prefer a smaller, more minimalistic approach. The "About" and "Settings" items could even reside just in the menubar like in other apps.
plover_minimal

morinted:
Please post art ideas in the art thread, you'll get more eyes there and a redesign won't find its way into this PR.

Achim63:
This is mainly regarding this pull request - I simply think that the output section with it's disabled/enable/enabled/disable texts/buttons is too confusing, whereas the checkbox approach is rather clear. I tried out this branch and the Plover window is taking up too much screen estate IMHO (and I'm on a 27' iMac).

morinted:
Ultimately we'd like to have the option to close to a tray icon.

Achim63:
That would be optimal! I already have 17 items on the right of my menubar, and I wouldn't mind one more ;-)

morinted:
I totally agree. Until then, this in-progress is mainly to help us get rid
of the placeholder red/green P that Plover has held for far too long.

As far as size, it is a little bigger than Plover was, but Plover isn't a
particularly large user interface even after these changes. And, as always,
things can be refined when we get more feedback. I will mock out the radio
buttons soon though, I'm curious to see how that would look.

stenoknight:
I always assumed that everyone just ran Plover with the window minimized
whenever they weren't directly interacting with its configuration options.
No?

On Thu, Feb 18, 2016 at 4:11 PM, Ted Morin <notifications@github.com> wrote:

> I totally agree. Until then, this in-progress is mainly to help us get rid
> of the placeholder red/green P that Plover has held for far too long.
>
> As far as size, it is a little bigger than Plover was, but Plover isn't a
> particularly large user interface even after these changes. And, as always,
> things can be refined when we get more feedback. I will mock out the radio
> buttons soon though, I'm curious to see how that would look.
>
> On Thu, Feb 18, 2016 at 4:09 PM Achim <notifications@github.com> wrote:
>
> > That would be optimal! I already have 17 items on the right of my
> menubar,
> > and I wouldn't mind one more ;-)
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/openstenoproject/plover/pull/365#issuecomment-185922568
> >
> > .
>
> >
>
> —
> <https://github.com/openstenoproject/plover/pull/365#issuecomment-185923696>
> .
>



-- 
Mirabai Knight, CRC, RDR
StenoKnight CART Services
917 576 4989
mkk@stenoknight.com
http://stenoknight.com

Achim63:
That depends - when I'm running Stentura mode I like to leave it open to see at a glance if it's connected, since I have to reconnect every time I switch on my Cybra. When I'm using TX Bolt, I usually minimize it to the dock.

jeremy-w:
Wonder if we could listen for hardware attach notifications and try to reconnect automatically…
--
Jeremy W. Sherman
http://jeremywsherman.com/

morinted:
@jeremy-w that's a good idea, certainly. I also wouldn't mind a reconnect button to at least trigger it manually...

BUT this has been updated with the radio buttons. I think this is it. After a code review, I'll squash the changes and we can merge if all is well.

morinted:
All right, trying to make the window a little smaller in response to feedback, also rewrote the layout logic with wxSizerFlags for brevity and clarity. Just removed the 2 extra groups and played with the resulting spacing to get things looking good.

screen shot 2016-02-19 at 10 35 32 am
screen shot 2016-02-19 at 10 35 34 am
screen shot 2016-02-19 at 10 35 40 am

Achim63:
That size seems much better, thank you. We'd need an UI/UX expert ... the "Output" label still seems a bit lost at its current location. Maybe I'll ask one of my UI colleagues next week if they have an idea.

jeremy-w:
It would look less lost without buttons above it. Configure/Preferences… and About are usually app menu items - that would be an option to move them out of the main UI. About is especially unlikely to be clicked, but it has a lot of prominence in the UI now.

jeremy-w:
To be clear, I don't want any other changes to hold this up. It looks good. We can work over the UI further in future PRs.

morinted:
Thanks @jeremy-w. I'd like to do more with the UI and its organization, for sure, such as filling the menu bar on the Mac, and separating the machine config from the steno config. I welcome people to refine the UI further... it's a tough one when everyone has different opinions.

morinted:
@benoit-pierre reports that this renders very badly on Ubuntu, so it will require additional work before we can merge it.

jeremy-w:
@benoit-pierre Could you attach a screenshot of this under Ubuntu?

jeremy-w:
TFW you think "not the default button, must be Cancel" and instead close a PR *and* leave the comment you were trying to cancel.

benoit-pierre:
So the issue if not as clear cut as "it never looks good on Ubuntu", this only happen when I use a configuration with only one small dictionary (for faster startup). It otherwise looks fine, except for the radio buttons that are a bit cramped (vertically) with my theme on Arch Linux:

2016-02-19-205023_242x125_scrot

But this is a minor issue, so let's merge this.

morinted:
@benoit-pierre I rejigged the UI slightly to make it look good on Windows, I'll have to confirm Mac looks all right, does this also fix the crampedness on your machine?

benoit-pierre:
@morinted: yes.

benoit-pierre:
With those additional changes, the layout is always correct, no matter the configuration, and whether I start with the keyboard and switch to another machine or vice versa:
```diff
diff --git i/plover/gui/main.py w/plover/gui/main.py
index cd968a2..4afb9bb 100644
--- i/plover/gui/main.py
+++ w/plover/gui/main.py
@@ -99,34 +99,34 @@ class MainFrame(wx.Frame):
			   style=wx.DEFAULT_FRAME_STYLE & ~(wx.RESIZE_BORDER |
							    wx.RESIZE_BOX |
							    wx.MAXIMIZE_BOX))
-	 self.root = wx.Panel(self, style=wx.DEFAULT_FRAME_STYLE)
+	 root = wx.Panel(self, style=wx.DEFAULT_FRAME_STYLE)

	 # Menu Bar
	 MenuBar = wx.MenuBar()
	 self.SetMenuBar(MenuBar)

	 # Configure button.
-	 self.configure_button = wx.Button(self.root,
+	 self.configure_button = wx.Button(root,
					   label=self.CONFIGURE_BUTTON_LABEL)
	 self.configure_button.Bind(wx.EVT_BUTTON, self._show_config_dialog)

	 # About button.
-	 self.about_button = wx.Button(self.root, label=self.ABOUT_BUTTON_LABEL)
+	 self.about_button = wx.Button(root, label=self.ABOUT_BUTTON_LABEL)
	 self.about_button.Bind(wx.EVT_BUTTON, self._show_about_dialog)

	 # Status radio buttons.
-	 self.radio_output_enable = wx.RadioButton(self.root,
+	 self.radio_output_enable = wx.RadioButton(root,
						   label=self.ENABLE_OUTPUT_LABEL)
	 self.radio_output_enable.Bind(wx.EVT_RADIOBUTTON,
				       lambda e: self.steno_engine.set_is_running(True))
-	 self.radio_output_disable = wx.RadioButton(self.root,
+	 self.radio_output_disable = wx.RadioButton(root,
						    label=self.DISABLE_OUTPUT_LABEL)
	 self.radio_output_disable.Bind(wx.EVT_RADIOBUTTON,
					lambda e: self.steno_engine.set_is_running(False))

	 # Machine status.
	 # TODO: Figure out why spinner has darker gray background.
-	 self.spinner = wx.animate.GIFAnimationCtrl(self.root, -1, SPINNER_FILE)
+	 self.spinner = wx.animate.GIFAnimationCtrl(root, -1, SPINNER_FILE)
	 self.spinner.GetPlayer().UseBackgroundColour(True)
	 self.spinner.Hide()

@@ -134,7 +134,7 @@ class MainFrame(wx.Frame):
					   wx.BITMAP_TYPE_PNG)
	 self.disconnected_bitmap = wx.Bitmap(self.DISCONNECTED_IMAGE_FILE,
					      wx.BITMAP_TYPE_PNG)
-	 self.connection_ctrl = wx.StaticBitmap(self.root,
+	 self.connection_ctrl = wx.StaticBitmap(root,
						bitmap=self.disconnected_bitmap)

	 border_flag = wx.SizerFlags(1).Border(wx.ALL, self.BORDER)
@@ -147,30 +147,30 @@ class MainFrame(wx.Frame):
	 settings_sizer.AddF(self.about_button, border_flag.Expand())

	 # Create Output Status Box
-	 box = wx.StaticBox(self.root, label=self.HEADER_OUTPUT)
+	 box = wx.StaticBox(root, label=self.HEADER_OUTPUT)
	 status_sizer = wx.StaticBoxSizer(box, wx.HORIZONTAL)

	 status_sizer.AddF(self.radio_output_enable, border_flag)
	 status_sizer.AddF(self.radio_output_disable, border_flag)

	 # Create Machine Status Box
-	 self.machine_sizer = wx.BoxSizer(wx.HORIZONTAL)
+	 machine_sizer = wx.BoxSizer(wx.HORIZONTAL)
	 center_flag =\
	     wx.SizerFlags()\
	       .Align(wx.ALIGN_CENTER_VERTICAL)\
	       .Border(wx.LEFT | wx.RIGHT, self.BORDER)

-	 self.machine_sizer.AddF(self.spinner, center_flag)
-	 self.machine_sizer.AddF(self.connection_ctrl, center_flag)
+	 machine_sizer.AddF(self.spinner, center_flag)
+	 machine_sizer.AddF(self.connection_ctrl, center_flag)
	 longest_machine = max(machine_registry.get_all_names(), key=len)
	 longest_state = max((STATE_ERROR, STATE_INITIALIZING, STATE_RUNNING),
			     key=len)
	 longest_machine_status = '%s: %s' % (longest_machine, longest_state)
-	 self.machine_status_text = wx.StaticText(self.root, label=longest_machine_status)
-	 self.machine_sizer.AddF(self.machine_status_text, center_flag)
-	 refresh_bitmap = wx.Bitmap(self.REFRESH_IMAGE_FILE, wx.BITMAP_TYPE_PNG)
-	 self.reconnect_button = wx.BitmapButton(self.root, bitmap=refresh_bitmap)
-	 self.machine_sizer.AddF(self.reconnect_button, center_flag)
+	 self.machine_status_text = wx.StaticText(root, label=longest_machine_status)
+	 machine_sizer.AddF(self.machine_status_text, center_flag)
+	 refresh_bitmap = wx.Bitmap(self.REFRESH_IMAGE_FILE, wx.BITMAP_TYPE_PNG)
+	 self.reconnect_button = wx.BitmapButton(root, bitmap=refresh_bitmap)
+	 machine_sizer.AddF(self.reconnect_button, center_flag)

	 # Assemble main UI
	 global_sizer = wx.GridBagSizer(vgap=self.BORDER, hgap=self.BORDER)
@@ -183,7 +183,7 @@ class MainFrame(wx.Frame):
			  border=self.BORDER,
			  pos=(1, 0),
			  span=(1, 2))
-	 global_sizer.Add(self.machine_sizer,
+	 global_sizer.Add(machine_sizer,
			  flag=wx.CENTER | wx.ALIGN_CENTER | wx.EXPAND | wx.LEFT,
			  pos=(2, 0),
			  border=self.BORDER,
@@ -193,8 +193,8 @@ class MainFrame(wx.Frame):
	 border = wx.BoxSizer()
	 border.AddF(global_sizer,
		     wx.SizerFlags(1).Border(wx.ALL, self.BORDER).Expand())
-	 border.Fit(self.root)
-	 self.SetSizerAndFit(border)
+	 root.SetSizerAndFit(border)
+	 border.SetSizeHints(self)

	 self.Bind(wx.EVT_CLOSE, self._quit)
	 self.Bind(wx.EVT_MOVE, self.on_move)
@@ -297,7 +297,6 @@ class MainFrame(wx.Frame):
		 self.connection_ctrl.SetBitmap(self.connected_bitmap)
	     elif state == STATE_ERROR:
		 self.connection_ctrl.SetBitmap(self.disconnected_bitmap)
-	     self.machine_sizer.Layout()
	 if not self.steno_engine.machine or self.steno_engine.machine.state is not STATE_RUNNING:
	     status = self.STATUS_DISCONNECTED
	 elif self.steno_engine.is_running:
```
The important change is the `SetSizerAndFit`/`SetSizeHints` part, the rest is just follow-up cleanups.

benoit-pierre:
And it works fine on Windows too, let's hope it does on Mac too...

morinted:
Yes, seems great, as usual. Thanks! It's been applied, works on Mac, and should be ready to submit.

PR #366, opened 2016-02-16
Replace triple-dot with ellipsis in user strings

jeremy-w:
Fixes visual appearance of too-wide triple-dot.

Required setting a coding through a
[magic comment](https://www.python.org/dev/peps/pep-0263/)
in order to avoid ugly raw Unicode escapes.

### DIFF
--- a/plover/gui/serial_config.py
+++ b/plover/gui/serial_config.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 # Copyright (c) 2010 Joshua Harlan Lifton.
 # See LICENSE.txt for details.
 
@@ -20,7 +21,7 @@ RTS_CTS_STR = 'RTS/CTS'
 XON_XOFF_STR = 'Xon/Xoff'
 OK_STR = 'OK'
 SCAN_STR = "Scan"
-LOADING_STR = "Scanning ports..."
+LOADING_STR = u"Scanning ports…"
 CANCEL_STR = 'Cancel'
 CONNECTION_STR = 'Connection'
 PORT_STR = 'Port'
--- a/plover/gui/config.py
+++ b/plover/gui/config.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 # Copyright (c) 2010-2011 Joshua Harlan Lifton.
 # See LICENSE.txt for details.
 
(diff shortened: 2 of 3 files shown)


morinted:
This needs to be tested on Windows and Linux.

plover/gui/serial_config.py
```diff
@@ -90,7 +91,7 @@ def __init__(self, serial, parent, config):
         self.gif = wx.animate.GIFAnimationCtrl(self, -1, SPINNER_FILE)
         self.gif.GetPlayer().UseBackgroundColour(True)
         # Need to call this so the size of the control is not
-        # messed up (100x100 instead of 16x16) on Linux...
+        # messed up (100x100 instead of 16x16) on Linux…
```
benoit-pierre: I'm okay with updating the labels, but there is really no need to patch comments...

benoit-pierre:
The display is fine on Linux, but not on my Windows 10 VM.

benoit-pierre:
capture

morinted:
It might be necessary to convert to ASCII for use on Windows.
http://www.ascii-code.com/

Not sure exactly what the process involved would be.

morinted:
Maybe try defining the labels with u'String' ?

benoit-pierre:
I don't think that will change anything, the wxPython build are Unicode, and the strings are supposed to be automatically converted when passed to wx functions, no?

benoit-pierre:
Except they are not, or maybe from the wrong encoding, as using Unicode strings fix the problem!

morinted:
I think the UTF-8 encoding is the problem. Try removing the encoding, then
replacing the … with unichr(0x2026)? That should work just the same… Maybe

benoit-pierre:
From the wxPython doc:

> Strings passed as parameters to wx functions and methods will first be converted to a unicode object using the default encoding

> UPDATE: Starting with wxPython 2.5.4.1 the codec used to convert string objects to/from unicode objects can be set by user code using the wx.SetDefaultPyEncoding(encoding) function. Also the default codec used is the value of the locale.getdefaultlocale()[1] expression, or it will fall back to sys.getdefaultencoding() if there is a problem with the default locale settings. 

```shell
> echo '…' | iconv -f cp1252 -t utf-8
â€¦
```

morinted:
So wx.SetDefaultPyEncoding('utf-8')?

benoit-pierre:
No, just make sure those are Unicode string by adding `u`.

plover/gui/serial_config.py
jeremy-w: It was a regex replacement in the file. I thought the comment getting caught up was humorous, but I'll drop it out.

jeremy-w:
Updated to use Unicode strings. Appropriate fireballs despatched towards legacy Python. ;)

benoit-pierre:
OK, works fine in Linux (including with non-UTF-8 locales) and Windows. :+1:

morinted:
\#shipit

PR #367, opened 2016-02-16
update AppVeyor support so we can easily do weekly pre-releases

benoit-pierre:
When pushing to a "weekly" branch, a new pre-release will be automatically deployed to GitHub.

The release will looks [like this](https://github.com/benoit-pierre/plover/releases/tag/weekly-v2.5.8+344.g90c6102).

Note: the version is patched in, so the about dialog will show it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,23 @@
+version: '{build}'
+
+skip_tags: true
+
+pull_requests:
+
+  do_not_increment_build_number: true
+
+deploy:
+
+- provider: GitHub
+  release: weekly-v$(appveyor_build_version)
+  description: Weekly Pre-release
+  prerelease: true
+  auth_token:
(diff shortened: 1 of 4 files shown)


morinted:
I'll review this. Is there any way for me to patch in the version to the OSX version?

benoit-pierre:
After installing Python, you could use: `python windows/helper.py patch_version`. And `python windows/helper.py show_version` if you want to get the version to set a variable.

benoit-pierre:
OK, I lied... Using `windows/helper.py` will not work on Mac OS X, but I can make it work if you want?

morinted:
I'd appreciate it!

benoit-pierre:
OK, I moved the version stuff to `setup.py`:

```shell
> ./setup.py patch_version
running patch_version
> ./setup.py --version
2.5.8+336.gc5c8f2c
```

morinted:
You've already tested this in your branch, so until we push a weekly there'll be no way to tell that it works for sure without trying it. Merge now, I guess we'll try it after getting the OSX one in.

PR #368, opened 2016-02-16
Fix a Mac crash relating to long Unicode after switching to 64-bit

morinted:
We switched over to the PyObjc version of set Unicode string when going to 64-bit, and this broke a work around that [Hesky had figured out](http://stackoverflow.com/questions/18027342/how-can-i-call-cgeventkeyboardsetunicodestring-from-python) a while back. With the current dependencies, we don't need the work around.

In master right now, emojis crash output, this change fixes that.

Tested on Mac.

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -29,6 +29,7 @@ from Quartz import (
     kCGKeyboardEventKeycode,
     kCGSessionEventTap,
 )
+import Foundation
 import threading
 import collections
 import sys
@@ -291,11 +292,9 @@ def characters(s):
         character = encoded[start:end].decode('utf-32-be')
         yield character
 
-native_utf16 = 'utf-16-le' if sys.byteorder == 'little' else 'utf-16-be'
-
 def set_string(event, s):
-    buf = s.encode(native_utf16)
-    CGEventKeyboardSetUnicodeString(event, len(buf) / 2, buf)
+    buf = Foundation.NSString.stringWithString_(s)
+    CGEventKeyboardSetUnicodeString(event, len(buf), buf)
 
 class KeyboardEmulation(object):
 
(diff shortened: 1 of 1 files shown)


benoit-pierre:
So does this fix some of the crashes mentioned in #353 ?

morinted:
No, those issues are still present, unfortunately.

benoit-pierre:
Damn...

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -285,18 +286,15 @@ def suppress_keyboard(self, suppressed_keys=()):
...
         character = encoded[start:end].decode('utf-32-be')
         yield character
 
-native_utf16 = 'utf-16-le' if sys.byteorder == 'little' else 'utf-16-be'
-
 def set_string(event, s):
-    buf = s.encode(native_utf16)
-    CGEventKeyboardSetUnicodeString(event, len(buf) / 2, buf)
+    buf = Foundation.NSString.stringWithString_(s)
+    CGEventKeyboardSetUnicodeString(event, len(buf), buf)
```
jeremy-w: Does this automatically convert the NSString to a buffer of unichar for us? I'd have expected to need to do `string.getCharacters_range_()` with a buffer and an appropriately sized range to massage the data into the expected format, but I'm not terribly familiar with PyObjC.

morinted: Actually, when I made an ord array, PyObjc yelled at me, telling me it needed a string as argument 3, so honestly this stuff feels a little bit like voodoo. I always sacrifice a lamb before touching `osxkeyboardcontrol.py`

jeremy-w:
@morinted What further testing does this need, since it only affects Mac, and that's where you tested?

morinted:
I was more just looking for a code review, which you seem to have done, and also to run with it for the day. So far, it seems all fine. The only thing I noticed was I had one or two misplaced characters, which is an issue that crops up now and then, usually when the Macbook is under heavy load: as in, I'll get an output such as "kebyoard". I'll merge this for now as I think all is good!

PR #370, opened 2016-02-16
Shorten unicode on Add Translation dialog

morinted:
Prevent another Unicode-caused crash with the shorten Unicode function on Mac:

screen shot 2016-02-16 at 6 22 02 pm

Only tested on Mac.

### DIFF
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -145,6 +145,7 @@ class AddTranslationDialog(wx.Dialog):
                 label = '%s maps to %s' % (strokes, translation)
             else:
                 label = '%s is not in the dictionary' % strokes
+            label = util.shorten_unicode(label)
         else:
             label = ''
         self.stroke_mapping_text.SetLabel(label)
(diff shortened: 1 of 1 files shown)


morinted:
@jeremy-w Could you test this out on your Mac for me?

jeremy-w:
I'll check it when I get some time tonight.

What was the original crash you were seeing?

morinted:
Just try putting an emoji in the Translation box, it should crash immediately.

jeremy-w:
Yup, it does. It's frustrating – we should be able to set the font encoding to UTF-8 and pass that out, but it doesn't seem to work when I try. :(

morinted:
Yeah, for now we'll use workarounds until we find a more permanent fix.

jeremy-w:
Verified not crashing. Merging. Pushed some notes and a test harness for hacking away at the encoding issues [in this gist.](https://gist.github.com/jeremy-w/9683173a5c0a273e9524) Short version: It looks like a core Python function that's supposed to spit out `wchar_t` from a Python Unicode object isn't doing what wxPython thinks it does. wxWidgets expects its `wchar_t` to be utf-32-native, but the Python function is spitting out utf-16-le half-word stuffed in the high bytes of an otherwise zeroed word.

jeremy-w:
Finished analysis of the wxPython Unicode handling issue. [Write-up and test files are in this gist.](https://gist.github.com/jeremy-w/9683173a5c0a273e9524) The Unicode => wchar_t routine provided by Python2 doesn't handle codepoints outside the Basic Multilingual Plane, just braindead UCS-2, so no emoji for us as long as wxWidgets is built to use wchar_t rather than UTF-8 as its string representation.

PR #371, opened 2016-02-17
Add Linux support to Travis CI

benoit-pierre:
* add Linux to the build matrix
* run test suite for both OSX and Linux builds
* add egg and wheel deployments to the Linux build
* reduce overlap in dependencies handling when possible by using `setup.py develop` to install those that are supported by pip

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,65 @@
-language: python
+language: generic
+
 matrix:
   include:
+    - os: linux
+      dist: trusty
+      sudo: required
     - os: osx
-      language: generic
+
 before_install:
- - >
-  if [[ "$TRAVIS_OS_NAME" == "osx" ]];
-  then
-    brew update &&
-    brew install python --framework &&
-    brew link --overwrite python &&
-    brew install wxpython &&
-    (cd osx && ./bootstrap.sh); fi
+  - |
+    case "$TRAVIS_OS_NAME" in
+    osx)
+      brew update &&
+      brew install python --framework &&
+      brew link --overwrite python &&
+      brew install wxpython &&
+      true
+      ;;
+    linux)
+      sudo apt-add-repository -y ppa:adamwolf/kicad-trusty-backports &&
+      sudo apt-get update -qq &&
+      sudo apt-get install -y cython libudev-dev libusb-1.0-0-dev python-dev python-wxgtk3.0 python-xlib &&
+      true
+      ;;
+    esac &&
+    python -c 'import setup; setup.write_requirements()' &&
+    pip --disable-pip-version-check --timeout=5 --retries=2 install --upgrade --user -r requirements.txt &&
+    # We need to downgrade setuptools for py2app to work correctly...
+    pip --disable-pip-version-check --timeout=5 --retries=2 install --upgrade --user 'setuptools==19.2' &&
+    pip --disable-pip-version-check list &&
+    true
 
 install: true
+
 script:
+  - depth=50; while ! git describe --tag '--match=v[0-9]*'; do depth=$(($depth+10)); git fetch --depth $depth; done
   - python setup.py patch_version
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then (cd osx && make); fi
+  - python setup.py test
+  - |
+    case "$TRAVIS_OS_NAME" in
+    osx)
+      make -C osx &&
+      true
+      ;;
+    linux)
+      python setup.py bdist_egg &&
+      python setup.py bdist_wheel &&
+      true
+      ;;
+    esac
+
 deploy:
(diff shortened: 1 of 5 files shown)


morinted:
Is it the case that there are less Linux servers available for Travis?

benoit-pierre:
I don't know, but the builds seem to be queued for longer.

I've had a few transient errors too, something I never had with AppVeyor...

I wonder if we could not speed up those build, by detecting we're not going to deploy, and trim the dependencies; like I did in my first pass at using Travis: for example we don't need wxPython for the test suite.

morinted:
Would you be in favor of merging this and then optimizing later?

morinted:
Also, can you send over a .dmg that was produced from this?

benoit-pierre:
See an example [here](https://github.com/benoit-pierre/plover/releases/tag/weekly-v2.5.8+354.g711c5d1).

Notice how the numbering is different, but the hash is the same? Annoying... I guess this is because Travis CI does a shallow clone. This is why I had to add the fetch loop to retrieve the branch history first reachable tag:

```shell
depth=50; while ! git describe --tag '--match=v[0-9]*'; do depth=$(($depth+10)); git fetch --depth $depth; done
```

benoit-pierre:
And it looks like the '+' are getting replaced by '-'.... It should be `Plover-2.5.8+354.g711c5d1.exe`, not `Plover-2.5.8-354.g711c5d1.exe`. This is GitHub doing it seems, as the artifacts are correctly named on AppVeyor/Travis CI... Which is not good for the egg, as it should not be renamed.

morinted:
That's odd, especially since it doesn't mind the first dash.

benoit-pierre:
But the dashes are not the problem, the plus are. And for the binaries only (the source archives are fine). I tried manually editing the release and it does not work either. I contacted GitHub support.

morinted:
And we can't replace the plus signs with something else?

For the OSX build, it doesn't launch. Can you require setuptools==19.2 for OSX? The `requirements.txt` had `setuptools == 19.2` to fix this crash. The brewed Python comes with 19.4, upgrading to 20+ also didn't resolve the issue.

Also, since we are grabbing dependencies through `setup.py develop`, maybe we can replace the requirements.txt in bootstrap.sh with that command (@jeremy-w)

jeremy-w:
@benoit-pierre You might be able to replace the loop with a single `git pull --unshallow`, which will convert a shallow repo to one with full history.

We probably should switch to that in bootstrap, so that all the dependencies end up in one place.

benoit-pierre:
> And we can't replace the plus signs with something else?

Not without breaking PEP compatibility.

> For the OSX build, it doesn't launch. Can you require setuptools==19.2 for OSX? The requirements.txt had setuptools == 19.2 to fix this crash. The brewed Python comes with 19.4, upgrading to 20+ also didn't resolve the issue.

Oops, missed that, it's the same issue as with PyInstaller then... Will amend.

> You might be able to replace the loop with a single git pull --unshallow, which will convert a shallow repo to one with full history.

It's not exactly the same, as I'm not fetching the whole history. If we simplify the builds, then we don't need to get the last tag for most builds, because we don't need to patch the version since there is not point in building the Mac OS X app / Linux egg/wheel if we don't deploy them somewhere.

benoit-pierre:
OK, please check if the dmg [available here](https://github.com/benoit-pierre/plover/releases/tag/weekly-v2.5.8+355.gba76d69) works, and I'll amend this branch with the [required change](https://github.com/openstenoproject/plover/commit/ba76d696014834127dede66b77197eb1d3a702a9).

morinted:
Okay, two things.

I think the encoding settings are lost with py2app: `SyntaxError: Non-ASCII character '\xcf' in file MacOS/Plover on line 1, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details`. It might be better to revert to using unichar instead of directly encoding in UTF-8, but that's just my opinion.

Second, I *think* I'm still getting the import error, but it's a little hard to tell with this error going off. I'll have to do some local testing tonight to see if I can fix the UTF-8 error.

benoit-pierre:
OK, it would be good to confirm if setuptools still needs to be downgraded when pinning mock to 1.0.1:

```shell
    # Don't use a more recent version of mock or
    # we run into issues with setuptools and six...
    pip install --user --upgrade 'mock==1.0.1' pytest &&
```

(It's not possible to downgrade setuptools when using the system Python on the Linux environment)

morinted:
Well, what's worse is I'm not getting this issue when building locally.

benoit-pierre:
OK, everything should be working now. Example release output [here](https://github.com/benoit-pierre/plover/releases/tag/weekly-v2.5.8+365.g0ad8ef8).

PR #374, opened 2016-02-19
Override Cancel labels, fix #373

### LINKED ISSUE #373: Cmd + C closes dialog boxes.
#### What I expected to happen

I expected Cmd + C to let me copy text from a text field.

#### What happened instead

Whenever I press Cmd + C to copy text in the add translation dialogue, it would close the define window instead of copying the text. I noticed it happening also with other dialogue boxes containing text fields. For instance, if I try to copy the path to the logs to paste into Finder, it closes the dialogue. Plover remains functional and running, though. I've been getting around it by copying/pasting exclusively by right-clicking and selecting copy and paste from the menu.

#### Specs

OS X 10.11.3
Plover 2.5.8
Lightspeed steno machine
TX protocol

morinted:
- Default label for Cancel labels is '&Cancel', meaning any time
  the user tries to copy on OSX (Cmd-C), the active window would
  "cancel".
- See wx bug: http://trac.wxwidgets.org/ticket/15678

### DIFF
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -27,7 +27,7 @@ class AddTranslationDialog(wx.Dialog):
         self.strokes_text = wx.TextCtrl(self, style=wx.TE_PROCESS_ENTER)
         self.translation_text = wx.TextCtrl(self, style=wx.TE_PROCESS_ENTER)
         button = wx.Button(self, id=wx.ID_OK, label='Add to dictionary')
-        cancel = wx.Button(self, id=wx.ID_CANCEL)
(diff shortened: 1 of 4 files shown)


benoit-pierre:
What about buttons that are created with `self.CreateButtonSizer(wx.OK|wx.CANCEL)` (which should really be the preferred way instead of doing it manually ourselves) ?

morinted:
Yes, the same thing happens. Command-C still causes the dialog to close, I didn't "fix" this in the `keyboard_config.py` because there's no copyable area.

benoit-pierre:
Great... Another nice wx "feature"...

morinted:
It almost feels like "write once, then write two more times."

PR #375, opened 2016-02-20
Fix currency retroformatting exception

morinted:
When trying to currency format after an output that wasn't a number, there
was an exception and Plover's output would cease completely.

To work around this is just to define the case where the user's previous
word is not a number, at which point the command should do nothing.
Explicitly added tests for the 0 cases as well, because they are edge.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -511,26 +511,8 @@ def _atom_to_action_spaces_before(atom, last_action):
                 action.replace = last_action.text
                 action.text = _upper(last_action.text)
         elif meta.startswith(META_RETRO_FORMAT):
-            if (meta.startswith(META_RETRO_FORMAT) and meta.endswith(')')):
-                dict_format = meta[len(META_RETRO_FORMAT):-len(')')]
-                action = last_action.copy_state()
-                action.replace = last_action.word
-                try:
-                    float(last_action.word)
-                except ValueError:
-                    pass
-                else:
-                    format = dict_format.replace('c', '{:,.2f}')
-                    cast_input = float(last_action.word)
-                try:
-                    int(last_action.word)
-                except ValueError:
-                    pass
-                else:
-                    format = dict_format.replace('c', '{:,}')
-                    cast_input = int(last_action.word)
-                action.text = format.format(cast_input)
-                action.word = action.text
+            if meta.startswith(META_RETRO_FORMAT) and meta.endswith(')'):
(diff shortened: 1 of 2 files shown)


plover/formatting.py
```diff
@@ -698,23 +700,25 @@ def _atom_to_action_spaces_after(atom, last_action):
...
-                    cast_input = int(last_action.word)
-                action.text = format.format(cast_input) + SPACE
-                action.word = format.format(cast_input)
+                    currency_format = dict_format.replace('c', '{:,}')
+                    cast_input = integer_word
+                if cast_input is not False:
+                    action.replace = last_action.word + SPACE
+                    action.word = currency_format.format(cast_input)
+                    action.text = action.word + SPACE
         elif meta.startswith(META_COMMAND):
```
benoit-pierre: Same as above. How about factorizing the code a little? The only difference is the extra space:
```diff
                 if cast_input is not None:
-                    action.replace = last_action.word
+                    action.replace = last_action.word + SPACE
                     action.word = currency_format.format(cast_input)
-                    action.text = action.word
+                    action.text = action.word + SPACE
```

morinted:
@benoit-pierre how about now?

benoit-pierre:
I would just get rid of `word_suffix`:

```diff
diff --git i/plover/formatting.py w/plover/formatting.py
index 11936fc..0cdd04f 100644
--- i/plover/formatting.py
+++ w/plover/formatting.py
@@ -791,7 +791,6 @@ def _apply_mode(text, case, space_char, begin, last_attach,


 def _apply_currency(meta, last_action, spaces_after=False):
-    word_suffix = SPACE if spaces_after else ''
     dict_format = meta[len(META_RETRO_FORMAT):-len(')')]
     action = last_action.copy_state()
     cast_input = None
@@ -808,9 +807,12 @@ def _apply_currency(meta, last_action, spaces_after=False):
     else:
         currency_format = dict_format.replace('c', '{:,}')
     if cast_input is not None:
-        action.replace = last_action.word + word_suffix
+        action.replace = last_action.word
         action.word = currency_format.format(cast_input)
-        action.text = action.word + word_suffix
+        action.text = action.word
+        if spaces_after:
+            action.replace += SPACE
+            action.text += SPACE
     return action

 def _change_mode(command, action):
```

Or alternatively, keep it in replacement of the `spaces_after` argument.

PR #380, opened 2016-02-24
OSX: Send key codes instead of strings when possible

### LINKED ISSUE #246: OSX: Keyboard events send keycode zero
All keyboard events in OSX send keycode zero but they also set the string. Most apps honor the string but some do not. Those that do not will see all of our key strokes as the letter A (at least when the keyboard is set to be the US keyboard). Ideally, the other app would just honor the string but we can mitigate the situation by sending the correct keycode, when possible. 

There are a couple of options. We can just assume the keyboard is set to US English and choose those keycodes. A better option would be to find out what the current keyboard is and send the right codes for that keyboard. The problem with either solution is that this means the user cannot type things that don't exist on that keyboard. But maybe that's the best we can do.

morinted:
### About

Change `osxkeyboardcontrol.py` to use key codes based on the user's current layout, instead of just setting the string and sending key code 0. I imported `mac_layout.py`, a creation from @willwade where he used the base code in `PyKeycode` for the purpose of porting over functionality to `PyUserInput`. Because that code is not published, I had to copy and add attribution instead of some clever `import`. The new OSX `send_string` also tries to group modifier presses to avoid to avoid unnecessary key presses, as would be in the case of sending in all CAPS output.

### Issues

Fixes #246, which in turn has a couple of results for a typical Mac user. I've noticed the keycodes to be helpful in the following specific cases:

- Plover no longer crashes when it emulates keypresses onto itself
- Anki accepts input instead of just seeing 'a' for every key sent
- KDiff3 has the same behavior as Anki
- The typing website, 10fastfingers, now works with Plover on OSX
- Key-bindings with modifiers now work on the Atom-editor's Vim-mode plugin

### Release Notes

- Broaden supported applications on OS X (Anki, KDiff3, 10fastfingers.com)
- Bug fix: OS X commands are now layout-independent (before, they were hardcoded to QWERTY)

### Tested Platforms

I only tested on OS X, assuming that the code will not affect other platforms.

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 from Quartz import (
     CFMachPortCreateRunLoopSource,
     CFRunLoopAddSource,
@@ -34,32 +35,12 @@ from Quartz import (
 import Foundation
 import threading
 import collections
-import sys
+from plover.oslayer import mac_keycode
 
 
-# This mapping only works on keyboards using the ANSI standard layout. Each
-# entry represents a sequence of keystrokes that are needed to achieve the
-# given symbol. First, all keydown events are sent, in order, and then all
-# keyup events are send in reverse order.
 BACK_SPACE = 51
 
 KEYNAME_TO_KEYCODE = collections.defaultdict(list, {
-    # The order follows that of the plover guide.
-    # Keycodes from http://forums.macrumors.com/showthread.php?t=780577
-    '0': [29], '1': [18], '2': [19], '3': [20], '4': [21], '5': [23],
-    '6': [22], '7': [26], '8': [28], '9': [25],
-
-    'a': [0], 'b': [11], 'c': [8], 'd': [2], 'e': [14], 'f': [3], 'g': [5],
-    'h': [4], 'i': [34], 'j': [38], 'k': [40], 'l': [37], 'm': [46], 'n': [45],
-    'o': [31], 'p': [35], 'q': [12], 'r': [15], 's': [1], 't': [17], 'u': [32],
-    'v': [9], 'w': [13], 'x': [7], 'y': [16], 'z': [6],
-
-    'A': [56, 0], 'B': [56, 11], 'C': [56, 8], 'D': [56, 2], 'E': [56, 14],
-    'F': [56, 3], 'G': [56, 5], 'H': [56, 4], 'I': [56, 34], 'J': [56, 38],
-    'K': [56, 40], 'L': [56, 37], 'M': [56, 46], 'N': [56, 45], 'O': [56, 31],
-    'P': [56, 35], 'Q': [56, 12], 'R': [56, 15], 'S': [56, 1], 'T': [56, 17],
-    'U': [56, 32], 'V': [56, 9], 'W': [56, 13], 'X': [56, 7], 'Y': [56, 16],
-    'Z': [56, 6],
 
     'Alt_L': [58], 'Alt_R': [61], 'Control_L': [59], 'Control_R': [62],
     'Hyper_L': [], 'Hyper_R': [], 'Meta_L': [], 'Meta_R': [],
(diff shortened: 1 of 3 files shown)


morinted:
I was having issues with transpositioning (`gabrled` text output) while developing this, hence the addition of the timestamp modifications. I'm trying to get @stanographer to test it out today to see if the issue is resolved in a build with this code.

morinted:
I don't notice any difference in frequency with transposition errors, with or without the timestamp code. I've removed it and this should be good for review/merge.

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -371,9 +387,11 @@ def send_key_combination(self, combo_string):
             else:
                 current_command.append(c)
         # Record final command key.
-        keystring = ''.join(current_command)
-        seq = KEYNAME_TO_KEYCODE[keystring]
-        keycode_events.extend(down_up(seq))
+        if len(current_command) is not 0:
```
benoit-pierre: Should be changed to:
```python
if current_command:
```

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -393,10 +430,18 @@ def _send_sequence(self, sequence):
...
-                MY_EVENT_SOURCE, keycode, key_down)
-            CGEventSetFlags(event, event_mask)
+
+            event = CGEventCreateKeyboardEvent(MY_EVENT_SOURCE, keycode, key_down)
+
+            # The event comes with flags already, check if they contain our own event_mask.
+            if event_mask and (CGEventGetFlags(event) & event_mask) != event_mask:
+                # If our event_mask is missing from flags, set it.
+                CGEventSetFlags(event, event_mask)
+            elif event_mask is 0 and keycode not in MODIFIER_KEYS_TO_MASKS:
```
benoit-pierre: Why not:
```python
if not event_mask and keycode not in MODIFIER_KEYS_TO_MASKS:
```

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -300,37 +249,89 @@ def characters(s):
...
         """Emulate a sequence of key combinations.
 
-        Argument:
-
-        combo_string -- A string representing a sequence of key
+        Args:
+            combo_string: A string representing a sequence of key
         combinations. Keys are represented by their names in the
         Xlib.XK module, without the 'XK_' prefix. For example, the
         left Alt key is represented by 'Alt_L'. Keys are either
```
benoit-pierre: Should this be indented the same as the first paragraph line?

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -222,6 +167,7 @@ def down_up(seq):
     36 : "Return", 124: "Right", 48: "Tab", 126: "Up",
 }
```
benoit-pierre: I just noticed (well pylint did :P) that both BackSpace and Tab have the same keycode.

morinted: Ah, how funny. Will have to modify BS to 51.

plover/oslayer/osxkeyboardcontrol.py
morinted: I Googled around and I think the correct format would be

```
Args:
    name: asetasetasetaset aset aset aseta aaset
        asetasetasetasetase. aasetaset.
```

morinted:
@benoit-pierre I applied the minor changes you suggested, and I tried out the BackSpace capture in the Keyboard config -- all looks good.

plover/oslayer/osxkeyboardcontrol.py
benoit-pierre: If think the convention we use elsewhere is right:

```python
    def send_key_combination(self, combo_string):
        """Emulate a sequence of key combinations.

        Arguments:

        combo_string -- A string representing a sequence of key combinations.
        Keys are represented by their names in the Xlib.XK module, without the
        'XK_' prefix. For example, the left Alt key is represented by 'Alt_L'.
        Keys are either separated by a space or a left or right parenthesis.
        Parentheses must be properly formed in pairs and may be nested. A key
        immediately followed by a parenthetical indicates that the key is
        pressed down while all keys enclosed in the parenthetical are pressed
        and released in turn. For example, Alt_L(Tab) means to hold the left
        Alt key down, press and release the Tab key, and then release the left
        Alt key.
        """
```

See [PEP-0257](https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation).

morinted:
@benoit-pierre is this okay to merge now?

benoit-pierre:
:+1:

PR #381, opened 2016-02-24
config: allow first stroke to be capitalized and attached

morinted:
### About

Add output configuration options so that the first stroke after Plover is enabled can be capitalized or attached (no space). The default behavior is the same as Plover 2.5.8, and this shows up as checkboxes in the output section of the configuration.

Configuration screenshot

### Reason

This never really bothered me until @stanographer asked for it. It seems like a sensible feature, because some users (like me) always want no space and capitals, some users might want capitals *with* a space (writing text, for example), and some users may not want this change at all.

### Release Notes

Configure whether Plover's first stroke after being turned on is capitalized and has a space.

### Tested Platforms

Only tried on OS X, but also included test cases to verify the functionality.

PR #382, opened 2016-02-24
Add stroke buffer configuration, default to 100

### LINKED ISSUE #300: History size (number of undo levels) should be configurable
- [ ] add a configuration entry
- [ ] add a way to change said configuration entry to the GUI

morinted:
Add buffer configuration, default 100. UI arbitrarily maxes out at 100,000, minimum is set to 1.

Fix #300 

screen shot 2016-02-29 at 4 41 32 pm

### DIFF
--- a/plover/gui/suggestions.py
+++ b/plover/gui/suggestions.py
@@ -16,7 +16,6 @@ ON_TOP_TEXT = "Always on top"
 UI_BORDER = 4
 LAST_WORD_TEXT = 'Last Word: %s'
 DEFAULT_LAST_WORD = 'N/A'
-HISTORY_SIZE = 10
 MAX_DISPLAY_LINES = 20
 STROKE_INDENT = 2
 
@@ -161,7 +160,8 @@ class SuggestionsDisplayDialog(wx.Dialog):
     def show_suggestions(self, suggestion_list):
 
         # Limit history.
-        if len(self.history) == HISTORY_SIZE:
+        undo_levels = self.config.get_undo_levels()
+        while len(self.history) >= undo_levels:
             _, text_length = self.history.pop(0)
             last_position = self.listbox.GetLastPosition()
             self.listbox.Remove(last_position - text_length, last_position)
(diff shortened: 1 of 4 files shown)


Achim63:
Very welcome as well. Plover even misbehaved majorly when I reached the limit - it started to delete large chunks of text unexpectedly in that case, instead of just stopping to delete anything.

benoit-pierre:
We should take this opportunity to [add a configuration value](https://github.com/benoit-pierre/plover/commit/11b40d64489a7a78d6436212f05ee6d15da5c31d), and partially fix #300. Bonus points if someone add this to the GUI. Not it!

morinted:
All right, all right, I'll do it and amend this commit. By Saturday, okay?

benoit-pierre:
Sure, no hurry!

morinted:
@benoit-pierre UI ui-ified. Probably means reworking my #381 UI, but I'm okay with that.

plover/gui/suggestions.py
```diff
@@ -161,7 +161,8 @@ def lookup_suggestions(self, phrase):
     def show_suggestions(self, suggestion_list):
 
         # Limit history.
-        if len(self.history) == HISTORY_SIZE:
+        undo_levels = self.config.get_undo_levels()
+        while len(self.history) >= undo_levels:
```
benoit-pierre: This code does not work if undo levels is 0, fix [here](https://github.com/morinted/plover/compare/buffer...benoit-pierre:buffer).

morinted: Maybe I'll just limit the minimum to 1 because that's what the shortest dictionary would set it to :)

morinted:
Threw in the assertion, changed the minimum undo length to 1.

morinted:
@benoit-pierre Using the exception caused problems because Plover wouldn't finish initializing, I've just opted for a warning notification instead. How's that?

morinted:
Updated and squashed, right now an invalid configuration (< 1) causes Plover to crash on startup.

PR #384, opened 2016-02-26
translation: fix undo strokes on empty state

benoit-pierre:
Don't add those strokes to the state history (since undo strokes are not supposed to be undo-able).


benoit-pierre:
And I need to update the tests...

PR #385, opened 2016-02-27
Update contributors in about dialog

### LINKED ISSUE #372: Update author attributions in About dialogue
For the credits in the About dialogue, we have:

`Stenographer: Mirabai Knight, Coder: Joshua Harlan Lifton, Contributors: Hesky Fisher`

Maybe instead, we can mention Mirabai as the lead stenographer/founder, and then link to the GitHub contributors page, as it will be more detailed than we could ever be in the about dialogue: https://github.com/openstenoproject/plover/graphs/contributors

morinted:
### About

Update contributors dialog with a link to GitHub to reflect the many developers that we are lucky to have. Also reword to better highlight @stenoknight as the founder of the project.

- Include a non-clickable link to the GitHub repo
- Reformat to ensure that the about window doesn't resize when expanding and collapsing sections.

screen shot 2016-02-21 at 1 11 25 am

### Issues

Fixes #372

### Tested Platforms

Only tested OS X, I believe this window type is supposed to be cross platform (GTK-style about dialog)

### DIFF
--- a/plover/__init__.py
+++ b/plover/__init__.py
@@ -7,11 +7,16 @@ __version__ = '2.5.8'
 __copyright__ = '(C) Open Steno Project'
 __url__ = 'http://www.openstenoproject.org/'
 __download_url__ = 'http://www.openstenoproject.org/plover'
-__credits__ = ['Stenographer: Mirabai Knight',
-               'Coder: Joshua Harlan Lifton',
-               'Contributors: Hesky Fisher']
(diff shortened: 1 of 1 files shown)


morinted:
I tried it in Ubuntu, and it mostly worked nicely. The re-spacing of the about paragraph looked a little off, but the contributors list was fine. The other oddity was the link was clickable, but the `<link>` syntax triggered my mail client. I'm not too concerned about that, either, as it's still readable and you cannot click at all on the other platforms.

Windows also works exactly like OSX, so no worry there.

PR #386, opened 2016-02-27
Update Plover icon

morinted:
### About

Update Plover icon with the one we made in the art thread. The icon is in many sizes for OS X in the `icns` file (up to 1024), and in the `.ico` file for Windows (I believe 16x16 to 256x256), and the `png` is independent and to be used by Linux. I couldn't find too much info, other than that different DMs handle the icon sizes differently. Most of them seem to be able to downscale a high-resolution icon (128x128).

Also add icon to the titlebar on Windows with wx SetIcon.

#### Screenshots

screen shot 2016-02-25 at 7 31 58 pm

Here's the [.ai and .svg](https://github.com/openstenoproject/plover/files/149154/Plover-vectors.zip) for future reference and modifications.

### Issues

See #263 

### Release Notes

Remove placeholder green/red P, replacing with our very own Plover icon: a steno key with an asterisk.

### Tested Platforms

I've worked on OS X and Windows, needs testing on Linux.


morinted:
Here's a selection to show what it looks like on Windows 7:

Windows screenshots

morinted:
All right, I tried this on Ubuntu and all I needed to do to get it look nice in Unity was overwrite the default backlighting color for the launcher.

benoit-pierre:
I think at least the .svg should be versioned too.

morinted:
Where best to place it? It doesn't need to be included in the distributed
assets. I was thinking of a top-level art/ ?

morinted:
Ah, never mind my ramblings, the .svg is of course tiny. I've included it in the assets folder and amended.

windows/pyinstaller.spec
```diff
@@ -15,4 +15,4 @@ exe = EXE(pyz,
           strip=None,
           upx=True,
           console=False,
-          icon='plover.ico')
+          icon=os.path.join('plover/assets', 'plover.ico'))
```
benoit-pierre: `windows/plover.ico` should be removed, and `windows/helper.py` updated instead:
```diff
diff --git i/windows/helper.py w/windows/helper.py
index f7b1eb2..a68ae21 100755
--- i/windows/helper.py
+++ w/windows/helper.py
@@ -32,7 +32,7 @@ from plover import (

 APPNAME = __software_name__.capitalize()
 VERSION = __version__
-ICON = os.path.join(WIN_DIR, '%s.ico' % __software_name__)
+ICON = os.path.join(TOP_DIR, __software_name__, 'assets', '%s.ico' % __software_name__)


 if sys.stdout.isatty():
```

I'll remove `windows/pyinstaller.spec` in a subsequent [pull request](https://github.com/openstenoproject/plover/compare/openstenoproject:12b554d...benoit-pierre:ce3335f).

morinted:
Okay, just updated helper.py, the pyinstaller.spec change doesn't matter either way, may as well keep it accurate anyways until you delete it. Are we good to merge this now?

benoit-pierre:
:+1:

PR #388, opened 2016-02-29
Windows related tweaks for the build environment

benoit-pierre:
- update build instructions and batch file
- update setup instructions
- add batch file to setup environment


windows/README.md
```diff
@@ -1,31 +1,38 @@
...
+- `pip install -r requirements.txt`
+
+### Running Plover in development
 
 To run from source, from the root of the Git repository, use `python launch.py`.
 
 ### Building
 
-As a prerequesite, ensure that there is a file named `__init__.py` in the pywinusb directory in site-packages. If not, create it (as an empty file)
+As a pre-requisite, ensure that there is a file named `__init__.py` in the pywinusb directory in site-packages. If not, create it (as an empty file).
```
benoit-pierre: Is it still required? @morinted: does your Treal work with the AppVeyor builds?

morinted: As far as I'm aware, pip-installed pywinusb never had that problem.

benoit-pierre: Okay, so is there any point in keeping it?

morinted: No, we should remove that detail.

benoit-pierre: OK, removed.

PR #393, opened 2016-03-01
More robust init

### LINKED ISSUE #273: Failure to initialize machine crashes Plover
Plover tries to initialize the steno machine on startup. If that throws an exception the Plover crashes with a stack trace. Usually, the solution is to delete plover.cfg and start again. On OSX, this happens on first startup because the machine is the keyboard, by default, and that requires special security permissions to be set.

It would be better if:
1) Plover caught exceptions from initializing the machine and
2) support running with no machine configured 

so that the program doesn't just crash when an invalid machine is specified but just switches to no machine and the user can go change the machine settings or use the dictionary editor or whatever else they want to do.

benoit-pierre:
Fix #273.

* use blanket except clauses so Plover does not crash if something goes wrong during the init.
* use the logger for reporting exceptions.
* use a custom formatter for GUI log handlers: don't show traceback on exception, since the messages are usually limited in length, and the information is not really useful to most users (the full traceback is still logged to the console and the logfile).

### DIFF
--- a/plover/app.py
+++ b/plover/app.py
@@ -62,16 +62,16 @@ def update_engine(engine, old, new, reset_machine=False):
         old = NoneConfig()
 
     machine_type = new.get_machine_type()
-    machine_mappings = new.get_system_keymap(machine_type)
-    machine_options = new.get_machine_specific_options(machine_type)
+    try:
+        machine_mappings = new.get_system_keymap(machine_type)
+        machine_options = new.get_machine_specific_options(machine_type)
+    except NoSuchMachineException as e:
+        raise InvalidConfigurationError(unicode(e))
     if (reset_machine or
             old.get_machine_type() != machine_type or
             old.get_system_keymap(machine_type) != machine_mappings or
             old.get_machine_specific_options(machine_type) != machine_options):
-        try:
-            machine_class = machine_registry.get(machine_type)
-        except NoSuchMachineException as e:
-            raise InvalidConfigurationError(unicode(e))
+        machine_class = machine_registry.get(machine_type)
         machine = machine_class(machine_options)
         if machine_mappings is None:
             log.warning('no mappings defined for %s, the machine won\'t be usable', machine_type)
--- a/plover/app.py
+++ b/plover/app.py
@@ -40,8 +40,8 @@ class SimpleNamespace(object):
 
 def init_engine(engine, config):
     """Initialize a StenoEngine from a config object."""
-    update_engine(engine, config)
     engine.set_is_running(config.get_auto_start())
+    update_engine(engine, config)
 
 def reset_machine(engine, config):
     """Set the machine on the engine based on config."""
--- a/plover/app.py
+++ b/plover/app.py
@@ -40,75 +40,50 @@ class SimpleNamespace(object):
 
 def init_engine(engine, config):
     """Initialize a StenoEngine from a config object."""
-    update_engine(engine, None, config)
+    update_engine(engine, config)
     engine.set_is_running(config.get_auto_start())
 
 def reset_machine(engine, config):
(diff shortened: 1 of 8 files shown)


morinted:
- Plover doesn't complain when opening multiple times now.
- When I open without assistive devices enabled, I'm presented with the error and the configuration dialogue shows up. So then I select "Tréal" from the machine list, and hit "save". Then, the user interface shows "Gemini PR" while the console is talking about the Tréal, so I try to close Plover, and then the user interface turns into a shell and no longer responds
screen shot 2016-03-01 at 3 58 38 pm

Log is mostly empty:

```
2016-03-01 15:56:41,632 ERROR: updating engine configuration failed
   […]
    raise Exception("Enable access for assistive devices.")
Exception: Enable access for assistive devices.
2016-03-01 15:56:56,463 INFO: Treal device not found: open failed
2016-03-01 15:56:56,464 WARNING: Treal is not connected
2016-03-01 15:56:56,478 INFO: Treal device not found: open failed
2016-03-01 15:56:56,478 WARNING: Treal is not connected
2016-03-01 15:56:56,481 INFO: set_stroke_filename(/Users/ted/Library/Application Support/plover/strokes.log)
2016-03-01 15:56:56,481 INFO: enable_stroke_logging(True)
2016-03-01 15:56:56,481 INFO: enable_translation_logging(True)
```

benoit-pierre:
> Plover doesn't complain when opening multiple times now.

I can't reproduce that (both on Linux and with a Windows 10 VM).

morinted:
Ah, yeah, seems that I was mistaken on that one.

benoit-pierre:
> When I open without assistive devices enabled, I'm presented with the error and the configuration dialogue shows up. So then I select "Tréal" from the machine list, and hit "save". Then, the user interface shows "Gemini PR" while the console is talking about the Tréal, so I try to close Plover, and then the user interface turns into a shell and no longer responds 

I tried simulating that by adding a call to `raise Exception("Enable access for assistive devices.")` in `oslayer/xkeyboardcontrol.py`, and switching to "Tréal" when the configuration dialog shows up and it works.

benoit-pierre:
You're log also does not match the errors you're describing: 
```
2016-03-01 15:56:56,478 INFO: Treal device not found: open failed
2016-03-01 15:56:56,478 WARNING: Treal is not connected
```

morinted:
Well, in that case, the Treal wasn't connected. The log was complaining about the Treal, but the UI is showing Gemini PR.

morinted:
Here's a little more detail.

- Open Plover: engine initialization failed:
screen shot 2016-03-01 at 4 08 54 pm
- Switch to Treal as machine, hit save...
screen shot 2016-03-01 at 4 09 07 pm
- Resulting state, pressing keys on the Treal has no effect:
screen shot 2016-03-01 at 4 09 17 pm

benoit-pierre:
If you start with the config set to `Machine1`, and there's an error during init, the configuration dialog is show, and if you then switch to `Machine2`, but the configuration update fails too (which can happen when retrieving the machine options, not just for the machine init), then the interface still shows `Machine1`. Is that what is happening?

morinted:
I'm starting with Keyboard, switching to Tréal, and then it's giving me this state. If I force close Plover and restart it, the Tréal is the selected device.

benoit-pierre:
OK, so on OSX with WX 3.0, `wx.CallAfter` cease to work after showing the modal configuration dialog in the GUI init loop... I removed said loop and fixed the logic in `update_engine`.

morinted:
I ran through this on Windows and OSX… It is indeed much more stable. I can break the dictionary configuration, break the machine, it all resumes correctly. It's very satisfying

PR #395, opened 2016-03-03
Add "carry capitalize" meta

### LINKED ISSUE #394: Entries that start with an apostrophe should be capitalized if it's after a period or a capitalization command
## Summary

Dictionary entries that start with an apostrophe should be capitalized if they're used at the beginning of a sentence, after a punctuation that necessitates a capitalized letter immediately after, or after a capitalization flag.

So if the entry is ['cause], then if I wrote the sentence, "So we should go with option B. 'Cause that was..."
the C in 'cause should be capitalized.

Accordingly, if there's a speaker marker:

MIRABAI: 'Cause that's...

same thing.

## Steps to Reproduce

Write 'cause after anything that would normally immediately be followed by a initially-capitalized word.

## Version

Plover version 2.5.8 with tweaked settings for reduced transposition errors.

Installed via: Ted

## Configuration

OSX 10.11.3

morinted:
### About

Add a 'carry capitalize' meta command that allows users to output text without affecting the capitalization (which the attach flag does).

Allows users to define quotes that don't affect capitalization, such as opening quotes at the start of a sentence or closing quotes after a full stop.

Also adds a valid workaround for `'cause` to be able to output as `'Cause' with `{~|'^}cause`

I updated the dictionary to include strokes that take advantage of this command.

### Issues

Fixes #394 

### Release Notes

Add "carry capitlalize" command, to allow you to define strokes that don't reset whether or not the next word is capitalized, great for puntuation that shouldn't affect capitalization (like quotes, parentheses, and apostrophes). The default dictionary has been updated to better handle words starting with apostrophes, like `'cause` or `'em`. The default quotation mark and parentheses strokes (`KW-GS`, `KR-GS`; `PREPB` and `PR*EPB`) also should no longer force lowercase, making them better for use in your text.

Use the new command with the syntax `~|`. You invoke the command as others, in braces, and you can optionally specify whether you want to suppress spaces on either side with the familiar `^` symbol. For example, you could begin a word with an apostrophe: `{~|'^}cause`. Here is a set of opening and closing quotations, where the opening quotations attach to the next word, and the closing quotations attach to the previous word: `{~|"^}` and `{^~|"}`.


plover/formatting.py
```diff
@@ -513,6 +514,9 @@ def _atom_to_action_spaces_before(atom, last_action):
             else:
                 action.replace = last_action.text
                 action.text = _upper(last_action.text)
+        elif (meta.startswith(META_CARRY_CAPITALIZATION) or
+              meta.startswith(META_ATTACH_FLAG + META_CARRY_CAPITALIZATION)):
+            action = _apply_carry_capitalize(meta, last_action, spaces_after=False)
```
benoit-pierre: This seems weird: `spaces_after` is not a keyword argument. I would convert it to a keyword argument (default=`False`), and omit it here, for consistency with the way it's handled elsewhere (e.g. `_apply_currency`).

plover/formatting.py
```diff
@@ -819,6 +826,33 @@ def _apply_currency(meta, last_action, spaces_after=False):
     return action
 
 
+def _apply_carry_capitalize(meta, last_action, spaces_after):
```
benoit-pierre: See note above.

plover/formatting.py
```diff
@@ -819,6 +826,33 @@ def _apply_currency(meta, last_action, spaces_after=False):
...
+    # Meta format: ^~|content^ (attach flags are optional)
+    attach_last = meta.startswith(META_ATTACH_FLAG)
+    attach_next = meta.endswith(META_ATTACH_FLAG)
+
+    content_start = meta.index(META_CARRY_CAPITALIZATION) + len(META_CARRY_CAPITALIZATION)
+    content_end = -(len(META_ATTACH_FLAG)) if attach_next else None
```
benoit-pierre: No need for the extra pair of parenthesis around `len(...)`.

morinted:
Updated with your feedback! Thanks, as always.

PR #397, opened 2016-03-04
Add preliminary system support

benoit-pierre:
Move all system/theory related parameters to a dedicated module:
* steno keys, number keys, suffix keys
* undo stroke
* orthography word list and rules
* default keymaps for supported machines

Note:
* the keyboard keymap configuration option was moved out of the `[Keyboard]` section to a new `[System: English Stenotype]` section
* although not supported in the GUI, it was simpler and cleaner to use a generic approach, so it possible to change a machine (other than the keyboard) keymap through the configuration, e.g.:
```python
[System: English Stenotype]
keymap[treal] = [["#", ["#1", "#2", "#3", "#4", "#5", "#6", "#7", "#8", "#9", "#A", "#B", "X3"]], ["S-", ["S1-", "S2-"]], ["T-", ["T-"]], ["K-", ["K-"]], ["P-", ["P-"]], ["W-", ["W-"]], ["H-", ["H-"]], ["R-", ["R-"]], ["A-", ["A-"]], ["O-", ["O-"]], ["*", ["*1", "*2"]], ["-E", ["-E"]], ["-U", ["-U"]], ["-F", ["-F"]], ["-R", ["-R"]], ["-P", ["-P"]], ["-B", ["-B"]], ["-L", ["-L"]], ["-G", ["-G"]], ["-T", ["-T"]], ["-S", ["-S"]], ["-D", ["-D"]], ["-Z", ["-Z"]], ["no-op", ["X1-", "X2-"]]]
```

### DIFF
--- a/plover/app.py
+++ b/plover/app.py
@@ -26,6 +26,7 @@ from plover.exception import InvalidConfigurationError,DictionaryLoaderException
 from plover.machine.registry import machine_registry, NoSuchMachineException
 from plover import log
 from plover.dictionary.loading_manager import manager as dict_manager
+from plover import system
 
 # Because 2.7 doesn't have this yet.
 class SimpleNamespace(object):
@@ -61,15 +62,22 @@ def update_engine(engine, old, new, reset_machine=False):
         old = NoneConfig()
 
     machine_type = new.get_machine_type()
+    machine_mappings = new.get_system_keymap(machine_type)
     machine_options = new.get_machine_specific_options(machine_type)
     if (reset_machine or
             old.get_machine_type() != machine_type or
+            old.get_system_keymap(machine_type) != machine_mappings or
             old.get_machine_specific_options(machine_type) != machine_options):
         try:
             machine_class = machine_registry.get(machine_type)
         except NoSuchMachineException as e:
             raise InvalidConfigurationError(unicode(e))
-        engine.set_machine(machine_class(machine_options))
+        machine = machine_class(machine_options)
+        if machine_mappings is None:
+            log.warning('no mappings defined for %s, the machine won\'t be usable', machine_type)
+        else:
+            machine.set_mappings(machine_mappings)
+        engine.set_machine(machine)
 
     dictionary_file_names = new.get_dictionary_file_names()
     if old.get_dictionary_file_names() != dictionary_file_names:
--- a/plover/config.py
+++ b/plover/config.py
@@ -10,6 +10,7 @@ from cStringIO import StringIO
 from plover.exception import InvalidConfigurationError
 from plover.machine.registry import machine_registry
 from plover.oslayer.config import ASSETS_DIR, CONFIG_DIR
+from plover import system
 
 SPINNER_FILE = os.path.join(ASSETS_DIR, 'spinner.gif')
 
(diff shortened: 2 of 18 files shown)


morinted:
I look forward to reviewing this tomorrow. It's the main block before a release. I'll also be debugging the wx init on Mac tomorrow morning.

morinted:
Okay, everything seems pretty good, but I guess the real test will be people trying it on their machines in the real world.

I've noticed what may be some unexpected user interface traits that I'd like to get your feedback on (on OSX, of course):

- First, the default column width cuts the "Steno Keys / Actions" label, and the text seems to be touching the left side of the list box.

screen shot 2016-03-04 at 9 10 14 am

- The clear button is offset, and is there a reason we don't have a reset to default for a single key?

screen shot 2016-03-04 at 9 10 29 am

benoit-pierre:
Regarding the absence of "reset' button when editing a single mapping: I don't think it make sense to reset a single mapping to default if the layout is different: e.g. if it has been moved to the Fn+number keys bar, resetting the "S-" mapping to "Q/A" does not make sense by itself. And if the user is starting from the default keymap, then it's always possible for him to abort the change with "cancel", and start the editing over.

morinted:
Okay, I'm all right with that decision. The UI "glitches" aren't major, either.

benoit-pierre:
OK, I amended the TX Bolt layout, and tweaked the keyboard config GUI:

* fix "clear" mapping button alignment
* harmonize borders
* enable horizontal and vertical rules for the mappings list

Please check on OSX.

benoit-pierre:
I don't know why `plover/oslayer/keyboardcontrol.py` appears as completely changed, EOLs? I'm going to see if I can fix this.

benoit-pierre:
I fixed the issue with `plover/oslayer/keyboardcontrol.py` (It was a change in EOL format from DOS to Unix).

morinted:
The clear button looks great, the table has an odd alignment issue (I'll toy around with it for a few minutes...)

screen shot 2016-03-04 at 12 39 24 pm

morinted:
Okay, well at the very least I preferred it before. It's not *that* bad so we don't need to try and "fix" it.

What do you think of this patch?

```diff
diff --git a/plover/gui/keyboard_config.py b/plover/gui/keyboard_config.py
index 889536d..2eb58b4 100644
--- a/plover/gui/keyboard_config.py
+++ b/plover/gui/keyboard_config.py
@@ -86,11 +86,11 @@ class EditKeymapWidget(wx.ListCtrl, listmix.ListCtrlAutoWidthMixin):
                              id=wx.ID_ANY,
                              pos=wx.DefaultPosition,
                              size=(300, 200),
-                             style=wx.LC_REPORT|wx.LC_HRULES|wx.LC_VRULES)
+                             style=wx.LC_REPORT)
         listmix.ListCtrlAutoWidthMixin.__init__(self)
         self.Bind(wx.EVT_LIST_ITEM_ACTIVATED, self.edit_item)
-        self.InsertColumn(0, 'Steno Keys / Actions')
-        self.InsertColumn(1, 'Keys')
+        self.InsertColumn(0, 'Steno Keys / Actions', width=wx.LIST_AUTOSIZE_USEHEADER)
+        self.InsertColumn(1, 'Keys', width=wx.LIST_AUTOSIZE_USEHEADER)

     def edit_item(self, event):
         # Disallow editing of first column.
```

benoit-pierre:
Not a fan of the rules? I think this make it more readable. I would have enabled [alternate row colors](http://docs.wxwidgets.org/stable/classwx_list_ctrl.html#a274e98c1c5101725bbdca9b19d8678a4), but it's not available with the version of wxWidgets we're using.

morinted:
I don't like the look in the screenshot above, maybe H rules only?

morinted:
Ah, I should add that rulers aren't common on OSX lists of any kind, so it might just be a platform thing.

benoit-pierre:
IMHO, usability > looks. I'm fine with the `width=wx.LIST_AUTOSIZE_USEHEADER` change.

morinted:
Sure, let's add the autosize to header then merge :)

benoit-pierre:
Done.

PR #403, opened 2016-03-06
setup: add missing 'system' module to packages

benoit-pierre:
Unfortunately, our egg and wheel files for the first weekly don't work because of this missing module...


morinted:
Okay, this will be a good opportunity to create another "weekly". Are we still prepopulating the release notes when creating our own tag? May be good to make the install instructions static so that we aren't copying and pasting those all the time.

PR #409, opened 2016-03-07
Fix retro delete space to work with meta actions and numbers

### LINKED ISSUE #402: Retrospective delete space ( {*!} ) is not working for numbers
Classification: Bug

Reproducibility: Always

## Summary

Retrospective delete space ( {*!} ) is not working for numbers.

## Steps to Reproduce

1. input a word like: user
2. enter numbers like: 123
3. input key combination for Retrospective delete space

## Expected Results

4. user123

## Actual Results

5. Instead, this happens: user 123

## Version

Plover version 2.5.8.317.gced379f.

Installed via: download from https://github.com/openstenoproject/plover/releases/tag/weekly-v137

## Notes

Confirmed that the Retrospective delete space works as expected with other words like: username (spelled: AOURS/TPHAEUPL/TK\*LS where TK\*LS is the stoke for Retrospective delete space)

## Configuration

OS: Windows 8.1

### LINKED ISSUE #405: Retrospective delete space ( {*!} ) is not working after period
Classification: Bug

Reproducibility: Always

## Summary

Retrospective delete space ( {*!} ) is not working after period.

## Steps to Reproduce

1. Enter text after a regular period (-P)
2. Try to do a Retrospective delete space

## Expected Results

3. Space after period should be deleted

## Actual Results

3. No change

## Version

Plover version Plover version 2.5.8.317.gced379f

Installed via: download from https://github.com/openstenoproject/plover/releases/tag/weekly-v137

## Notes

Not sure if this behavior is expected, but it would be useful for correcting typos when looking for dotfiles on the console. Example: . zshrc -> .zshrc   (-P/S*EU/S*/H*/R*/KR*/TK*LS [where TK*LS is the stoke for Retrospective delete space])

workaround: use an entry on your dict for {^.^} to suppress the space after the period.

## Configuration

OS:

OS: Windows 8.1

morinted:
### About

Retro-del-space would only join the previous two translations, which didn't work for cases like "{.}" because the output would be "{.}word" which equates to ". Word". Instead, now we join with `{^~|^}` which will remove the space and maintain capitalization (".Word" in this example).

Also handle numbers. Previously we only used the translation's english value, but if the stroke is a number it has no english value.

### Issues

Fixes #402, #405

### Release Notes

Fix retroactive delete space with punctuation, numbers, and other meta commands.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -838,6 +838,10 @@ def _apply_carry_capitalize(meta, last_action, spaces_after=False):
     action = last_action.copy_state()
     action.attach = attach_next
 
+    # Spaces after: delete last space if we're attaching.
+    replace_last = last_action.text.endswith(SPACE) and attach_last
+    action.replace = SPACE if replace_last else NO_SPACE
+
     if meta_content:
         action.word = meta_content
 
@@ -846,10 +850,6 @@ def _apply_carry_capitalize(meta, last_action, spaces_after=False):
         # Only suffix a space if spaces are after or there's an attach_next flag.
         suffix = NO_SPACE if attach_next or not spaces_after else SPACE
         action.text = prefix + meta_content + suffix
-
-        # Spaces after: delete last space if we're attaching.
-        replace_last = last_action.text.endswith(SPACE) and attach_last
-        action.replace = SPACE if replace_last else NO_SPACE
 
     return action
 
--- a/plover/test_formatting.py
+++ b/plover/test_formatting.py
@@ -539,6 +539,11 @@ class FormatterTestCase(unittest.TestCase):
           action(text='EQUIPPED ', word='EQUIPPED', replace='equipped ', upper_carry=True),
          ]),
 
+        (('{.}{^~|^}zshrc', action(), True),
+         [action(text='. ', capitalize=True),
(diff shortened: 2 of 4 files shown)


david-pitty:
this same issue happens after or before a finger spell, like O*/WORD/TK*LS (useful for vim). Let me know if I should issue a separate bug for this, or if this was fixed along with the period issue.

morinted:
I believe it was fixed as well, this was a general fix that really just chained the core way that we were "attaching" translations. Please try it out with the AppVeyor artifact: https://ci.appveyor.com/project/morinted/plover/build/141-qovtrtnv/artifacts

morinted:
@david-pitty would you try the linked executable and see if it fixes the problem for you?

david-pitty:
Sorry for the delay in response. Yes, the issue is fixed in that binary.
 By the way, I did notice that if I first decapitalize with "{*>}" and then try retro removing the space it doesn't work, but it does work in the inverse order. Let me know if I should submit a separate bug for this.

morinted:
I believe that the space removal only handles literally the last two "strokes", so since there was no space between your second stroke and the third (decapitalize), you'll see no change. I believe that's to be expected.

david-pitty:
ah, true, that makes sense.

PR #412, opened 2016-03-09
osxkeyboardcontrol: explicity remove unwanted mask flags

morinted:
### About

Explicitly remove mask flags that we *don't* want, if they are not properly set. This extends the already present functionality. The results is that complex strings like `>µ>µ>µ>µ` work, where before they would come out as `≥Â˘Â˘Â˘Â˘µ>` or similar.

The change to HIDSystemEvent also fixes this case in KDiff3 and other legacy applications.

Also add a delay to output when sending key sequences, may fix the transposition issue.

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -35,6 +35,7 @@ from Quartz import (
 )
 import Foundation
 import threading
+from time import time
 import collections
 from plover.oslayer import mac_keycode
 
@@ -441,6 +442,11 @@ class KeyboardEmulation(object):
                 goal_flags = (event_flags & ~KeyboardEmulation.MODS_MASK) | mods_flags
                 if event_flags != goal_flags:
                     CGEventSetFlags(event, goal_flags)
+
+                # Half millisecond pause after keydown
+                deadline = time() + 0.0005
+                while time() < deadline:
+                    pass
 
             CGEventPost(kCGSessionEventTap, event)
 
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -31,6 +31,7 @@ from Quartz import (
     kCGHeadInsertEventTap,
     kCGKeyboardEventKeycode,
     kCGSessionEventTap,
+    kCGEventSourceStateHIDSystemState,
 )
 import Foundation
 import threading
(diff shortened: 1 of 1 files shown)


morinted:
Little patch to reduce the number of times that I'm checking the flags:

```diff
diff --git a/plover/oslayer/osxkeyboardcontrol.py b/plover/oslayer/osxkeyboardcontrol.py
index bbd7bb6..4685234 100644
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -433,6 +433,7 @@ class KeyboardEmulation(object):
             kCGEventFlagMaskShift |
             kCGEventFlagMaskCommand
         )
+
         for keycode, key_down in sequence:
             if not key_down and keycode in MODIFIER_KEYS_TO_MASKS:
                 event_mask &= ~MODIFIER_KEYS_TO_MASKS[keycode]
@@ -440,11 +441,12 @@ class KeyboardEmulation(object):

             event = CGEventCreateKeyboardEvent(MY_EVENT_SOURCE, keycode, key_down)

-            event_flags = CGEventGetFlags(event)
-            # Add wanted flags, remove unwanted flags.
-            goal_flags = (event_flags & ~inverted_event_mask) | (~event_flags & event_mask)
-            if event_flags != goal_flags and keycode not in MODIFIER_KEYS_TO_MASKS:
-                CGEventSetFlags(event, goal_flags)
+            if (key_down and keycode not in MODIFIER_KEYS_TO_MASKS):
+                event_flags = CGEventGetFlags(event)
+                # Add wanted flags, remove unwanted flags.
+                goal_flags = (event_flags & ~inverted_event_mask) | (~event_flags & event_mask)
+                if event_flags != goal_flags:
+                    CGEventSetFlags(event, goal_flags)

             CGEventPost(kCGSessionEventTap, event)
```

morinted:
Here's a build of this PR right now: https://drive.google.com/file/d/0B_4v0_ER8UlwaklPc01SNDVpQnc/view?usp=sharing

tkpwaeub:
I've gotten two scrambles since I installed it twenty minutes ago. So...a little better.

morinted:
The half-millisecond "sleep" has been pretty reliable for me. There's no noticeable change in Plover's ouput speeds (at 5ms, there is), and I haven't gotten any transpotion errors. I also don't get the slowdown that I would with `time.sleep(0.005)`.

tkpwaeub:
me, too, I haven't had any problems whatsoever.

morinted:
@benoit-pierre I've been using this and it's been very reliable, Stan also hasn't had any issues. Could we merge this and make another release as a test-bed?

PR #414, opened 2016-03-09
add Linux specific readme

benoit-pierre:
Add Linux specific readme with instructions on how to install dependencies and install and run Plover.


benoit-pierre:
Note: this should fix #316.

PR #419, opened 2016-03-12
travis: deploy draft instead of release

morinted:


benoit-pierre:
I was [testing this](https://github.com/openstenoproject/plover/compare/travis-draft...benoit-pierre:weekly), and for some reason the OSX build will draft another release instead of using the existing one, so if the builds finish/deploy in this order:

* AppVeyor
* Travis Linux
* Travis OSX

Then I get 2 draft releases, one with the Windows and Linux artifacts, and another with the result of the OSX build.

morinted:
Okay, if we don't publish a draft, then we're just public and in-progress for what, 15 to 20 minutes? Is it that bad?

benoit-pierre:
I still think it's better, even if we have to download and re-upload some of the artifacts to the right release, and of course do a few cursory tests before publishing the release.

PR #420, opened 2016-03-12
winkeyboardcontrol: verify active layout on output

### LINKED ISSUE #415: Switching keyboard layout on Windows after start breaks special characters
Classification: Bug

Reproducibility: Always

## Summary

Generally special characters work fine but if you change the keyboard layout after Plover is already started it types the characters that would be in that position on the new keyboard.

## Steps to Reproduce

Start Plover
-  After start  on qwertz : that's 

Switch keyboard layout
-  After switching to qwerty: that?s

Log entry in either case:
2016-03-10 22:25:56,293 Translation(('THATS',) : that's)

## Version

I first noticed the issue after updating plover from the master branch 3151c85 and pulled ec93718 to check if it was fixed before creating this issue. The problem itself might be older because I used a fork for quite a while, though.

I am currently running a fairly up to date version of windows 10.

Edit: Also, capslock seems to affect Plover's output. I usually don't notice because I redirect capslock to ctrl but it probably shouldn't be the case.

morinted:
### About

Check the active layout from the foreground window with each output, and if the keyboard layout has changed, then we reload the keyboard layout. I tried this with English and German in my regional list, then put one window in German, one in English, and switched between them. Plover had to update the layout and still got the symbols right (", @, et cetera)

### Issues

@tarmean, could you verify this fixes #415 ?

### Release Notes

- Handle changing keyboard layouts while Plover is running on Windows

### Tested Platforms

Windows 7

### DIFF
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -19,11 +19,13 @@ import functools
 import threading
 import types
 import win32api
+import win32gui
 import pyHook
 
 SendInput = ctypes.windll.user32.SendInput
 MapVirtualKey = ctypes.windll.user32.MapVirtualKeyW
 GetKeyboardLayout = ctypes.windll.user32.GetKeyboardLayout
+GetWindowThreadProcessId = ctypes.windll.user32.GetWindowThreadProcessId
 LONG = ctypes.c_long
 DWORD = ctypes.c_ulong
 ULONG_PTR = ctypes.POINTER(DWORD)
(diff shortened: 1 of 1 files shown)


morinted:
@tarmean The build is available here: https://ci.appveyor.com/project/morinted/plover/build/147-cgehsndr/artifacts

benoit-pierre:
Works fine when switching between Qwerty/Azerty on a Windows 10 VM.

plover/oslayer/winkeyboardcontrol.py
```diff
@@ -481,6 +491,11 @@ def _key_press(self, keyname):
         self._key_down(keyname)
         self._key_up(keyname)
 
+    def _refresh_keyboard_layout(self):
+        layout_id = WindowsKeyboardLayout.current_layout_id()
+        if layout_id is not self.keyboard_layout.layout_id:
```
benoit-pierre: Is testing for object equality really needed?

morinted: No, you're right. I'll adjust it.

morinted: Amended.

PR #421, opened 2016-03-12
treal: automatically reconnect

morinted:
- Add Treal reconnect, which starts after losing the connection.
- Try to connect to all 10 Treal devices, which can be switched using a switch on the bottom of the machine.
- Change to hidapi on Windows in order to unify modules cross platform.

Tested on OS X and Windows (used AppVeyor builds)


morinted:
For fun I took a video of this:

https://goo.gl/photos/LqQwSjvp8zXr43ZR8

morinted:
Pushed your patch to get the dependencies.

PR #422, opened 2016-03-12
Support --help and --version arguments to the plover binary.

timthelion:
### About

This adds support for the `--help` and `--version` arguments when running plover at the cli.

### Reason

I ran plover with `--help` and was rather surprised that it just ignored the argument and launched.

### Tested Platforms

Linux

### DIFF
--- a/plover/main.py
+++ b/plover/main.py
@@ -6,6 +6,7 @@
 import os
 import sys
 import traceback
+import argparse
 
 WXVER = '3.0'
 if not hasattr(sys, 'frozen'):
@@ -22,6 +23,8 @@ import plover.oslayer.processlock
 from plover.oslayer.config import CONFIG_DIR, ASSETS_DIR
 from plover.config import CONFIG_FILE, DEFAULT_DICTIONARIES, Config
 from plover import log
+from plover import __name__ as __software_name__
+from plover import __version__
 
 def show_error(title, message):
     """Report error to the user.
@@ -69,6 +72,11 @@ def init_config_dir():
 
 def main():
     """Launch plover."""
+    usage = "usage: plover"
+    description = "Run the plover stenotype engine. This is a graphical application."
+    parser = argparse.ArgumentParser(usage=usage, description=description, prog=__software_name__.capitalize())
+    parser.add_argument('--version', action='version', version='%(prog)s ' + __version__)
+    args =  parser.parse_args(args=sys.argv[1:])
     try:
         # Ensure only one instance of Plover is running at a time.
         with plover.oslayer.processlock.PloverLock():


benoit-pierre:
Thanks, but `optparse` is deprecated, `argparse` should be used instead. Also it's a little too much information for `--version` IMHO, a simple `Plover 2.5.8` would suffice.

timthelion:
I switched to using argparse, but is there a particual reason why you do NOT want the extra info?

timthelion:
I provided one more commit that provides the shorter version output, so you can squash them if you want both...

benoit-pierre:
I would expect the extra information to be retrieved with the distribution package manager. This much information is not in line with how other programs behave. Also, you can get `argparse` to do most of the work for you: see the help for the `version` action.

timthelion:
Better?

benoit-pierre:
Getting there.

plover/main.py
```diff
@@ -22,6 +23,13 @@
 from plover.oslayer.config import CONFIG_DIR, ASSETS_DIR
 from plover.config import CONFIG_FILE, DEFAULT_DICTIONARIES, Config
 from plover import log
+from plover import __name__ as __software_name__
+from plover import __version__
+from plover import __copyright__
+from plover import __long_description__
+from plover import __url__
+from plover import __credits__
+from plover import __license__
```
benoit-pierre: Please cleanup unused imports.

timthelion:
I see that you're still not happy :P

plover/main.py
```diff
@@ -69,6 +77,11 @@ def copy_dictionary_to_config(name):
 
 def main():
     """Launch plover."""
+    usage = "usage: plover"
+    description = "Run the plover stenotype engine. This is a graphical application."
+    parser = argparse.ArgumentParser(usage=usage,description=description,prog=__software_name__.capitalize())
```
benoit-pierre: `s/,/, /g`

timthelion:
Aha, sorry about the imports.

benoit-pierre:
I would further tweak it like this:
```diff
diff --git i/plover/main.py w/plover/main.py
index 4ea63eb..39d6dff 100644
--- i/plover/main.py
+++ w/plover/main.py
@@ -72,11 +72,11 @@ def init_config_dir():

 def main():
     """Launch plover."""
-    usage = "usage: plover"
     description = "Run the plover stenotype engine. This is a graphical application."
-    parser = argparse.ArgumentParser(usage=usage, description=description, prog=__software_name__.capitalize())
-    parser.add_argument('--version', action='version', version='%(prog)s ' + __version__)
-    args =  parser.parse_args(args=sys.argv[1:])
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument('--version', action='version', version='%s %s'
+                        % (__software_name__.capitalize(), __version__))
+    args = parser.parse_args(args=sys.argv[1:])
     try:
         # Ensure only one instance of Plover is running at a time.
         with plover.oslayer.processlock.PloverLock():
```

Rational:
* if you override `usage`, you don't get the more useful automatic form with the options, e.g.: `usage: plover [-h] [--version]`
* it's better to not override `prog` and let `argparse` automatically set it based on `argv[0]` (so the name is correct if the main script is renamed, or a symlink, useful when installing multiple versions)

timthelion:
OK, is there a way for you to accept your version and ignore mine? Or do I need to do something at this point?

benoit-pierre:
I can merge manually on my side and push to GitHub. @morinted, are you okay with this?

timthelion:
In any case, it is one am and I am off to bed.

morinted:
Yes, that works for me.

Issue #125, opened 2013-10-10
Add a way to save raw steno


balshetzer:
Provide a way to save paper tape output for steno student homework.

morinted:
Steno display now scrolls... make we could add an export function to that window?

timthelion:
This will be fixed once #423 is merged.

PR #423, opened 2016-03-12
Save notes

### LINKED ISSUE #406: Extend steno notes and offer export
Classification: Feature (New)

## Summary

Request to print notes.

It would be good to be able to print the notes -- or at least be able to make a .pdf with them.  The idea there is when you take one of the court reporting exams, sometimes they want your notes, and it would be good to be able to provide them.   

## Steps to Reproduce

1. I do this -- open the Notes Display
2. I do that -- and try to print.  

## Expected Results

3. This should happen: Plover should be able to use the system's printer routines to produce a PDF, or maybe even print to paper.  

## Actual Results

3. Instead, this happens. :(  Currently, there is no evidence that the printing routine exists -- no menu, nothing through a right-click (or control-click on the Mac); hence this request.  

## Version

Plover version 2.5.8+317.gced379f

timthelion:
### About

Scroll infintely in the notes window. Allow saving the notes.

### Issues

Fixes #406, #125

### DIFF
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -13,6 +13,7 @@ TITLE = 'Plover: Stroke Display'
 ON_TOP_TEXT = "Always on top"
 UI_BORDER = 4
 MAX_STROKE_LINES = 30
+MAX_STROKES_IN_LOG = 2000000
 STYLE_TEXT = 'Style:'
 STYLE_PAPER = 'Paper'
 STYLE_RAW = 'Raw'
@@ -21,7 +22,7 @@ STYLES = [STYLE_PAPER, STYLE_RAW]
 class StrokeDisplayDialog(wx.Dialog):
     
     other_instances = []
-    strokes = deque(maxlen=MAX_STROKE_LINES)
+    strokes = deque(maxlen=MAX_STROKES_IN_LOG)
 
     def __init__(self, parent, config):
         self.config = config        
@@ -72,7 +73,7 @@ class StrokeDisplayDialog(wx.Dialog):
         sizer.Add(wx.StaticLine(self), flag=wx.EXPAND)
 
         self.listbox = wx.TextCtrl(self,
-                                   style=wx.TE_MULTILINE|wx.TE_READONLY|wx.TE_DONTWRAP|wx.BORDER_NONE|wx.HSCROLL,
+                                   style=wx.TE_MULTILINE|wx.TE_READONLY|wx.TE_DONTWRAP|wx.BORDER_NONE|wx.HSCROLL|wx.VSCROLL,
                                    # Will show MAX_STROKE_LINES lines.
                                    size=wx.Size(scroll_width + text_width,
                                                 scroll_height + text_height * MAX_STROKE_LINES))
@@ -105,8 +106,6 @@ class StrokeDisplayDialog(wx.Dialog):
         event.Skip()
         
     def show_text(self, text):
-        if len(self.line_lengths) == MAX_STROKE_LINES:
-            self.listbox.Remove(0, self.line_lengths.pop(0))
         if self.listbox.IsEmpty():
             self.listbox.AppendText(text)
         else:
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -4,10 +4,13 @@
 """A gui display of recent strokes."""
 
 import wx
+import time
+import os
 from wx.lib.utils import AdjustRectToScreen
 from collections import deque
 from plover import system
 from plover.gui.util import find_fixed_width_font
+from plover import log
 
 TITLE = 'Plover: Stroke Display'
 ON_TOP_TEXT = "Always on top"
@@ -18,6 +21,12 @@ STYLE_TEXT = 'Style:'
 STYLE_PAPER = 'Paper'
 STYLE_RAW = 'Raw'
 STYLES = [STYLE_PAPER, STYLE_RAW]
+CLEAR_BUTTON_NAME = 'Clear'
+SAVE_BUTTON_NAME = 'Save'
+CLEAR_PROMPT = 'Would you like to clear the notes?'
+CLEAR_PROMPT_TITLE = 'Clear?'
+SAVE_NOTES_DIALOG_TITLE = 'Save notes'
+SAVE_NOTES_FILE_DIALOG_TEXT_FILE_FILTER = 'Text files (*.txt)|*'
 
 class StrokeDisplayDialog(wx.Dialog):
     
@@ -82,6 +91,21 @@ class StrokeDisplayDialog(wx.Dialog):
         sizer.Add(self.listbox,
                   flag=wx.ALL|wx.EXPAND|wx.ALIGN_LEFT,
                   border=UI_BORDER)
+
+        buttons = wx.BoxSizer(wx.HORIZONTAL)
+        sizer.Add(buttons, flag=wx.EXPAND|wx.ALL)
+
+        clear_button = wx.Button(self, id=wx.ID_CLEAR, label=CLEAR_BUTTON_NAME)
+        buttons.AddSpacer(5)
+        buttons.Add(clear_button, flag=wx.EXPAND|wx.ALIGN_LEFT)
+        self.Bind(wx.EVT_BUTTON, self._clear, clear_button)
+        buttons.AddStretchSpacer()
+        save_button = wx.Button(self, id=wx.ID_SAVE, label=SAVE_BUTTON_NAME)
+        buttons.Add(save_button, flag=wx.EXPAND|wx.ALIGN_RIGHT)
+        buttons.AddSpacer(5)
+        self.Bind(wx.EVT_BUTTON, self._save, save_button)
+
+        sizer.AddSpacer(5)
 
         self.on_style()
 
@@ -149,6 +173,24 @@ class StrokeDisplayDialog(wx.Dialog):
         text = ''.join(text)
         return text        
 
+    def _clear(self, event=None):
+        message_dialog = wx.MessageDialog(self, CLEAR_PROMPT, CLEAR_PROMPT_TITLE)
+        message_dialog.SetOKLabel("Clear")
+        if message_dialog.ShowModal() == wx.ID_OK:
+            self.listbox.Clear()
+            self.strokes.clear()
+
+    def _save(self, event=None):
+        file_name_suggestion = 'steno-notes-%s.txt' % time.strftime('%Y-%m-%d-%H-%M')
+        file_dialog = wx.FileDialog(self, SAVE_NOTES_DIALOG_TITLE, "", file_name_suggestion,
+                                    SAVE_NOTES_FILE_DIALOG_TEXT_FILE_FILTER, wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT)
+        if file_dialog.ShowModal() == wx.ID_OK:
+            try:
+                with open(file_dialog.GetPath(), 'w') as fd:
+                    fd.write(self.listbox.GetValue()+'\n')
+            except IOError:
+                log.error("Cannot save notes in file '%s'." % file_dialog.GetPath(), exc_info=True)
+
     def raw_format(self, stroke):
         return stroke.rtfcre
 
@@ -168,7 +210,6 @@ class StrokeDisplayDialog(wx.Dialog):
     def display(parent, config):
         # StrokeDisplayDialog shows itself.
         StrokeDisplayDialog(parent, config)
-
 
 class fake_config(object):
     def __init__(self):


timthelion:
Oops, moment, gotta remove those extraneous commits

timthelion:
There, that's better...

timthelion:
Tested on Linux

morinted:
Thanks for this. Feedback incoming...

- You should provide a format hint when saving (`.txt`), and perhaps a name (steno-notes-2016-12-30.txt)
    Save dialogue
- The button placement doesn't jibe well with me.

   Stroke display

You could extend the buttons so that they are 50/50 at the bottom, and if they look too long, add some slight padding in all directions (might help indicate they are buttons).

- When saving, the dialogue halts the program, and I cannot write a new name if the steno machine is in Keyboard mode (because it won't output anything while blocked, and because it is blocking my keyboard output)
- In general, you've been calling it "Transcription", but that's not the right word. Transcription would refer to the end product. "Stroke display" is the window, and it displays strokes using "notes" or "raw steno". Use accordingly... I imagine most people saving will be doing so because in steno school you need to submit your notes.
- febe3ec could be squashed into 3238b13
- If you are not using a UI element in other parts of the application, there's no need to save it as a class member (e.g. `self.clear_button` is never refered to out of local scope, so it could be `clear_button`)
- Might be style, but if there's no *reason* to use a unicode-encoded string, why use it? I get there's no reason *not to*, but maybe we can just be consistent and only use `u''` when using Unicode characters.
- Because the clear button is right next to the save button, it would be good to confirm on clear to ensure no terrible accidents.
- Otherwise, it's a great little change. The save does work well, and it will solve a problem that has existed for quite a few years. :) Thanks for the pull request.

timthelion:
Uf, on the blocking. Somehow, I didn't think of that, and it didn't come up in my testing because I'm not a steno user yet :P

Are there any other blocking dialogs in Plover that I could take inspiration from?

timthelion:
I guess what I need to do is to Show the file dialog non-modaly and then somehow wire up a hook to do the actual saving...

morinted:
Couldn't you just make either the save in a new thread?

timthelion:
Well, the thing that is blocking is the wx dialog. I think that you cannot/should not create a new thread and then call wx functions from that thread if the parent window has been launched from another thread because wx is a wrapper for gtk on linux and gtk is not threaded. I have to figure out a way to launch the file save dialog as a non-modal dialog. Modal vs nonmodal in wx is blocking or non blocking. in the rest of the world.

timthelion:
I think that I have addressed all of your concerns. Can you please take another look?

timthelion:
Wait, sorry, this needs more testing... Moment.

timthelion:
The problem is, that I managed to make the dialog non-modal but hooking up the `EVT_CLOSE` doesn't work. That event doesn't get triggered. And monkey patching also doesn't work because the actual execution is happening in c land. Back to the drawing board.

timthelion:
I'm not the first to run into this, and it seems that the general answer is that it's just not possible to not block on a FileDialog http://wxpython-users.1045709.n5.nabble.com/wxPython-modeless-wxDialog-tt2285276.html#a2285284

timthelion:
I think architecturally, the right thing(tm) is to have the event loop for plover, that is handling the keypress events actually run in a different thread than wx is running in.

timthelion:
I just uploaded something that seems to work. I cannot make any gaurantees about it's resiliance, but hopefully it will do.

morinted:
All rightey, I tried it out on my system, and just to note the "clear" button also blocks output so I can't dismiss/accept it with the steno machine.

The save button does nothing at all for me. Maybe @benoit-pierre would have some advice for heading in the right direction here.

I do like the button placement more now, though. Thanks.

timthelion:
Perhaps you didn't get the very latest version? The save button works on linux, in the latest version. I'll fix the clear button to non-blocking later today.

morinted:
Yes, the save modal is just not showing... The init function runs but it doesn't do anything. I'll investigate. You'll find wx is not consistent across platforms.

That being said, looks like the dictionary add dialog is blocking, too, so it's not a big problem if this one is blocking too.

morinted:
You should also use PEP8 to format the code, please. As a rule, local class methods should be "_leading_snake_case"

benoit-pierre:
This indeed works fine on Linux. I think we should keep the use of modal dialogs (simpler code), and try to fix the underlying issue on OSX.

morinted:
The underline issue being that the engine doesn't run in a different thread from the UI?

benoit-pierre:
I think that globally our handling of threads leaves a lot to be desired: except for making sure WX calls are done in the main thread (necessary to prevent crashes), we don't use locks, and it really only (mostly) works because the majority of events are generated by user actions and the real time constraints are really low.

plover/gui/paper_tape.py
```diff
@@ -170,6 +198,26 @@ def display(parent, config):
...
         StrokeDisplayDialog(parent, config)
 
+class NonModalSaveDialog(wx.FileDialog):
+    def __init__(self,parent,notes_to_save):
+        self.notes_to_save = notes_to_save
+        file_name_suggestion = None
+        counter = 0
+        while not file_name_suggestion or os.path.exists(file_name_suggestion):
+            file_name_suggestion = "steno-notes-"+time.strftime("%Y-%M-%d")+"("+str(counter)+").txt"
+            counter += 1
```
morinted: The string generation could be made a little more readable with string formatting:

```python
        counter = 0
        while not file_name_suggestion or os.path.exists(file_name_suggestion):
            counter_string = ' (%i)' % counter if counter else ''
            file_name_suggestion = "steno-notes-%s%s.txt" % (time.strftime("%Y-%M-%d"), counter_string)
            counter += 1
```

timthelion:
What do you consider to be an internal class method? Do you consider the `save` and `clear` event handlers to be internal?

morinted:
Yes, I'd say the event handlers are internal, because you don't intend them to be imported by other modules.

Also, Benoit Pierre fixed the issue with the blocking wait, we don't wait for the GUI thread to send output anymore.

timthelion:
OK, I think this is done! I hope I hope I hope ;)

plover/gui/paper_tape.py
```diff
@@ -150,6 +168,29 @@ def paper_format(self, stroke):
...
+            file_name_suggestion = 'steno-notes-%s%s.txt' % (time.strftime('%Y-%m-%d'), counter_string)
+            counter += 1
+        file_dialog = wx.FileDialog(self, "Save notes", "", file_name_suggestion,
+                                   "txt files (*.txt)|*", wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT)
+        if file_dialog.ShowModal() == wx.ID_OK:
+            try:
+              with open(file_dialog.GetPath(), 'w') as fd:
+                  fd.write(self.listbox.GetValue())
+            except IOError as e:
+                wx.LogError("Cannot save notes in file '%s'.\n'%s"%(file_dialog.GetPath(),str(e)))
```
benoit-pierre: Use the Plover log:
```python
from plover import log
[...]
            except IOError:
                log.error('saving notes to %s', file_dialog.GetPath(), exc_info=True)
```

plover/gui/paper_tape.py
```diff
@@ -150,6 +168,29 @@ def paper_format(self, stroke):
...
+        while not file_name_suggestion or os.path.exists(file_name_suggestion):
+            counter_string = ' (%i)' % counter if counter else ''
+            file_name_suggestion = 'steno-notes-%s%s.txt' % (time.strftime('%Y-%m-%d'), counter_string)
+            counter += 1
+        file_dialog = wx.FileDialog(self, "Save notes", "", file_name_suggestion,
+                                   "txt files (*.txt)|*", wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT)
+        if file_dialog.ShowModal() == wx.ID_OK:
+            try:
+              with open(file_dialog.GetPath(), 'w') as fd:
+                  fd.write(self.listbox.GetValue())
```
benoit-pierre: Please an extra '\n' for the last line.

plover/gui/paper_tape.py
timthelion: The current code shows a dialog that lets the user know that the file couldn't be saved, and why. Using the plover log just prints to stdout. This isn't debugging code, this is UX, so that when the user tries to save to a read only directory or the disk is full, the user finds out...

plover/gui/paper_tape.py
timthelion: The last line of the saved file? I don't understand.

plover/gui/paper_tape.py
benoit-pierre: The Plover log does not just print to stdout, there is a custom GUI handler, with a notification.

plover/gui/paper_tape.py
benoit-pierre: The last line is missing an EOL.

plover/gui/paper_tape.py
timthelion: When I tested it out just now, there wasn't any dialog.

plover/gui/paper_tape.py
timthelion: The last line of what? The saved file?

plover/gui/paper_tape.py
benoit-pierre: Then it's a bug, you don't get a notification? You using Linux right? What WM are you using?

plover/gui/paper_tape.py
benoit-pierre: Yes, sorry, not the code.

plover/gui/paper_tape.py
timthelion: I am on linux with stumpwm. (Are you sending the notification via dbus, because I'm running plover in a sepparate dbus namespace). But I think it is better to have a dialog and specifically not a notification. If the user saves something and their saving failed, then they should be loudly told, so they don't end up thinking that they saved successfully.

benoit-pierre: We already use the log in other cases, if the notification system does not work in your case, I'd rather fix it, and use the same code. dbus notify is used if available, with a fallback to `wx.NotificationMessage` if not.

benoit-pierre: What do you get when running `python2 -c 'from pynotify import *; init("test"); print get_server_info(), get_server_caps()'`?

timthelion:
````
  File "<string>", line 1, in <module>
ImportError: No module named pynotify
````

So, I just don't have it installed apparently.

**Comment by timthelion** [comment]
So, I cannot test the notification code, but is this version to your liking?

**Comment by benoit-pierre** [comment]
If pynotify cannot be used, then wx is, so you really should get a notification box anyway... The code is okay.

**Comment by morinted** [comment]
It's looking pretty good, thanks for being so responsive.

Just three little things left.

1. The GUI strings for the dialogues should be in constants like you have already done for CLEAR_BUTTON_NAME.
2. The save commit should be reworded to remove "transcript". 
3. The OS path naming check probably won't do anything with just a file name, you could add in the current working directory (I tried this and it seems pretty reliable):

```diff
diff --git a/plover/gui/paper_tape.py b/plover/gui/paper_tape.py
index 72fb4c5..0947a4f 100644
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -179,12 +179,13 @@ class StrokeDisplayDialog(wx.Dialog):
     def _save(self, event=None):
         file_name_suggestion = None
         counter = 0
-        while not file_name_suggestion or os.path.exists(file_name_suggestion):
+        cwd = os.getcwd()
+        while not file_name_suggestion or os.path.isfile(os.path.join(cwd, file_name_suggestion)):
             counter_string = ' (%i)' % counter if counter else ''
             file_name_suggestion = 'steno-notes-%s%s.txt' % (time.strftime('%Y-%m-%d'), counter_string)
             counter += 1
-        file_dialog = wx.FileDialog(self, "Save notes", "", file_name_suggestion,
-                                   "txt files (*.txt)|*", wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT)
+        file_dialog = wx.FileDialog(self, "Save notes", cwd, file_name_suggestion,
+                                    "Text files (*.txt)|*", wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT)
         if file_dialog.ShowModal() == wx.ID_OK:
             try:
                 with open(file_dialog.GetPath(), 'w') as fd:
```

timthelion:
Did you have a problem with the current name suggestion code on Mac? Because it works on linux.

morinted:
Yes, without providing the directory in the file dialog, the wx directory
defaults to my last used directory. So we could just hand in the cwd to wx

timthelion:
Aha, I see the problem. But it's actually better to use the "last used" directory than the "cwd" which is probably locked to whatever directory plover was started from. Which makes the file name suggestion useless, because no one is going to save their notes to the desktop or whatever the default cwd is when launching plover from the applications menu.

timthelion:
Does plover have any way of saving settings, that I could save a `last-notes-directory` path to? That would be the best. Is to save the directory which was last used to select a destination for saving notes to.

morinted:
I wouldn't store that in config because our config should ideally be
portable. The file dialog has a flag to update the working directory, but I
think that may break config reloading. Maybe for now we could just use
YYYY-M-D-h:m to avoid conflicts?

timthelion:
Well, I can make it portable by making it default to `$HOME` if `last-notes-directory` doesn't exist.

timthelion:
It can't be `HH:MM` because `:` is a restricted character in file names on Windows.

timthelion:
Just uploaded a version with `HH-MM`.

timthelion:
You guys realize, that there are now 50 comments discussing 45 lines of code, right?

morinted:
Better 50 comments than 0 :) It's looking good. Only thing I noticed is minor complaints from PEP8 when viewing the code:

```diff
diff --git a/plover/gui/paper_tape.py b/plover/gui/paper_tape.py
index 46c5ec0..373da3d 100644
--- a/plover/gui/paper_tape.py
+++ b/plover/gui/paper_tape.py
@@ -183,12 +183,12 @@ class StrokeDisplayDialog(wx.Dialog):
     def _save(self, event=None):
         file_name_suggestion = 'steno-notes-%s.txt' % time.strftime('%Y-%m-%d-%H-%M')
         file_dialog = wx.FileDialog(self, SAVE_NOTES_DIALOG_TITLE, "", file_name_suggestion,
-                                   SAVE_NOTES_FILE_DIALOG_TEXT_FILE_FILTER, wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT)
+                                    SAVE_NOTES_FILE_DIALOG_TEXT_FILE_FILTER, wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT)
         if file_dialog.ShowModal() == wx.ID_OK:
             try:
                 with open(file_dialog.GetPath(), 'w') as fd:
                     fd.write(self.listbox.GetValue()+'\n')
-            except IOError as e:
+            except IOError:
                 log.error("Cannot save notes in file '%s'." % file_dialog.GetPath(), exc_info=True)

     def raw_format(self, stroke):
```

Should be good to merge. Thanks for your perseverance!

timthelion:
Is it good now?

morinted:
Yup! Your patience is appreciated. I'm looking forward to hearing what people think of the feature after releasing the next weekly, this was been a long time coming (#125 is from October 2013!)

Thanks for taking the time.

PR #429, opened 2016-03-14
gui/main: don't defer output to the GUI thread

benoit-pierre:
Output from the machine thread, this fix an issue with modal dialogs not receiving output on Mac OSX.

Tested on Linux and Windows 10 VM.


morinted:
Works well on OS X.

PR #435, opened 2016-03-18
Fix Windows and Mac distributions so pkg_resources can be used

benoit-pierre:
Add the required packages metadata so `pkg_resources` can be used.

Tested with a Windows 10 VM.

### DIFF
--- a/osx/makefile
+++ b/osx/makefile
@@ -19,9 +19,7 @@ app: $(APP)
 dmg: $(DMG)
 
 $(APP):
-	cd $(topdir) && $(PYTHON) setup.py py2app
-	zip -d $(TMP_APP)/Contents/Resources/lib/python2.7/site-packages.zip 'plover/assets/*'
-	rm -rf $(TMP_APP)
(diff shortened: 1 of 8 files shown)


osx/makefile
```diff
@@ -21,7 +21,6 @@ dmg: $(DMG)
 $(APP):
 	cd $(topdir) && $(PYTHON) setup.py py2app
 	zip -d $(TMP_APP)/Contents/Resources/lib/python2.7/site-packages.zip 'plover/assets/*'
-	ditto --arch x86_64 $(TMP_APP) $(APP)
```
morinted: Seems like even with the py2app flag, the DMG size benefits from the ditto stripping. On my machine, the DMG is ~80MB, and after ditto it is 50MB.

morinted:
Looked through everything, it all seems good other than the ditto still being useful. Is it worth pulling this in for the weekly today?

benoit-pierre:
OK, I added back a call to ditto.

PR #438, opened 2016-03-20
oslayer/xkeyboardcontrol: add support for media keys

benoit-pierre:
Add support for media keys on Linux. Both 'XF86_AudioMute' and 'AudioMute' are supported.

### DIFF
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -162,9 +162,9 @@ class WindowsKeyboardLayout:
     # No letters yet, those are customized per-layout.
     keyname_to_keycode = collections.defaultdict(list, {
         # Adding media controls for windows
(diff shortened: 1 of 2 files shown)


benoit-pierre:
I just pushed an additional commit to fix media keys names for Windows: use the same names as for X11 (sans the 'XF86_' prefix).

Note: it's untested, as I don't have a working audio driver on my Windows 10 VM.

morinted:
Okay! Thanks for the super quick turnaround. I'll test this on Windows
before merging. Is there a list of possible XF86_ keys somewhere?

benoit-pierre:
List of supported keys: `python2 -c 'from Xlib.keysymdef import xf86; print "\n".join(sym for sym in dir(xf86) if sym.startswith("XK_"))'`. Full list in `/usr/include/X11/XF86keysym.h`.

PR #439, opened 2016-03-21
OSX: Add special media keys, extend Windows' Media Keys

morinted:
Add OSX media keys. Supported keys (x names):

```python
    "AudioRaiseVolume": 0,
    "AudioLowerVolume": 1,
    "MonBrightnessUp": 2,
    "MonBrightnessDown": 3,
    "AudioMute": 7,
    "Num_Lock": 10,
    "Eject": 14,
    "AudioPause": 16,
    "AudioPlay": 16,
    "AudioNext": 17,
    "AudioPrev": 18,
    "AudioRewind": 20,
```

I didn't include keyboard brightness up, down, and toggle. There doesn't seem to be an XF86 equivalent, and I'm not sure what the value of controlling keyboard brightness from the steno machine would be.

Also added the Fn modifier, but it's not exactly expected behavior because it is an external keyboard--the top-row equivalent of the Mac laptop keyboards won't occur.

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -32,6 +32,8 @@ from Quartz import (
     kCGKeyboardEventKeycode,
     kCGSessionEventTap,
     kCGEventSourceStateHIDSystemState,
+    NSEvent,
+    NSSystemDefined,
 )
 import Foundation
 import threading
@@ -42,13 +44,16 @@ from plover.oslayer import mac_keycode
 
 BACK_SPACE = 51
 
+NX_KEY_OFFSET = 65536
+
 KEYNAME_TO_KEYCODE = collections.defaultdict(list, {
+    'Fn': [63],
 
     'Alt_L': [58], 'Alt_R': [61], 'Control_L': [59], 'Control_R': [62],
     'Hyper_L': [], 'Hyper_R': [], 'Meta_L': [], 'Meta_R': [],
     'Shift_L': [56], 'Shift_R': [60], 'Super_L': [55], 'Super_R': [55],
 
-    'Caps_Lock': [57], 'Num_Lock': [], 'Scroll_Lock': [], 'Shift_Lock': [],
+    'Caps_Lock': [57], 'Scroll_Lock': [], 'Shift_Lock': [],
 
     'Return': [36], 'Tab': [48], 'BackSpace': [BACK_SPACE], 'Delete': [117],
     'Escape': [53], 'Break': [], 'Insert': [], 'Pause': [], 'Print': [],
@@ -139,6 +144,23 @@ LITERALS = collections.defaultdict(str, {
     '?': 'question', '\t': 'Tab', ' ': 'space'
 })
 
+NX_KEYS = {
+    "AudioRaiseVolume": 0,
+    "AudioLowerVolume": 1,
+    "MonBrightnessUp": 2,
+    "MonBrightnessDown": 3,
+    "AudioMute": 7,
+    "Num_Lock": 10,
+    "Eject": 14,
+    "AudioPause": 16,
+    "AudioPlay": 16,
+    "AudioNext": 17,
+    "AudioPrev": 18,
+    "AudioRewind": 20,
+    "KbdBrightnessUp": 21,
+    "KbdBrightnessDown": 22
+}
+
 # Maps from keycodes to corresponding event masks.
 MODIFIER_KEYS_TO_MASKS = {
     58: kCGEventFlagMaskAlternate,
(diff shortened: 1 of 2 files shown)


benoit-pierre:
The XF86 equivalent is:
```
XK_XF86_KbdBrightnessDown
XK_XF86_KbdBrightnessUp
```
No toggle.

benoit-pierre:
What about Windows? We only support a subset of those media keys, can we update the list?

morinted:
Sure. I was presented with a convenient list for OSX but I'll check into what Windows can do.

benoit-pierre:
I have this in `/usr/include/wine/windows/winuser.rh`:

```
#define VK_BROWSER_BACK        0xA6
#define VK_BROWSER_FORWARD     0xA7
#define VK_BROWSER_REFRESH     0xA8
#define VK_BROWSER_STOP        0xA9
#define VK_BROWSER_SEARCH      0xAA
#define VK_BROWSER_FAVORITES   0xAB
#define VK_BROWSER_HOME        0xAC
#define VK_VOLUME_MUTE         0xAD
#define VK_VOLUME_DOWN         0xAE
#define VK_VOLUME_UP           0xAF
#define VK_MEDIA_NEXT_TRACK    0xB0
#define VK_MEDIA_PREV_TRACK    0xB1
#define VK_MEDIA_STOP          0xB2
#define VK_MEDIA_PLAY_PAUSE    0xB3
#define VK_LAUNCH_MAIL         0xB4
#define VK_LAUNCH_MEDIA_SELECT 0xB5
#define VK_LAUNCH_APP1         0xB6
#define VK_LAUNCH_APP2         0xB7
```

benoit-pierre:
More official list: https://msdn.microsoft.com/en-us/library/windows/desktop/dd375731(v=vs.85).aspx

morinted:
Added some Windows keys, they don't all do what they are supposed to on Windows 10. e.g. the Calculator key doesn't work in Windows 10, but did in 7... Might be that this doesn't work without a keyboard "driver".

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -344,7 +365,10 @@ def send_key_combination(self, combo_string):
         # that, if executed in order, would emulate the key combination
         # represented by the argument.
         def _keystring_to_sequence(keystring):
-            if keystring in KEYNAME_TO_KEYCODE:
+            if keystring in NX_KEYS:
+                self._send_media_key(NX_KEYS[keystring])
+                seq = []
```
benoit-pierre: Could not this code result in sending key events in the wrong order? E.g. when using something like `{#RegularKey MediaKey}`?

morinted: You're right. I'll have to revisit this.

morinted:
Okay, modified the OSX version so that I send it in sequence like existing keycodes.

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -431,24 +469,27 @@ def _send_sequence(sequence):
         mods_flags = 0
 
         for keycode, key_down in sequence:
-            if not key_down and keycode in MODIFIER_KEYS_TO_MASKS:
-                mods_flags &= ~MODIFIER_KEYS_TO_MASKS[keycode]
+            if keycode >= NX_KEY_OFFSET:
+                KeyboardEmulation._send_media_key(keycode - NX_KEY_OFFSET, key_down)
+            else:
```
benoit-pierre: How about using a continue here, so the rest of the code does not need to be shifted? Alternatively, change `_send_media_key` to `_get_media_event`, and factorize the call to `CGEventPost`? Is the fact that the event tap location is 0 (`kCGHIDEventTap`?) and not `kCGSessionEventTap` required?

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -417,10 +442,23 @@ def _set_event_string(event, s):
...
     )
 
     @staticmethod
+    def _send_media_key(key_id, key_down):
+        # Credit: https://gist.github.com/fredrikw/4078034
+        event = NSEvent.otherEventWithType_location_modifierFlags_timestamp_windowNumber_context_subtype_data1_data2_(
+            NSSystemDefined, (0,0),
+            0xa00 if key_down else 0xb00,
+            0, 0, 0, 8,
+            (key_id << 16) | ((0xa if key_down else 0xb) << 8),
```
benoit-pierre: This could be simplified/factorized a little:
```python
	flags = 0xa00 if key_down else 0xb00
	event = NSEvent.otherEventWithType_location_modifierFlags_timestamp_windowNumber_context_subtype_data1_data2_(
	    NSSystemDefined, (0,0),
	    flags,
	    0, 0, 0, 8,
	    (key_id << 16) | flags,
	    -1
	)
	CGEventPost(0, event.CGEvent())
```

morinted:
Thanks for the feedback. I'll update the PR tomorrow most likely. I think after getting this in, we should do a weekly to get a small test bed for the metadata change, and then consider it a feature freeze for v3.0.0 for Saturday.

morinted:
I copied your flags factorization, and decided to take your advice for the "get_media_key" method.

PR #443, opened 2016-03-21
a few dictionary related changes

benoit-pierre:
2 sets of changes:
* changes that I think should be part of the next release:
 - fix call to `util.SetTopApp`
 - change steno dictionary save API, so the dictionary editor can be made to only save back those dictionaries that have been modified (instead of always saving **all** dictionaries)
* preliminary changes for multilingual/plugins support (if deemed too risky, I'll rewrite the PR to remove those, and defer them after the release):
 - don't uppercase strokes in the add translation dialog or dictionary editor: some systems use lower case letters too (e.g. Italian)
 - better handle *read-only* dictionaries, i.e. if a dictionary's `save` field is `None`; this will be needed for proper support of Python dictionaries and dictionaries loaded directly from assets

### DIFF
--- a/plover/dictionary_editor_store.py
+++ b/plover/dictionary_editor_store.py
@@ -141,7 +141,7 @@ class DictionaryEditorStore():
             dict = (self.engine
                         .get_dictionary()
                         .get_by_path(added_item.dictionary))
-            dict.__setitem__(self._splitStrokes(added_item.stroke),
+            dict.__setitem__(normalize_steno(added_item.stroke),
                              added_item.translation)
 
         # Updates
@@ -150,7 +150,7 @@ class DictionaryEditorStore():
             dict = (self.engine
                         .get_dictionary()
                         .get_by_path(modified_item.dictionary))
-            dict.__setitem__(self._splitStrokes(modified_item.stroke),
+            dict.__setitem__(normalize_steno(modified_item.stroke),
                              modified_item.translation)
 
         # Deletes
@@ -158,7 +158,7 @@ class DictionaryEditorStore():
             dict = (self.engine
                         .get_dictionary()
                         .get_by_path(deleted_item.dictionary))
-            dict.__delitem__(self._splitStrokes(deleted_item.stroke))
+            dict.__delitem__(normalize_steno(deleted_item.stroke))
 
         self.engine.get_dictionary().save_all()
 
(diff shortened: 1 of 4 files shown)


benoit-pierre:
I have pushed an additional commit to improve selection handling in the dictionary editor:
* track selection (so multiple-rows selections work as expected)
* correctly ignore read-only rows on deletion
* insert new row before the selection first non-read only selected rows (if any), and automatically show/select/focus it

Note: tested on Linux and with a Windows 10 VM, using the following patch (to check read-only dictionaries support):

```diff
diff --git i/plover/dictionary/base.py w/plover/dictionary/base.py
index e528166..74b269e 100644
--- i/plover/dictionary/base.py
+++ w/plover/dictionary/base.py
@@ -41,7 +41,8 @@ def load_dictionary(filename):
         raise DictionaryLoaderException(unicode(e))

     d.set_path(filename)
-    d.save = ThreadedSaver(d, filename, dict_type.save_dictionary)
+    if not filename.endswith('main.json'):
+        d.save = ThreadedSaver(d, filename, dict_type.save_dictionary)
     return d

 def save_dictionary(d, filename, saver):
```

morinted:
All seems good except:

```diff
diff --git a/plover/dictionary_editor_store.py b/plover/dictionary_editor_store.py
index 98839ba..c82337d 100644
--- a/plover/dictionary_editor_store.py
+++ b/plover/dictionary_editor_store.py
@@ -146,18 +146,18 @@ class DictionaryEditorStore():
         # Creates
         for item in self.added_items:
             item.dictionary[normalize_steno(item.stroke)] = item.translation
-            needs_saving.add(dictionary.get_path())
+            needs_saving.add(item.dictionary.get_path())

         # Updates
         for item_id in self.modified_items:
             item = self.all_keys[item_id]
             item.dictionary[normalize_steno(item.stroke)] = item.translation
-            needs_saving.add(dictionary.get_path())
+            needs_saving.add(item.dictionary.get_path())

         # Deletes
         for item in self.deleted_items:
             del item.dictionary[normalize_steno(item.stroke)]
-            needs_saving.add(dictionary.get_path())
+            needs_saving.add(item.dictionary.get_path())

         self.engine.get_dictionary().save(needs_saving)
```

benoit-pierre:
Good catch... Hooray for code review, and shame on me for not testing the save code one last time after rewriting the history... Should be fixed now.

PR #446, opened 2016-03-23
oslayer/keyboardcontrol: fix test code

benoit-pierre:
Tested on Linux. The test code does not work on Windows, but that is a separate issue.


benoit-pierre:
I pushed an additional commit to fix the test code when run under Windows.

PR #447, opened 2016-03-23
System support

benoit-pierre:
It seems like some of the changes for preliminary system support fell through the cracks: the orthography module was not using the system module...


morinted:
Doh...

PR #450, opened 2016-03-23
Update dictionary with feedback from #400

morinted:
- Remove fingerspelled acronyms
- Fix or remove invalid entries
- Add briefs
- Add some new words and phrases
- Change pre/suffix style, e.g. `{^}word` to `{^word}`, where appropriate.
- Remove `{ }`s


benoit-pierre:
> Change pre/suffix style, e.g. {^}word to {^word}, where appropriate.

I'm not sure this is a good idea, as it could lead backward incompatible changes when we want to add support for new meta commands and the characters are already used in the wild, especially characters like `$`, `<`, ... that would be good candidate for extension to the meta commands.

morinted:
Yeah, you're probably right. I'll revert that for most of the symbol entries, but not "-" entries as those are pretty close to real characters in text, like "pre-". I'll also have to update that manual "-cum-" entry as it won't attach the edges.

morinted:
Updated with more cautious use of attach metas, fixed/found some other errors relating to false suffix and prefixes.

PR #458, opened 2016-03-28
move the Windows keyboard capture to its own process

### LINKED ISSUE #454: Plover enabled but suddenly starts outputting qwerty with no apparent trigger
Classification: Bug

Reproducibility: Sometimes/Unpredictable (no clear triggers, happens at random, sometimes within minutes, sometimes within hours)

## Summary

Plover is enabled, functioning as expected, then suddenly starts outputting qwerty while Plover is still enabled. I'm using an NKRO keyboard. The problem can occur partially sometimes, with letters stuck to the ends of words (e.g., "The quickd brown fox jumps over the lazy dog", with the "d" being out of place here), but the bigger problem is when it ends up outputting qwerty completely and never goes back to steno (e.g., "The quick brownwevkl;"). And the partial and complete occurrences are not necessarily sequential and have mostly been isolated so far.

## Steps to Reproduce

1. Launch Plover
2. Enable translation
3. Type normally

## Expected Results

1. Output steno for as long as Plover is enabled

## Actual Results

3. Instead, it stops outputting steno and reverts to qwerty at unpredictable times with no apparent trigger. Stops logging as well. Plover is still enabled. Need to completely restart. :(

## Version

Plover version 2.5.8 + 440.gced379f

Installed via: download from openstenoproject.org.

## Notes

It stops logging when it stops working, so I don't have anything in the logs to show for it. Toggling off and on again doesn't get it working again; have to completely restart. Unplugging and replugging the keyboard with Plover enabled does not reproduce the problem; I can unplug and replug and still have it output steno. Like I said, nothing in particular triggers it. It works just fine and then stops right in the middle of things.

## Configuration

OS:

- Windows 10, 64-bit, x64-based processor (Idk how much info is relevant here, sorry)
-Using NKRO keyboard

benoit-pierre:
This allow the use of `pythoncom.PumpWaitingMessages` without interfering with the GUI loop, and ensure the keyboard hook is run in a timely manner: no competition with other threads and captured key events are posted to queue without further processing. This fixes #454 on Windows.

Note: a custom main with a set of specific workarounds must be used for correct multi-processing support when using PyInstaller, especially with '--onefile'. See: https://github.com/pyinstaller/pyinstaller/wiki/Recipe-Multiprocessing

Tested with a Windows 10 VM: the easiest way to reproduce the bug is to keep a key pressed during Plover startup.

### DIFF
--- a/plover/gui/main.py
+++ b/plover/gui/main.py
@@ -287,7 +287,6 @@ class MainFrame(wx.Frame):
             machine_name = machine_registry.resolve_alias(
                 self.config.get_machine_type())
             self.machine_status_text.SetLabel('%s: %s' % (machine_name, state))
-            self.reconnect_button.Show(state == STATE_ERROR)
             self.spinner.Show(state == STATE_INITIALIZING)
             self.connection_ctrl.Show(state != STATE_INITIALIZING)
             if state == STATE_INITIALIZING:
--- a/plover/oslayer/keyboardcontrol.py
+++ b/plover/oslayer/keyboardcontrol.py
@@ -84,11 +84,7 @@ if __name__ == '__main__':
     kc.start()
     print 'Press CTRL-c to quit.'
     try:
-        if sys.platform.startswith('win32'):
-            while True:
-                pythoncom.PumpWaitingMessages()
-        else:
-            while True:
-                time.sleep(1)
+        while True:
+            time.sleep(1)
     except KeyboardInterrupt:
         kc.cancel()
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -19,10 +19,13 @@ import functools
 import multiprocessing
 import threading
 import types
+import _winreg as winreg
 
 import win32api
 import win32con
 import win32gui
+
+from plover import log
 
 
 SendInput = ctypes.windll.user32.SendInput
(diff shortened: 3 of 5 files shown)


benoit-pierre:
While this version is more robust, I've got a report that the bug can still occur. However, resetting the "machine" (by way of temporarily switching to another machine type) fixes the problem (since the hook is now in another process). Maybe the reconnect button should always be shown for those cases where we cannot detect an error?

The Microsoft official documentation for low level keyboard hooks mentions a workaround:
```
The hook procedure should process a message in less time than the data entry specified in the LowLevelHooksTimeout value in the following registry key:

HKEY_CURRENT_USER\Control Panel\Desktop

The value is in milliseconds. If the hook procedure times out, the system passes the message to the next hook. However, on Windows 7 and later, the hook is silently removed without being called. There is no way for the application to know whether the hook is removed.
```

I can confirm that after setting `LowLevelHooksTimeout` to 5000 and rebooting, I can't reproduce the bug when using the 'keep a key pressed during init' test with the current master, or after introducing a sleep in this PR hook.

morinted:
I'd be okay with the reconnect button always being present. Would help for cases where the serial port doesn't reconnect, too.

As for the low-level timeout…what sort of UI and permissions would be involved with modifying that registry key?

benoit-pierre:
Looks like we do not need any special rights, as the following works:

```python
import _winreg as winreg


KEY_NAME = 'LowLevelHooksTimeout'
KEY_TYPE = winreg.REG_DWORD
KEY_VALUE = 5000

key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, r'Control Panel\Desktop', 0, winreg.KEY_ALL_ACCESS)
try:
    key_value, key_type = winreg.QueryValueEx(key, KEY_NAME)
except WindowsError:
    key_value, key_type = (None, None)

if key_value != KEY_VALUE or key_type != KEY_TYPE:
    print 'changing key from', (key_value, key_type), 'to', (KEY_VALUE, KEY_TYPE)
    winreg.SetValueEx(key, KEY_NAME, 0, KEY_TYPE, KEY_VALUE)
```

benoit-pierre:
But we can't just set it automatically at startup, because a reboot is needed too.

morinted:
Is the default 1000? I'm surprised we are hanging that long. Could we handle this by clocking when we receive the hook vs when we reply to it, and then if we are over the registry value, restart the hook?

I'm wary of changing registry values at the risk of affecting Windows and being non-portable.

benoit-pierre:
I could not find what the default is. Note again that the behavior changed with Windows 7, and there was apparently no limit before. I don't think we can reliably time the hook, as it starts in pyHook (and in the generated SWIG wrapper). We can check the registry and log a warning, and mention the registry key in the documentation.

morinted:
Well, having the short hook won't necessarily break anything, would you want to update the registry key and then notify that Plover may not work fully until after a computer restart?

benoit-pierre:
I've added the registry workaround (with a warning if the key has been updated and a reboot is needed), and the reconnect button is now always shown.

morinted:
Maybe we should prompt for the registry key change. Think of city workers who don't have user permissions to edit registry... what's the behavior when the setting fails?

morinted:
We could prompt to set the registry key, then after they dismiss it, store that in the config.

benoit-pierre:
This is getting rather complicated... How do we prompt the user from that part of the code?

morinted:
Not sure. As you can tell, I'm just wary over resetting registry keys in a portable program :/ Maybe the registry workaround could be moved into a check when opening config? Since it's not explicitly required for Plover to work (most users don't encounter this issue)

benoit-pierre:
I don't know, I think it's possible for any Windows user using the keyboard to encounter the bug under the "right" circumstances: the system load hit a spike and there goes the hook...

We either: don't modify the registry, and just had a note in the FAQ/documentation, or change the code to output a warning if the registry cannot be changed (but keep going with the machine initialization). In fact we should mention the problem somewhere in the documentation either way.

morinted:
I agree that we should mention it in docs, for sure. Maybe in the GitHub wiki (I want to port more information to it)

We could display a warning in the machine config if the current keyboard is set? Then have a button that says "Update Key" and notify to reboot?

benoit-pierre:
That would mean no notification on first run (since the keyboard is the default machine).

I tweaked the code so not being able to create/update the key does not prevent the machine from being initialized.

morinted:
Okay, I think it would be good to merge this, create a build, then let it
loose and see how it works for people.

benoit-pierre:
OK.

PR #462, opened 2016-04-03
linux: update setup documentation for Ubuntu

benoit-pierre:
Add missing `libudev-dev` dependency (required for compiling `hidapi` support).


benoit-pierre:
I should have picked it up sooner as it's in the list of dependencies we install for the Travis build...

PR #478, opened 2016-04-20
formatting: rework handling of actions 'replace'

### LINKED ISSUE #476: {MODE:SET_SPACE:} not compatible with Space Placement: After Output
Classification: Bug

Reproducibility: Always

## Summary

When {MODE:SET_SPACE:} is on and Space Placement is set to "After Output", any stroke that begins with {^ will delete the last character of the preceding stroke.

## Steps to Reproduce

1. Set Space Placement to "After Output" on the Output tab
2. Activate {MODE:SET_SPACE:}
3. Stroke TPAO/PWAR/P-P/TPHET/R-R

## Expected Results

1. Output should be "foobar.net" + newline

## Actual Results

3. Output is "fooba.ne" + newline; the "r" is deleted by the P-P stroke and the "t" by the R-R stroke.

## Version

Plover version 2.5.8+567.g52ebe0b

Installed via: git

## Notes

This problem does not occur when Space Placement: Before Output is selected.

## Configuration

OS: Windows 10 Home

benoit-pierre:
Don't honor `replace` if it does not match at the end of the previous
text. 

This is done by providing the previous translation formatting to the `OutputHelper` class so it can render the initial text (so the matching actually be done when the first undo/to translation action uses `replace`).

This fixes #476.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -95,7 +95,7 @@ class Formatter(object):
         rendered translations. If there is no context then this may be None.
 
         """
-        prev_formatting = prev.formatting if prev else None
+        prev_formatting = prev.formatting if prev else []
 
         for t in do:
             last_action = self._get_last_action(prev.formatting if prev else None)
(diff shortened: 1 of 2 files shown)


benoit-pierre:
I added some comments for the tests.

benoit-pierre:
OK, so output optimization is important, especially if we want to support things like chained meta-commands (see added tests).

benoit-pierre:
The more I think about the code, the more I find ways to trip it up... and holes in our testsuite!

Note: the last commit move the `undo/do` similar prefix optimization to the `OutputHelper`, where I think it belongs.

Issue #209, opened 2014-09-15
spacing bug when loading from RTF dictionary


balshetzer:
I think this is a bug. I've noticed it myself several times. I'll
define something like:

TPR*ED: {#Return}{#Return}{^^}FRED: {-|}

And for the duration of that Plover session, it'll work as specified,
with no space after the line breaks before the speaker ID. But then
when I close Plover and start it up again, suddenly it'll insert a
space at the beginning of the line, after the line breaks and before
the speaker ID. I've also seen problems with defining strokes like
{^-^} from within Plover. I think the code is doing something wonky
with spaces and delete-space commands when it saves the dictionary to
file, even though it's parsing it correctly immediately after the
definition is made. I'm not sure if this is true for rtf dictionaries
only (my bottom-most dictionary is rtf format) or if it's true of json
dictionaries as well, but it only seems to be true for definitions
made using the dictionary update window, as opposed to direct editing
with the file in a text editor.

morinted:
I think that I'd have to test this by creating an RTF dictionary and loading it up with commands.

PR #491, opened 2016-05-03
support embedded \n \r \t in translations

benoit-pierre:
It turns out that we already support a kind of undo-able return (at least on Windows): manually editing the JSON dictionary to change "{^}{#Return}{^}{-|}" to "{^\n^}{-|}" work. It's however not possible to add such an entry through Plover's GUI, and Linux and Mac OS X don't handle those embedded \n \r \t, which is what this PR is about.

Linux and OS X respective implementations have been fixed, and two helpers have been added to the `translation` module: `escape_translation` and `unescape_translation`:

* `escape_translation` is used when displaying a translation to the user (lookup / add translation dialogs...)
* `unescape_translation` is used on the user input (add translation dialog), to get the raw translation

Note:

1. the special case: `r'\\n' -> r'\n'`, so a literal "\n" can be entered (same for \t \r, but elsewhere `r'\\'` is left unchanged)
1. naively replacing "{^}{#Return}{^}{-|}" by "{^}\n{^}{-|}" will not work, as translations atoms are stripped
1. this does not change the format of the dictionaries in anyway
1. in the RTF/CRE dictionary, `r'\\\r|\\\n'` is replaced by `'{#Return}{#Return}'`, should this be replaced by `'{\n}'`? (Also why 2 returns?)
1. missing from this PR: I'm going to update the dictionary editor too

### DIFF
--- a/plover/dictionary_editor_store.py
+++ b/plover/dictionary_editor_store.py
@@ -1,5 +1,6 @@
 from plover.steno import normalize_steno
 from plover.gui.util import shorten_unicode
+from plover.translation import escape_translation, unescape_translation
 
 STROKE = "STROKE"
 TRANSLATION = "TRANSLATION"
@@ -58,7 +59,7 @@ class DictionaryEditorStore():
             for dk, translation in dictionary.iteritems():
                 joined = '/'.join(dk)
                 item = DictionaryItem(joined,
-                                      translation,
+                                      escape_translation(translation),
                                       dictionary,
                                       item_id)
                 self.all_keys.append(item)
@@ -145,13 +146,13 @@ class DictionaryEditorStore():
 
         # Creates
         for item in self.added_items:
-            item.dictionary[normalize_steno(item.stroke)] = item.translation
+            item.dictionary[normalize_steno(item.stroke)] = unescape_translation(item.translation)
             needs_saving.add(item.dictionary.get_path())
 
         # Updates
         for item_id in self.modified_items:
             item = self.all_keys[item_id]
-            item.dictionary[normalize_steno(item.stroke)] = item.translation
+            item.dictionary[normalize_steno(item.stroke)] = unescape_translation(item.translation)
             needs_saving.add(item.dictionary.get_path())
 
         # Deletes
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -3,8 +3,11 @@
 
 import wx
 from wx.lib.utils import AdjustRectToScreen
+
 from plover.steno import normalize_steno
 import plover.gui.util as util
+from plover.translation import escape_translation, unescape_translation
+
 
 TITLE = 'Plover: Add Translation'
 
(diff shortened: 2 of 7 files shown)


morinted:
This is great. Should we handle `\r\n` as just one carriage return, or is that getting a little specific?

benoit-pierre:
Since "\n" (and other control codes) are escaped in both JSON and RTF/CRE dictionaries, and the keyboard emulation layer does not distinguish between \n and \r, I don't think we should further complexify the escape/unescape code.

benoit-pierre:
I fixed the dictionary editor, so what's left is the question of the RTF/CRE conversion.

morinted:
As far as I can see, there's no escaped newline in RTF/CRE: http://www.legalxml.org/workgroups/substantive/transcripts/cre-spec.htm

If you want, you could just replace `{#Return}{#Return}` in the translation to `\n\n` and I'd be happy with that. As for going from `\n` to RTF, you could use `*/*`, which is a special paragraph token in RTF/CRE that some dictionaries use:

> \asterpar. Flag: \*/\* used for paragraph

http://www.legalxml.org/workgroups/substantive/transcripts/cre-spec.htm

benoit-pierre:
According to your link, text elements conform to the [RTF specification](https://www.microsoft.com/en-us/download/details.aspx?id=10725):

> A carriage return (character value 13) or line feed (character value 10) is treated as a \par control if the character is preceded by a backslash. You must include the backslash; otherwise, RTF ignores the control word. (You may also want to insert a carriage-return/line feed pair without backslashes at least every 255 characters for better text transmission over communication lines.)
A tab (character value 9) should be treated as a \tab control word. Not all RTF readers understand this; therefore, an RTF writer should always emit the control word for tabs.

So I think this code in `rtfcre_dict.py` is wrong:
```python
    def _re_handle_escaped_newline(self, m):
        r'\\\r|\\\n'
        return '{#Return}{#Return}'
```

And should at least be:
```python
    def _re_handle_escaped_newline(self, m):
        '\\\r|\\\n'
        return '{#Return}{#Return}'
```

I agree that we should replace `{#Return}` by `\n`, and for `\par` too.

It also looks like the code for saving does not format back to `\par` or equivalent...

benoit-pierre:
And we should convert:
* `\line`→ `\n`
* `\tab` → `\t`

stenoknight:
Will that work just as well for executing commands in non-text fields?

morinted:
Yes, exactly the same except the asterisk key will send a Backspace

benoit-pierre:
Plus `{#Return}` is still available.

Regarding RTF/CRE dictionaries, I think it's best to leave it untouched for now, as I'm afraid of round-trip errors (like #209), due to the incomplete save code. The first step should really be to add some tests.

morinted:
Will be merging, in the next dictionary update I might replace some of the default return strokes to use `\n` instead. Thanks!

PR #495, opened 2016-05-04
fix for PyPI release

benoit-pierre:
I had to change back the project name to lower case, to match the existing name used on PyPI.

Note to devs: make sure to delete `Plover.egg-info` from you tree to avoid duplicate distributions info.

### DIFF
--- a/osx/makefile
+++ b/osx/makefile
@@ -7,9 +7,9 @@ builddir = $(topdir)/build
 distdir = $(topdir)/dist
 
 VER := $(shell cd $(topdir) && $(PYTHON) setup.py --version)
-TMP_APP := $(distdir)/Plover.app
-APP := $(distdir)/Plover-$(VER).app
-DMG := $(distdir)/Plover-$(VER).dmg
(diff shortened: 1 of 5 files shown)


benoit-pierre:
I updated the classifiers. Might be time to change the "Development Status" from "4 - Beta" to "5 - Production/Stable", no?

morinted:
Yes, for sure :D

benoit-pierre:
Done, we're stable!

morinted:
The readme could really use a refresh sometime, though we may need to update the Italian translation at the same time.

> Plover is not mature software.

benoit-pierre:
> The readme could really use a refresh sometime, though we may need to update the Italian translation at the same time.

Spanish!

> Plover is not mature software. Once running, it will interpret all keystrokes...

How about changing it too:

> Once Plover is running, it will interpret all keystrokes...

?

morinted:
That sounds like a good change. Maybe it would be good to minimize the read me entirely and link out to external sources like the wiki and the homepage instead, and then link to the OS-specific documentation.

PR #497, opened 2016-05-04
minor testsuite related changes

benoit-pierre:
* orthography: minor test code simplification
* passport: simplify test code
* formatting: tweak tests; better assert message


benoit-pierre:
* dictionary/rtfcre: minor test code simplification
* machine/treal: remove (non-functional) test code

PR #500, opened 2016-05-04
oslayer/winkeyboardcontrol: don't use pyHook anymore

### LINKED ISSUE #477: Windows: ALT codes do not work when Plover is open (even when disabled)
Classification: Bug

Reproducibility: Always

## Summary

[Alt Codes](https://en.wikipedia.org/wiki/Alt_code) do nothing when Plover is turned on. This is true even when Plover is disabled. A relatively minor issue that mostly only affects new steno users who, for whatever reason, use Windows, need to type non-ASCII characters sometimes, and use Alt codes to do so. (In other words, it might just affect me.)

## Steps to Reproduce

1. Open Plover; leave it disabled.
1. Hold ALT, hit "0233" on number pad

## Expected Results

2. é

## Actual Results

3. [no output]

## Version

Plover version 2.5.8+567.g52ebe0b

Installed via: git

## Notes

Workaround is simple; either close Plover or use another method to insert non-ASCII characters.

## Configuration

OS: Windows 10 Home

benoit-pierre:
This was originally done to make Python 3 support easier (no official Python 3 builds for PyHook), but turn out it also fixes #477, so...

Tested with a Windows 10 VM.

### DIFF
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -15,7 +15,6 @@ emulate keyboard input.
 
 import collections
 import ctypes
-import functools
 import multiprocessing
 import os
 import threading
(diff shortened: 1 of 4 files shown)


morinted:
It's crazy how little code this is and it removes a dependency. I'll play with it soon.

morinted:
Conflicts now because of the pip pyHook rename.

benoit-pierre:
Fixed.

PR #503, opened 2016-05-05
Extend implicit number hyphen support to right hand only

### LINKED ISSUE #490: Normalize steno should handle right hand numbers (6789)
Missing implicit hyphen, meaning that using the add translation dialogue will fail if you use one of the right hand numbers alone.

morinted:
Fix #490, essentially the user can now define a stroke with 6-9 in it as strokes and normalize steno will interpret the stroke as -6 through -9 as is proper.


benoit-pierre:
This does not pass the testsuite (see #506).

morinted:
Thanks, I've updated.

PR #505, opened 2016-05-05
Add display configuration to start Plover minimized

### LINKED ISSUE #475: Start-Minimized Option
Classification: Feature (New)

Reproducibility: N/A

## Summary

When adding a translation, the Plover window comes to front unless minimized. I often forget to minimize manually on startup, so having an option to always start minimized would be extremely useful.

## Steps to Reproduce

I check the "start minimized" box in config.

## Expected Results

Plover starts minimized.

## Actual Results

Plover has to be manually minimized to keep the main window to pop up when adding translations.

If helpful, submit screenshots of the issue to help debug. Your `plover.log` or `strokes.log` may also be useful.

## Version
2.5.8

## Notes
Workaround: Minimize Plover before adding any translations.

## Configuration
???

morinted:
Fix #475, add config to start Plover minimized.

Screenshot of the configuration option to start Plover minimized

### DIFF
--- a/plover/gui/main.py
+++ b/plover/gui/main.py
@@ -54,7 +54,13 @@ class PloverGUI(wx.App):
         from plover.gui import log as gui_log
         frame = MainFrame(self.config)
         self.SetTopWindow(frame)
+
+        osx = sys.platform.startswith('darwin')
+        if not osx:
+            frame.Iconize(self.config.get_start_minimized())
         frame.Show()
+        if osx:
+            frame.Iconize(self.config.get_start_minimized())
         return True
 
 
--- a/plover/config.py
+++ b/plover/config.py
@@ -40,6 +40,10 @@ DEFAULT_ENABLE_STROKE_LOGGING = False
 ENABLE_TRANSLATION_LOGGING_OPTION = 'enable_translation_logging'
 DEFAULT_ENABLE_TRANSLATION_LOGGING = False
 
+STARTUP_SECTION = 'Startup'
+START_MINIMIZED_OPTION = 'Start Minimized'
+DEFAULT_START_MINIMIZED = False
+
 STROKE_DISPLAY_SECTION = 'Stroke Display'
 STROKE_DISPLAY_SHOW_OPTION = 'show'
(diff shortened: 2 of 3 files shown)


morinted:
I've only tried this on my KDE install, I'll do OSX tomorrow. @stenoknight You can try with the AppVeyor artifact: https://ci.appveyor.com/api/buildjobs/pdrikwlin253d2ms/artifacts/dist/plover-3.0.0+12.ga45aad5.exe

stenoknight:
X'D

plover/gui/main.py
```diff
@@ -51,9 +51,9 @@ def __init__(self, config):
     def OnInit(self):
         """Called just before the application starts."""
         # Enable GUI logging.
-        from plover.gui import log as gui_log
```
benoit-pierre: Mistake?

benoit-pierre:
Works fine on Ubuntu and Arch Linux with Awesome. Idem on Windows 10.

plover/gui/main.py
morinted: Ah, it didn't look used, but does this import actually do something?

benoit-pierre: It initializes the GUI log: adding the platform specific log handler.

morinted:
On OS X, frame.Show() starts raises the window…

I tried removing set top app, nada. If I place an iconize event after the show, the Plover window raises then minimizes. If I set iconized false, then show, then set iconized false again…Plover starts in the dock, raises to display itself in its glory, and then minimizes again (which is very annoying on Mac with the genie lamp animation).

If I don't call show, then Plover is in the dock, but you can see that its mainframe isn't displayed…

morinted:
@benoit-pierre The best I can get on OSX is to show the window then immediately iconize it. The same code looks silly on Winux because you see a flash of the window and lose focus for a sliver of time.

The solution is as committed... maybe you could suggest something a little more elegant?

```python
        osx = sys.platform.startswith('darwin')
        if not osx:
            frame.Iconize(self.config.get_start_minimized())
        frame.Show()
        if osx:
            frame.Iconize(self.config.get_start_minimized())
```

PR #512, opened 2016-05-08
OS X: Avoid timeout disabling key capture event tap [fixes #484]

### LINKED ISSUE #484: OS X: Plover stops translating input till restart
Classification: Bug

Reproducibility: Eventually Always
## Summary

Plover eventually stops translating input from my keyboard. I switch it off with `{PLOVER:TOGGLE}`, and then I go to hit `{PLOVER:TOGGLE}` to turn it back on, and nothing – I just see the normal key output, rather than Plover re-enabling itself.
## Steps to Reproduce
1. Fire up Plover.
2. Leave it off most of the time, but turn it on for spurts, then back off.
3. Turn output translation back on a while later.

Side note: I was able to repro more under my control (vs the random drop-outs I seem to see most of the time) by backgrounding the process with C-z and then foregrounding with fg. I noticed it stopped receiving strokes after foregrounding. Fixing it to be "background-safe" might fix this problem in general.

Random reaching not based on any actual examination of the CGEventTap stuff: Perhaps it has to do with taking too long to handle some input and the OS just kicking it out?
## Expected Results

It turns back on.
## Actual Results

Nada.
## Version

I've been running HEAD (it's stable enough). Currently sitting on 403162ddcd032f842faee29387d7c7e48c90d7e4, which is reporting 2.5.8 in the About page. This has been happening for a while though.

Installed via GitHub clone, run via `sudo python setup.py launch`.
## Notes

I haven't found any particular trigger for this behavior. Sometimes I'll flip it on and off a few times in succession, go to flip it back on without much time with it off, and it just won't come back up. :(

Could there be a relationship to #444 and the out-of-order behavior noted in some of the later comments?
### Workaround

Quit and relaunch. This fixes it every time.
## Configuration

OS:
ProductName:    Mac OS X
ProductVersion: 10.11.4
BuildVersion:   15E65

jeremy-w:
### About
- Reformats `plover/oslayer/osxkeyboardcontrol.py` to make `flake8` happy. It found a couple unused imports and several too-long lines.
- Re-enables the event tap whenever it gets disabled by timeout. Turns out the system notifies you of this via one of those "meta events" we were ignoring.
- Tries to avoid timing out by shifting all the steno engine work outside the event tap callback.
   - Key events get sent over a `Queue.Queue` to a worker thread, where they get sent to `key_up` and `key_down`.
- Also tidies up some resource management. We were setting up some stuff in `run()` that wasn't getting torn down in `cancel()`.

NOTE: There's another meta-event to tell us the user intentionally shut off our tap. I'm guessing this gets sent if the user revokes our assistive device permissions. I wonder if we should self-cancel and raise an exception should that happen?

### Issues
Fixes #484.

### Release Notes
- OS X: Keyboard key translation no longer drops out. Previously, Plover would intermittently stop receiving keystrokes until it was relaunched. (#484)

### Tested Platforms
OS X.

This only affects OS X, so I think we're good to go.


jeremy-w:
@morinted Could you beat up on this a bit on OS X? Everything looks fine to me – I disabled and enabled repeatedly and poked around. Even with a 1-second sleep in the tap callback under the keyboard control script, the re-enabling approach was able to recover from a timeout-disable, so I think this issue is thoroughly squashed.

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -232,6 +234,13 @@ def callback(proxy, event_type, event, reference):
             # Don't pass on meta events meant for this event tap.
             is_unexpected_event = event_type not in self._KEYBOARD_EVENTS
             if is_unexpected_event:
+                if event_type == kCGEventTapDisabledByTimeout:
+                    # Re-enable the tap and hope we act faster next time
+                    CGEventTapEnable(self._tap, True)
+                    plover.log.warning(
+                        "Keyboard event tap was disabled by timeout. " +
```
morinted: OS X notifications have a pretty short character limit, does this fit?

morinted:
It looks pretty good, thanks. I'll have to try it out and play with it a little. Nice commenting as well.

plover/oslayer/osxkeyboardcontrol.py
jeremy-w: Good question. I actually wanted it just to go to console logs, not notifications. And since we seem to recover fine now, perhaps only at INFO or DEBUG level.

What would be the right tool for that in the Plover codebase?

jeremy-w: OK, I ended up triggering this today. It worked out OK, but I might want a newline after the "disabled by timeout" bit.

jeremy-w:
Here's what the warning alert looks like:

# Notification Bubble
This truncates, because the width of the bubble is limited and only allows one line.

notification bubble - captura de pantalla 2016-05-09 a las 12 46 18

# In Notification Center
This wraps, so the full message is visible. That means someone could pop over and see the whole thing after the transient message disappears.

in notification center - captura de pantalla 2016-05-09 a las 12 46 31

# Why do we still tap when disabled?
One thing seems weird: I've got output translation disabled, but Plover is still tapping my key events. I guess that's so it can pick up the "enable" command?

When I saw the "tap disabled" bit, it corresponded with a lag in key entry, and I had to retype some characters. I'm not sure why it got hung up with the new queue architecture and everything. :(

Guess I need to load-test the key handling somehow and profile it to see what makes it lag? The re-enabling worked though, so it was just a small hiccup, rather than a complete fall-out.

benoit-pierre:
> One thing seems weird: I've got output translation disabled, but Plover is still tapping my key events. I guess that's so it can pick up the "enable" command?

Yes.

> When I saw the "tap disabled" bit, it corresponded with a lag in key entry, and I had to retype some characters. I'm not sure why it got hung up with the new queue architecture and everything. :(

Because of the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock)? You could move it to a subprocess, like it's now done with the Window code, but you can still never guarantee  that tap events will be processed on time if the machine is under heavy load.

Do we know what's the limit (timeout)? Is it configurable?

jeremy-w:
I don’t think the timeout is documented or configurable. Putting it in a debugger and looking up-stack during the callback might quickly scare up wherever the timeout timer gets set up, but if we can’t configure it, I don’t think the knowledge is going to help us much.

I guess the only way to get reliable tap message processing would be to push it out of process to a non-Python process. :\ Which sounds like a build-time headache, since you’d have to wrangle another build system on top of the Python one, and package and locate that binary appropriately.
--
Jeremy W. Sherman
https://jeremywsherman.com/ <https://jeremywsherman.com/>

benoit-pierre:
> I guess the only way to get reliable tap message processing would be to push it out of process to a non-Python process. :\ Which sounds like a build-time headache, since you’d have to wrangle another build system on top of the Python one, and package and locate that binary appropriately.

No, look at the Windows code, it's using the [multiprocessing module](https://docs.python.org/2.7/library/multiprocessing.html).

jeremy-w:
Ah, I understood it as, if the GIL is the problem, then the only way to escape the GIL is to switch to handling the callbacks in a language without one. But you think that just having a dedicated process for the event handling would suffice?
--
Jeremy W. Sherman
https://jeremywsherman.com/ <https://jeremywsherman.com/>

benoit-pierre:
It think it would help. Suffice? Again I don't think you can guarantee the taps events will be processed fast enough. It did help on Windows, but changing the timeout (through a registry key) was really the only way to make it disappears (well, assuming we always take less than 5 seconds to process a key event :P).

jeremy-w:
The threading approach has been working well for me for a few days now.

I'll shorten the warning message that shows in the notification. The re-enabling is working now, so we don't need that "if this doesn't work" bit. At that point, I think this is good to merge.

jeremy-w:
Warning has been reworked to be more user-relevant. This also hasn't dropped out on me since that one time a few days ago, after almost a work week of leaving it running. :shipit:

morinted:
Great! We'll include it in the next weekly and see what people think. I've found it hard to test because I rarely use a keyboard as a machine. I tried it and it works, but I can't account for its stability.

PR #513, opened 2016-05-08
steno: optimize normalize_steno

benoit-pierre:
Test code:

```python
import timeit

from plover.dictionary.base import load_dictionary

def instrumented():
    load_dictionary('main.json')

# Note: timeit disables garbage collection, don't set number to high!
print '%.3fs' % timeit.timeit(instrumented, number=20)
```

On my machine (Intel(R) Core(TM) i7-3770 CPU @ 3.40GHz), the results are:
* master: 44.549s
* this branch: 22.421s
* with `normalize_steno = lambda s: tuple(s.split('/'))`: 10.997s

### DIFF
--- a/test/test_steno.py
+++ b/test/test_steno.py
@@ -11,33 +11,47 @@ from plover.steno import normalize_steno, Stroke
 class StenoTestCase(unittest.TestCase):
     def test_normalize_steno(self):
         cases = (
-        
-        # TODO: More cases
-        ('S', 'S'),
-        ('S-', 'S'),
-        ('-S', '-S'),
-        ('ES', 'ES'),
-        ('-ES', 'ES'),
-        ('TW-EPBL', 'TWEPBL'),
-        ('TWEPBL', 'TWEPBL'),
-        ('19', '1-9'),
-        ('14', '14'),
-        ('146', '14-6'),
-        ('67', '-67'),
-        ('6', '-6'),
-        ('9', '-9'),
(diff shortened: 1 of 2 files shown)


test/test_steno.py
```diff
@@ -11,33 +11,47 @@
...
+            ('9', '-9'),
+            ('5', '5'),
+            ('0', '0'),
+            ('456', '456'),
+            ('46', '4-6'),
+            ('4*6', '4*6'),
+            ('456', '456'),
+            ('S46', 'S4-6'),
+            # Number key.
+            ('#S', '#S'),
```
morinted: I guess this may have never been the case, but shouldn't \#S normalize to `1` ?

benoit-pierre: Good question, I did not want to change the behavior of `normalize_steno`, so I updated the tests before optimizing it. What if the stroke contains non-numbers?

morinted: I'd say, if you have numbers, no number symbol. If you don't, you need it.

`#K`: `#K`
`#S`: `1`
`#SK`: `1K`
`STKP#`: `12K3`
`K#`: `#K`

That being said, we could merge this as it's not a regression. If you want to continue down that path, you are welcome to. Let me know :grin:

benoit-pierre: Yes, I think we should merge this as is, and if a consensus is reached on another normalized format it can be addressed in a new PR.

PR #517, opened 2016-05-12
Allow non-ASCII non-control characters when saving a JSON dictionary

benoit-pierre:
This make it easier to externally edit such dictionaries.

### DIFF
--- a/plover/dictionary/json_dict.py
+++ b/plover/dictionary/json_dict.py
@@ -34,7 +34,6 @@ def load_dictionary(filename):
                            for x in dict(d).iteritems())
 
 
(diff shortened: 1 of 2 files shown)


plover/dictionary/json_dict.py
```diff
@@ -36,5 +36,7 @@ def load_dictionary(filename):
 
 # TODO: test this
```
jeremy-w: Looks like you are testing this, so the TODO can be removed?

benoit-pierre: Indeed! Amended, thanks.

Issue #483, opened 2016-04-25
Linux: "unsupported key name:"


morinted:
## Summary

When using a nested command, e.g. `{#Control_L(Shift_L(c))}`, one gets an "unsupported key name:" notification, with no content. It's because we call `_get_maping_from_keystring` (should be "mapping") after a close-parenthesis, even when there is no content in the "current_command". So, we should either not act if the `current_command` is empty, or should check if the keystring is empty in `_get_maping_from_keystring`

PR #519, opened 2016-05-12
Rework handling of key combos

benoit-pierre:
Factorize and simplify the handling of key combos. The syntax remains the same, but there are a few differences in the way the new implementation works:

* key names are now case insensitive, so `#{a}` is equivalent to `#{A}`, `#{Super_l(return)}` to `#{super_l(Return)}`, ...

* the goal of key combos is to support shortcuts, and key name are used instead of key codes because they are more user-friendly, as a way to indicate which physical key must be pressed / hold, so for example when using Qwerty `#{1}` and `#{exclam}` are equivalent, since both names refer to the same key (ditto for `#{Shift_L(1)}` and `#{Shift_L(exclam)}`

* a key press / hold event is not allowed if the key is already in the pressed state, so both `#{1(exclam)}` and `#{exclam(1)}` are invalid

* the whole combo string is validated (for syntax errors, invalid key names, invalid key press / hold events) before sending the resulting events

* alias for the modifiers have been added; of the form `mod` for `mod_l`, for `alt` `control` `shift` and `super`

* on OSX, `command` is an alias for `super` (`super_l`), and `option` for `alt` (`alt_l`)

* on Windows, `windows` is an alias for `super` (`super_l`)

Tested on Arch Linux and with a Windows 10 VM.

TODO:
* cleanup the Windows / OSX code (redundant tables...)
* the error messages on invalid combo are nice when a monospaced font is used, but hard to read when that's not the case, suggestions for an alternative format welcomed

### DIFF
--- a/plover/key_combo.py
+++ b/plover/key_combo.py
@@ -1,0 +1,88 @@
+
+import re
+
+
+_SPLIT_RX = re.compile(ur'(\s+|(?:\w+(?:\s*\()?)|.)')
+
+def parse_key_combo(combo_string, key_name_to_key_code=None):
+
+    if key_name_to_key_code is None:
+        key_name_to_key_code = lambda key_name: key_name
+
+    key_events = []
+    down_keys = []
+    token = None
+    count = 0
+
+    def _raise_error(exception, details):
+        msg = u'%s in "%s"' % (
+            details,
+            combo_string[:count] +
+            u'[' + token + u']' +
+            combo_string[count+len(token):],
+        )
+        raise exception(msg)
+
+    for token in _SPLIT_RX.split(combo_string):
+        if not token:
+            continue
+
+        if token.isspace():
+            pass
+
+        elif re.match(ur'\w', token):
+
+            if token.endswith(u'('):
+                key_name = token[:-1].rstrip().lower()
+                release = False
+            else:
+                key_name = token.lower()
+                release = True
+
+            key_code = key_name_to_key_code(key_name)
+            if key_code is None:
+                _raise_error(ValueError, 'unknown key')
+            elif key_code in down_keys:
+                _raise_error(ValueError, u'key "%s" already pressed' % key_name)
+
+            key_events.append((key_code, True))
+
+            if release:
+                key_events.append((key_code, False))
+            else:
+                down_keys.append(key_code)
+
+        elif token == u')':
+            if not down_keys:
+                _raise_error(SyntaxError, u'unbalanced ")"')
+            key_code = down_keys.pop()
+            key_events.append((key_code, False))
+
+        else:
+            _raise_error(SyntaxError, u'invalid character "%s"' % token)
+
+        count += len(token)
+
+    if down_keys:
+        _raise_error(SyntaxError, u'unbalanced "("')
+
+    return key_events
+
+
+def add_modifiers_aliases(dictionary):
+    ''' Add aliases for common modifiers to a dictionary of key name to key code.
+
+    - add `mod` for `mod_l` aliases for `alt`, `control`, `shift` and `super`
+    - add `command` and `windows` aliases for `super`
+    - add `option` alias for `alt`
+    '''
+    for name, extra_aliases in (
+        ('control', ''               ),
+        ('shift'  , ''               ),
+        ('super'  , 'command windows'),
+        ('alt'    , 'option'         ,)
+    ):
+        code = dictionary[name + '_l']
+        dictionary[name] = code
+        for alias in extra_aliases.split():
+            dictionary[alias] = code
--- a/plover/key_combo.py
+++ b/plover/key_combo.py
@@ -1,5 +1,138 @@
+# -*- coding: utf-8 -*-
 
 import re
+
+
+# Mapping of "standard" keynames (derived from X11 keysym names) to Unicode.
+KEYNAME_TO_CHAR = {
+    # Generated using:
+    #
+    # from Xlib import XK
+    # from plover.oslayer.xkeyboardcontrol import keysym_to_string
+    # for kn, ks in sorted({
+    #     name[3:].lower(): getattr(XK, name)
+    #     for name in sorted(dir(XK))
+    #     if name.startswith('XK_')
+    # }.items()):
+    #     us = keysym_to_string(ks)
+    #     if us == kn or not us:
+    #         continue
+    # print '    %-20r: %8r, # %s' % (kn, us, us)
+    'aacute'            :  u'\xe1', # á
+    'acircumflex'       :  u'\xe2', # â
+    'acute'             :  u'\xb4', # ´
+    'adiaeresis'        :  u'\xe4', # ä
+    'ae'                :  u'\xe6', # æ
+    'agrave'            :  u'\xe0', # à
+    'ampersand'         :     u'&', # &
+    'apostrophe'        :     u"'", # '
+    'aring'             :  u'\xe5', # å
+    'asciicircum'       :     u'^', # ^
+    'asciitilde'        :     u'~', # ~
+    'asterisk'          :     u'*', # *
+    'at'                :     u'@', # @
+    'atilde'            :  u'\xe3', # ã
+    'backslash'         :    u'\\', # \
+    'bar'               :     u'|', # |
+    'braceleft'         :     u'{', # {
+    'braceright'        :     u'}', # }
+    'bracketleft'       :     u'[', # [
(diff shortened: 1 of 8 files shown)


benoit-pierre:
Forgot to mention:

1. this fixes #483 
1. I found #520 when implementing, this PR does not address it

jeremy-w:
Should we make it so all the aliases can be resolved across all platforms? Otherwise we can get in the case where dictionaries in stenodict work on a machine with one OS but not another, simply because of "dialect" differences.

Though, having Cmd decay to Ctrl on a non-OS X platform would allow for cross-platform shortcuts in a lot of cases (Cmd-s and Ctrl+s both save, for example).

plover/key_combo.py
```diff
@@ -0,0 +1,68 @@
...
+
+    key_events = []
+    down_keys = []
+    token = None
+    count = 0
+
+    def _raise_error(exception, details):
+        msg = '%s:\n%s\n%s' % (
+            details, combo_string,
+            ' ' * count + '^' * len(token or ' '),
```
jeremy-w: I think underlining with `^^^^` is clear, but my experience with clang and similar leads me to expect the first character to be `^` and the rest to be `~`, so `^~~~` to point at the initial character and then underline a source range.

As you pointed out, neither works well if the output is set in a proportional font. To accommodate that, we'd probably need to do something like boxing the bogus token with `[…]` or linebreaking around the bogus token.

If we could trace the bogus key combo back to a specific dictionary file and line, that would also be font-independent, and we could include that info in the error message, too.

morinted:
For the token, the Scala test suite always uses square brackets around the problem token, but it's usually shown next to the expected output. Just out of curiosity:

> Unexpected ")" in "Super(l)[)]"

As for the resolving, I think it would be most useful to have super = Windows = command, and also useful to have Alt = option.

It would be theoretically nice to make command resolve to control on Windows (and vice versa?), but I can think of too many edge cases for it to be reasonable. That would perhaps be better suited to a plugin.

benoit-pierre:
OK, so now "Windows" and "Command" are always aliases for "Super" and, "Option" for "Alt". Note that this means that "XF86_Option" must be now explicitly used on Linux.

For the exceptions format, it's now:

```
SyntaxError: invalid character '+' in 'foo () bar ( foo[+] bar(b) c d)'
```

or:

```
ValueError: key 'bar' already pressed in 'foo () bar ( foo [BaR(]b) c d)'
```

morinted:
Looking good. Just a note on code style: I notice that you're using "NAME2CODE" and similar, currently variables are styled NAME_TO_CODE, would it make sense to keep the style?

Finally, would it be worth pulling out the modifier aliases somewhere because they are all the same?

benoit-pierre:
> Looking good.

But does it work (on OSX)?

> Just a note on code style: I notice that you're using "NAME2CODE" and similar, currently variables are styled NAME_TO_CODE, would it make sense to keep the style?

Sure.

> Finally, would it be worth pulling out the modifier aliases somewhere because they are all the same?

Will do.

morinted:
Actually, it doesn't work on OS X, because KEYNAME_TO_KEYCODE doesn't have most characters. You'll want to use `mac_keycode.KeyCodeForChar(c)` and just get the base key, ignoring the modifiers. The only problem is that this will pick up "!" as "1" and "%" as "5" and so on.

benoit-pierre:
@morinted: can you try with this code (near line 494):
```python
        def name2code(name):
            code = KEY_NAME2CODE.get(name)
            if code is not None:
                return code
            char = KEYNAME_TO_CHAR.get(name, name)
            code, mods = mac_keycode.KeyCodeForChar(char)[0]
            return code
```

morinted:
Yep, that seems to be working for me.

benoit-pierre:
OK, thanks. I just cleaned up the OSX code, please let me know if it still works :P.

benoit-pierre:
The Windows code has now been cleaned-up too.

plover/oslayer/winkeyboardcontrol.py
```diff
@@ -117,203 +118,150 @@ class WindowsKeyboardLayout:
...
+        'udiaeresis': 220, 'eacute': 233, 'eth': 240, 'eth': 240, 'iacute': 205,
+        'onesuperior': 185, 'ocircumflex': 212, 'odiaeresis': 214, 'oacute': 243
     }
 
     # Maps from literal characters to their key names.
-    literals = collections.defaultdict(str, {
+    LITERALS = {
         '~': 'asciitilde', '!': 'exclam', '@': 'at',
-        '#': 'numbersign', '$': 'dollar', '%': 'percent',
+        '#': 'numbersign', '$': 'dollar', '%': 'percent', '^': 'asciicircum',
```
morinted: I had an issue when writing the code originally where "^" send the Alt-Gr version instead of the Shift-6 version on a certain key layout, making stand alone "^"s impossible. Can you confirm that you can use the symbol without issue (curious about QWERTY, AZERTY, and the British layout)

benoit-pierre: If something needs to be commented, it's really this kind of gotcha... And yes, "^" gets treated like a dead-key when using Azerty, which I guess is why "`" (grave) is missing too... But really, the problem is with any key that act as a dead-key, e.g. "~" (tilde) with "~n" resulting in "ñ"... Clearly [VkKeyScanEx sucks](http://archives.miloush.net/michkap/archive/2006/09/10/748742.html)

morinted: Yes, shame on future me for not leaving a comment, I have no idea what I was thinking. VkKeyScanEx is a huge bummer.

plover/oslayer/winkeyboardcontrol.py
```diff
@@ -117,203 +118,150 @@ class WindowsKeyboardLayout:
...
-        'onehalf': 189, 'Thorn': 222, 'aacute': 225, 'icircumflex': 238,
-        'Udiaeresis': 220, 'eacute': 233, 'Eth': 240, 'eth': 240, 'Iacute': 205,
-        'onesuperior': 185, 'Ocircumflex': 212, 'Odiaeresis': 214, 'oacute': 243
+        'currency': 164, 'ugrave': 249, 'ucircumflex': 219, 'odiaeresis': 246,
+        'periodcentered': 183, 'uacute': 218, 'idiaeresis': 239, 'yacute': 253,
+        'sterling': 163, 'ae': 198, 'ediaeresis': 203, 'onequarter': 188,
+        'onehalf': 189, 'thorn': 222, 'aacute': 225, 'icircumflex': 238,
+        'udiaeresis': 220, 'eacute': 233, 'eth': 240, 'eth': 240, 'iacute': 205,
+        'onesuperior': 185, 'ocircumflex': 212, 'odiaeresis': 214, 'oacute': 243
     }
```
benoit-pierre: And this needs fixing too:
- remove duplicate, now that we only handle lowercase key names
- I noticed the code for `acute` is not valid (it's in fact the code for `iacute`)

benoit-pierre:
OK, so I've completely reworked the Windows keyboard layout handling:

Instead of trying to find what keys a character is mapped to, do it the other way: find the character mapped to each (key + modifiers) combo.  This include finding out those combinations which produce dead letters, to prevent accidental character combining when simulating output.

I've tested with a number of different keymaps, and it seems to be working fine: "~n" no longer result in "ñ" when using Azerty. Note that with Azerty, "{#assciicircum}" will now result in the "9/ç/^" key being pressed, pressing the "¨^" key can be done with "{#dead_circumflex}" or "{#dead_diaeresis}".

I also found a bug (which was already present); I have this entry in my test dictionary: `"-B": "{^\n^}¨¨¨¨¨¨¨¨{^\n^}12345678"`, and I noticed that sometimes not all the "¨" are rendered... A problem with repeated letters?

morinted:
You are incredible. I'll run with this today, is the diaersis a key that goes through Unicode instead of the key code channels? Because I don't get the same issue with repeated strings of aaaaaaaa or """""""""".

I'll be using this today, then it's time to merge (you left in-progress, but I'm assuming you're done?)

I think that after this we need a dictionary page in the GitHub wiki to talk over some of the subtlities in the key combo logic, like the dead keys you listed.

benoit-pierre:
> I'll run with this today, is the diaersis a key that goes through Unicode instead of the key code channels? Because I don't get the same issue with repeated strings of aaaaaaaa or """""""""".

Correct, and I can only reproduce it in some applications: Firefox, VIM, but not Notepad or Libreoffice Writer...

PR #522, opened 2016-05-18
gui/add_translation: fix strokes filtering

benoit-pierre:
On Linux, the engine output triggers transient focus out/in events which, coupled with the fact that enabling/disabling the strokes filter is done in another thread in parallel to the machine thread, make it really unreliable.

So instead of relying on focus out events, only use focus in events to track which field is focused.

Note: that with the previous code it was possible (with the add translation dialog open) to switch focus to another window and write normally. This is not supported anymore, as we don't know the strokes/translation field lost focus, and using `EVT_KILL_FOCUS` at the dialog/window level does not work (it's never triggered, on Linux or Windows).

Tested on Linux and with a Windows 10 VM.

### DIFF
--- a/plover/gui/add_translation.py
+++ b/plover/gui/add_translation.py
@@ -83,21 +83,14 @@ class AddTranslationDialog(wx.Dialog):
         
         # events
         button.Bind(wx.EVT_BUTTON, self.on_add_translation)
-        # The reason for the focus event here is to skip focus on tab traversal
-        # of the buttons. But it seems that on windows this prevents the button
-        # from being pressed. Leave this commented out until that problem is
-        # resolved.
-        #button.Bind(wx.EVT_SET_FOCUS, self.on_button_gained_focus)
         cancel.Bind(wx.EVT_BUTTON, self.on_close)
-        #cancel.Bind(wx.EVT_SET_FOCUS, self.on_button_gained_focus)
         self.strokes_text.Bind(wx.EVT_TEXT, self.on_strokes_change)
         self.translation_text.Bind(wx.EVT_TEXT, self.on_translation_change)
         self.strokes_text.Bind(wx.EVT_SET_FOCUS, self.on_strokes_gained_focus)
-        self.strokes_text.Bind(wx.EVT_KILL_FOCUS, self.on_strokes_lost_focus)
         self.strokes_text.Bind(wx.EVT_TEXT_ENTER, self.on_add_translation)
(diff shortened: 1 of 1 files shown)


morinted:
I always found this feature to be okay on Windows and OSX... Is it possible to leave it intact for those OS's if Linux is the issue here?

benoit-pierre:
I really don't want to have different code paths for Linux and Windows / OSX... So I took another look at this: please check if this work on OSX (and I would not mind another Windows environment test). Hopefully it works on OSX too, in which case I'll rewrite before merge.

morinted:
Works great on OSX, only issue I see is the console reporting the following when the window is exiting (whether triggered through the escape key or through clicking the close/cancel button:

```
  File "/Users/ted/git/plover/plover/gui/add_translation.py", line 143, in on_close
    self.Update()
  File "/usr/local/lib/wxPython-3.0.2.0/lib/python2.7/site-packages/wx-3.0-osx_cocoa/wx/_core.py", line 16711, in __getattr__
    raise PyDeadObjectError(self.attrStr % self._name)
wx._core.PyDeadObjectError: The C++ part of the AddTranslationDialog object has been deleted, attribute access no longer allowed.
```

benoit-pierre:
I don't think this is a regression, as this code was already present before. Really, calling `self.Update` after `self.Destroy` does not seem normal...

morinted:
Maybe we can safely remove the line?

morinted:
After removing it and exiting, it's now:

```
  File "/Users/ted/git/plover/plover/gui/add_translation.py", line 226, in on_activate
    elif self._win_id != focus:
  File "/usr/local/lib/wxPython-3.0.2.0/lib/python2.7/site-packages/wx-3.0-osx_cocoa/wx/_core.py", line 16711, in __getattr__
    raise PyDeadObjectError(self.attrStr % self._name)
wx._core.PyDeadObjectError: The C++ part of the AddTranslationDialog object has been deleted, attribute access no longer allowed.
```

benoit-pierre:
So you removed the `Update` call, and kept `Destroy`? Does it work if you add:
```python
self.Disconnect(-1, -1, wx.wxEVT_ACTIVATE)
```
at the start of `on_close`?

morinted:
Yep, that works.

benoit-pierre:
OK, please double check the commit I just pushed contains the same changes you tested, and I'll rewrite so we can merge.

morinted:
It still works 👍

benoit-pierre:
OK, ready for merge then!

PR #524, opened 2016-05-21
travis: fix possible OSX build failure

benoit-pierre:
Building pyobjc-framework-Cocoa fails if Cython is installed, see: https://github.com/pypa/setuptools/issues/488

This is fixed by upgrading setuptools to >=19.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,10 +36,12 @@ before_install:
       ;;
     esac &&
     ./setup.py write_requirements &&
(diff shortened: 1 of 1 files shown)


morinted:
Do you have a sample `.app` that I can try to verify it works?

benoit-pierre:
No, but note that after installing all the dependencies setuptools is downgraded back to 19.2, so the py2app generation should work as before.

Issue #258, opened 2015-09-25
Strokes being held back in OSX version (Transcriptor X)


stanographer:
Only in the OSX version, up to the last three strokes will be held back for an indeterminate amount of time (usually ranging from a fraction of a second to several seconds) before Plover receives them and translates them. Writing more strokes "pushes" the held-back strokes, however. Although I have a suspicion that this is due to BT going into power saving mode.  When the lag starts, I typically restart Plover and it will momentarily stop but usually comes back after a few minutes of writing.

All types of strokes are equally affected. Command strokes lag just as much as text-only ones. I haven't observed this lag in the Windows version.

Achim63:
I can reproduce this. But after a restart it mostly works fine for hours until this bug kicks in again.

morinted:
This sounds similar to #189. Can you try Plover version 2.5.7 to see if the problem persists there? https://github.com/openstenoproject/plover/releases/download/v2.5.7/Plover.dmg

Achim63:
I'll try to do this sometime this weekend. Might well be the same as #189 since I also use the TX Bolt protocol.

stanographer:
I will try using 2.5.7 today to see if the same problem keeps happening.

morinted:
Update: stroke display *does not* update when the slow down occurs, demonstrating that this is a problem with the reading of the machine itself on OSX.

PR #530, opened 2016-05-28
Adjust TX Bolt protocol bytes

stanographer:
Found that the TX Bolt bytes sent from the machine are actually 0-3, not 6 as the code assumes.

### DIFF
--- a/plover/machine/txbolt.py
+++ b/plover/machine/txbolt.py
@@ -63,17 +63,15 @@ class TxBolt(plover.machine.base.SerialStenotypeBase):
     def run(self):
         """Overrides base class run method. Do not call directly."""
         settings = self.serial_port.getSettingsDict()
-        settings['timeout'] = 0.1 # seconds
+        settings['timeout'] = 0.05 # seconds
         self.serial_port.applySettingsDict(settings)
         self._ready()
         while not self.finished.isSet():
             # Grab data from the serial port, or wait for timeout if none available.
             raw = self.serial_port.read(max(1, self.serial_port.inWaiting()))
-            
             # XXX : work around for python 3.1 and python 2.6 differences
             if isinstance(raw, str):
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Ah! My mistake actually, sorry... Why the timeout change?

plover/machine/txbolt.py
```diff
@@ -11,7 +11,7 @@
 # byte represents. The next bits are set if the corresponding key was
 # pressed for the stroke.
 
-# 00XXXXXX 01XXXXXX 10XXXXXX 110XXXXX
+# 00XXXXXX 01XXXXXX 10XXXXXX 11XXXXXX
 #   HWPKTS   UE*OAR   GLBPRF    #ZDST
```
benoit-pierre: I don't think this should be changed, as the line was correct: the last set has only 5 keys.

stanographer:
Fixed the comment change. 100ms is still noticeable while writing at speed. 50ms seems much more "realtime" as I'm writing.

benoit-pierre:
OK, does the last set detection help with regard to #258?

morinted:
Using one of Stan's machines right now... no, it's still occurring. It's happening because the 0-byte is sent only after 10 seconds. However, the ending byte is now working properly.

benoit-pierre:
But the timeout handling is supposed to finish the stroke anyway (that's why I asked why it was lowered, too low, and we could potentially split a stroke in 2).

morinted:
Okay, then perhaps 100ms would be okay.

I've been playing with the machine... I think the reason the slowdown is happening is the same as I had way back when sleeping between key press outputs... instead of sleeping 0.1s, Plover starts sleeping for about 10 seconds between reads! I can see it in the console pretty clearly.

benoit-pierre:
So it's a problem with the output code?

PR #531, opened 2016-05-30
Disable App Nap on OSX

### LINKED ISSUE #189: slowdown in 2.5.8
Lindsey reports:

Hi there!

I use Plover with a Macbook pro, latest verson of OSX.  I use a Cybra using a serial to USB adapter using the TX Bolt protocol.   I haven't updated Plover in a long long time, so I thought I'd see what was up.  I saw that V 2.5.8 was up so I downloaded and installed.  Everything worked fine for about the first 10-20 strokes then everything s l o w e d way down.  Like stroke - wait 2 full seconds - it shows up on screen.  SLOW.

I fiddled with settings and whatnot, but nothing seemed to change it.  I would mess with something, it would be fine for about 2 minutes and then back to being very very slow.

I gave up and installed 2.5.7 instead, but I wanted to put this out there to put it on the radar.  I know that the update from 2.5.7 to 2.5.8 had something to do with the TX Bolt protocol, and for me at least, something went horribly wrong.  2.5.7 is completely fine for me, no slowdown issues at all so whatever happened between those two versions is the culprit.

Thanks!

morinted:
Fixes #189 #258 

App Nap slows down apps when they aren't in the foreground, which really messes up Plover's sleeps. With appnope, we tell the OS that we don't want to be slowed down, and in turn our sleeps no longer "get long" after not being the foreground application for a time.

The result is that the TX Bolt strokes are no longer "held back" and also the output code can use a regular sleep instead of the hacky busy loop.

### DIFF
--- a/plover/machine/txbolt.py
+++ b/plover/machine/txbolt.py
@@ -69,7 +69,7 @@ class TxBolt(plover.machine.base.SerialStenotypeBase):
         while not self.finished.isSet():
             # Grab data from the serial port, or wait for timeout if none available.
             raw = self.serial_port.read(max(1, self.serial_port.inWaiting()))
-            
+
             # XXX : work around for python 3.1 and python 2.6 differences
             if isinstance(raw, str):
                 raw = [ord(x) for x in raw]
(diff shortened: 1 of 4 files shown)


morinted:
I pulled in Stan's TX ending byte stroke change

PR #532, opened 2016-05-30
Fix osx/bootstrap.sh

jeremy-w:
### About
The bootstrap script was updated to gather requirements by running setup.py write_requirements. When I ran it to update dependencies following the addition of appnope, I found this didn't work due to several issues. This PR resolves those issues.

### Issues
None that I know of, but this might be what's broken the weeklies for OS X, which I believe @benoit-pierre mentioned as a new issue on Discord recently.

### Release Notes
None; this is a purely technical change to the dev & build process with no bearing on the operation of Plover itself.

### Tested Platforms
I tested it on OS X by running it to pull in the latest dependencies so I could run `sudo python setup.py launch` successfully again. This PR is platform-specific, so we should be good to go with having just tested on OS X.

### Other
I noticed that `python setup.py build` succeeded in spite of my not having appnope. This is worrisome, because it means we can unknowingly build a broken app.

### DIFF
--- a/osx/bootstrap.sh
+++ b/osx/bootstrap.sh
@@ -93,8 +93,9 @@ EOT
 
 
     echo "$0: installing required libraries using pip"
-    "$TOPDIR/setup.py" write_requirements &&
-    "$PIP" install -r "$TOPDIR/requirements.txt"
+    # Writes requirements to PWD, not setup.py's directory.
+    "$TOPDIR/setup.py" write_requirements \
+        && "$PIP" install -r requirements.txt
 }
 
 main
--- a/osx/bootstrap.sh
+++ b/osx/bootstrap.sh
@@ -9,7 +9,8 @@ EX_FAILURE_NO_CCTOOLS=1
 EX_FAILURE_NO_PYTHON=2
 EX_FAILURE_NO_WXPYTHON=3
 
-TOPDIR="$(dirname "$(readlink -e "$0")")"
+TOPDIR="$(dirname "$(realpath -e "$0")")"
+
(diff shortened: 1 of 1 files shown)


benoit-pierre:
I've got to ask, do we really need `bootstrap.sh`? Travis does not use it (since part of process is common to Linux), and we already list the (few) required instructions in `osx/README.md`. I think we could just tweak the readme a little, similarly to the ones for Linux and Windows. We should definitively mention the fact that setuptools must be downgraded for `make -C osx` to work. That would be one less set of instructions that can be out-of-date.

jeremy-w:
The aim of a bootstrap script is to replace manual readme runbook following with "just run this and you're good to go". If anything, I'd hope to have it handle whatever platform. But perhaps we can just pull the travis script out of the yaml and use that directly? Not using the bootstrap script in CI does seem kind of smelly.

What should we do about setup.py build not ensuring it actually has all the modules needed to build?
--
Jeremy W. Sherman
http://jeremywsherman.com/

morinted:
I like the idea of having a script that runs and installs everything like in Travis, but we'd want it to run interactively in case the user doesn't want to update their Python (or whatever)

Especially the manual downgrade is concerning, so I'd definitely be in favor of a script that can run with -u to be unattended, like the Windows helper.

A document describing the gotchas of the script *may* be useful.

morinted:
I'm going to close this for now. Something still needs to be done, though, as bootstrap.sh is not functional. Maybe it should retired, maybe we should adjust Travis to use it. More PRs and discussion welcome.

Issue #533, opened 2016-05-31
fix keyboard layout handling on OSX


benoit-pierre:
It seems that some layouts are not correctly handled, e.g. Qwertz when using accented letters like "à". We should:
- [x] check if dead keys are correctly detected
- [x] change its `__main__` code to dump the resulting layout detection on stdout to make it easier to debug such issues
- [x] rename `oslayer/mac_keycode.py` to `oslayer/osxkeyboardlayout.py` for consistency
- [x] handle OSX layout changes
- [x] `key_combo.CHAR_TO_KEYNAME` must be used in `oslayer/osxkeyboardcontrol`, `LITERALS` should be removed (not used anymore)

morinted:
> "*U/TKREUT/KAOUT": "{>}{&ú}" just leaves the accent. á and à work as expected.

morinted:
The Qwertz issue came from the fact that Qwertz maps

PR #534, opened 2016-05-31
Improve OSX `send_key_combination`

benoit-pierre:
Use `key_combo.KEYNAME_TO_CHAR`. This address the last point of #533.

Note: untested!

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -44,7 +44,7 @@ from time import sleep
 import collections
 
 from plover.oslayer import mac_keycode
-from plover.key_combo import add_modifiers_aliases, parse_key_combo
+from plover.key_combo import add_modifiers_aliases, parse_key_combo, KEYNAME_TO_CHAR
 import plover.log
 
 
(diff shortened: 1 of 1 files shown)


Achim63:
It doesn't help my cause, unfortunately. I still cannot use e.g. #bracketleft, #bracketright, #braceleft and #braceright, which used to work fine, now they just put out 8, 9, 5 and 6 (the key combination to get the right sign would be alt+8, alt+9 etc.).

morinted:
It was a decision that we made that the "symbol" keys correspond to their base keys from now on.

So the behavior you described is correct…the correct combo would be {#alt(bracketleft)} or {#option(bracketleft)}. This is to prevent ambiguity when you nest these sorts of keys inside modifiers. If you are just trying to send bracket left, you can use the Unicode symbols instead of combos, since they are well supported now.

The symbol names were useful when we didn't have full Unicode support…but now it's available cross all platforms. The key combo keynames are only used now to identify physical keys on the keyboard.

Achim63:
{#alt(bracketleft)} doesn't make sense to me. And I'm really pissed that I'd have to go through my entire dictionary to search for this kind of stuff and change the entries. Guess I'm back to 3.0.0 for now.

morinted:
Why not define your entries as strings like `"{^}[{^}"` or `"{#alt(8)}"`?

Achim63:
"{{^}"  or "{^}}" don't work. No output at all.

morinted:
You'd need to escape the brace: `{\{^}` if you are adding until the add translation dialogue, or `{\\{^}` if editing by hand.

Achim63:
The only reference people have is here: https://sites.google.com/site/ploverdoc/appendix-the-dictionary-format - and that wont work any more.

morinted:
That can be updated → it's already incomplete by a long shot.

Achim63:
"{#alt(8)}" is no good as well – as soon as I switch input source to e.g. US, it wont work any more. I'd expect the dictionaries to have the same output no matter what my qwerty/z keyboard is set to. As mentioned before, mac_keyboard.py does give the right keycodes, no matter which language I picked in system preferences.

morinted:
So what should `{#Shift(braceleft)}` do on each layout? Either way key
combos are not 100% portable. The strings are meant to be "{^}\\{{^}"

benoit-pierre:
The goal of combos is to simulate shortcuts, by simulating a series of key press/release events, so depending on the layout, the keys change (and some are not available). If you want to send a specific character, you should use the normal text output: it's here for that, it's portable (it will automatically map the character if not available on Linux, send Unicode in Windows,...). As already stated, if you're particular issue is that a character does not work (like "à"), then it's a bug with the way we handle the Keyboard layout on Mac, and we need to fix that so you don't have to use a hack (a key combo) instead.

PR #536, opened 2016-06-01
formatting: check an action word too when handling replace

### LINKED ISSUE #535: Currency forming with a decimal fails to erase previous output
Classification: Bug

Reproducibility: Always

## Summary

Currency formatting a number with a decimal fails by not erasing the previous output.

## Steps to Reproduce

1. Write `1.2` **using number bar** (handles a definition set to "1.2" properly…)
2. Use `{*($c)}`

## Expected Results

3. Output is `$1.20`

## Actual Results

3. `1.2$1.20`

## Version

Plover a90528e2fcf7bf2625ab39916bd8d9328fd34922 and above, @benoit-pierre perhaps you could investigate this particular case?

```
a90528e2fcf7bf2625ab39916bd8d9328fd34922 is the first bad commit
commit a90528e2fcf7bf2625ab39916bd8d9328fd34922
Author: Benoit Pierre <benoit.pierre@gmail.com>
Date:   Sat Apr 23 18:34:48 2016 +0200

    Rework handling of actions 'replace', fix 476 (#481)

    * formatting: rework handling of actions 'replace'

    Don't honor 'replace' if it does not match at the end of the previous
    text.
```

benoit-pierre:
Fix #535.

Use case: when `{*($c)}` is used after stroking `1.2` (with 3 separate strokes), the formatter will try to undo 1 action only with `text` set to `2` and `word` set to `1.2`, with one action replacing `1.2` by `$1.20`. So we need to take `action.word` into account when rendering the initial (previous) text.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -158,7 +158,11 @@ class OutputHelper(object):
         for a in action_list:
             if a.replace and text.endswith(a.replace):
                 text = text[:-len(a.replace)]
-            if a.text:
+            # With numbers, it's possible to have a.text='2' with a.word='1.2'
+            # folowing by an action that replaces '1.2' by '$1.20'...
+            if len(a.word) > len(a.text) and a.word.endswith(text):
+                text = a.word
+            else:
                 text += a.text
         return text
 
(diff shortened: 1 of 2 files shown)


morinted:
I love the blackbox test!

PR #537, opened 2016-06-04
Prevent crash on OSX caused by freeing up resources too quickly

morinted:
Refactor the code to read a little simpler, putting the cleanup code after the loop stops instead of in the cancel method.

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -3,7 +3,7 @@ from Quartz import (
     CFMachPortCreateRunLoopSource,
     CFMachPortInvalidate,
     CFRunLoopAddSource,
-    CFRunLoopRemoveSource,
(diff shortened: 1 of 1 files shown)


morinted:
@benoit-pierre updated with your feedback!

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -224,40 +220,36 @@ def callback(proxy, event_type, event, reference):
...
-            self._running_thread, self._source, kCFRunLoopCommonModes)
-        CFRelease(self._source)
-        self._source = None
+        CFRunLoopRun()
 
+        # Wake up event handler.
+        self._event_queue.put_nowait(None)
+        handler_thread.join()
         CFMachPortInvalidate(self._tap)
+        handler_thread.join()
```
benoit-pierre: I think one join is enough. ;)

morinted:
I removed the extra join

PR #538, opened 2016-06-08
Fix Mac OS X keyboard layout

### LINKED ISSUE #533: fix keyboard layout handling on OSX

morinted:
Fix #533 

- osxkeyboardlayout is now a class
- Useful main method that debugs the parsed layout
- Favor key combos with less modifiers
- Detect key layouts changing

### DIFF
--- a/plover/misc.py
+++ b/plover/misc.py
@@ -2,8 +2,19 @@
 # See LICENSE.txt for details.
 
 
-def popcount32(v):
-    """Population count for a 32bits integer."""
+def popcount_8(v):
+    """Population count for an 8 bit integer"""
+    assert 0 <= v <= 255
     v -= ((v >> 1) & 0x55555555)
     v = (v & 0x33333333) + ((v >> 2) & 0x33333333)
     return ((v + (v >> 4) & 0xF0F0F0F) * 0x1010101) >> 24
+
+
+def characters(string):
+    """Python 2.7 has narrow Unicode on Mac OS X and Windows"""
+
+    encoded = string.encode('utf-32-be')  # 4 bytes per character
+    for char_start in xrange(0, len(encoded), 4):
+        end = char_start + 4
+        character = encoded[char_start:end].decode('utf-32-be')
+        yield character
--- a/plover/misc.py
+++ b/plover/misc.py
@@ -1,0 +1,9 @@
+# Copyright (c) 2016 Open Steno Project
+# See LICENSE.txt for details.
+
+
+def popcount32(v):
+    """Population count for a 32bits integer."""
+    v -= ((v >> 1) & 0x55555555)
+    v = (v & 0x33333333) + ((v >> 2) & 0x33333333)
+    return ((v + (v >> 4) & 0xF0F0F0F) * 0x1010101) >> 24
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -41,8 +41,7 @@ import Foundation
 import threading
 import Queue
 from time import sleep
-import collections
-
+from plover.misc import characters
 from plover.oslayer.osxkeyboardlayout import KeyboardLayout
 from plover.key_combo import add_modifiers_aliases, parse_key_combo, KEYNAME_TO_CHAR
 import plover.log
@@ -292,18 +291,6 @@ class KeyboardCapture(threading.Thread):
             key, is_keyup = pair
             handler = self.key_up if is_keyup else self.key_down
             handler(key)
-
-
-# "Narrow python" unicode objects store characters in UTF-16 so we
-# can't iterate over characters in the standard way. This workaround
-# lets us iterate over full characters in the string.
-def characters(s):
-    encoded = s.encode('utf-32-be')
-    for i in xrange(len(encoded) / 4):
-        start = i * 4
-        end = start + 4
-        character = encoded[start:end].decode('utf-32-be')
-        yield character
 
 
 class KeyboardEmulation(object):
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -96,6 +96,13 @@ for name, code in NX_KEYS.iteritems():
     KEYNAME_TO_KEYCODE[name.lower()] = code + NX_KEY_OFFSET
 add_modifiers_aliases(KEYNAME_TO_KEYCODE)
 
+DEADKEY_SYMBOLS = {
+    'dead_acute': '´',
+    'dead_circumflex': '^',
+    'dead_diaeresis': '¨',
+    'dead_grave': '`',
+    'dead_tilde': '~',
+}
 
 def down(seq):
     return [(x, True) for x in seq]
@@ -289,7 +296,7 @@ class KeyboardCapture(threading.Thread):
 
 # "Narrow python" unicode objects store characters in UTF-16 so we
 # can't iterate over characters in the standard way. This workaround
-# let's us iterate over full characters in the string.
+# lets us iterate over full characters in the string.
 def characters(s):
     encoded = s.encode('utf-32-be')
     for i in xrange(len(encoded) / 4):
@@ -404,11 +411,19 @@ class KeyboardEmulation(object):
 
         """
         def name_to_code(name):
+            # Static key codes
             code = KEYNAME_TO_KEYCODE.get(name)
             if code is not None:
-                return code
-            char = KEYNAME_TO_CHAR.get(name, name)
-            code, mods = self._layout.char_to_key_sequence(char)[0]
+                pass
+            # Dead keys
+            elif name.startswith('dead_'):
+                code, mod = self._layout.deadkey_symbol_to_key_sequence(
+                    DEADKEY_SYMBOLS.get(name)
+                )[0]
+            # Normal keys
+            else:
+                char = KEYNAME_TO_CHAR.get(name, name)
+                code, mods = self._layout.char_to_key_sequence(char)[0]
             return code
         # Parse and validate combo.
         key_events = parse_key_combo(combo_string, name_to_code)
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -43,7 +43,7 @@ import Queue
 from time import sleep
 import collections
 
-from plover.oslayer import mac_keycode
+from plover.oslayer.osxkeyboardlayout import KeyboardLayout
 from plover.key_combo import add_modifiers_aliases, parse_key_combo, KEYNAME_TO_CHAR
 import plover.log
 
@@ -77,7 +77,7 @@ KEYNAME_TO_KEYCODE = {
     'caps_lock': 57,
 
     'return': 36, 'tab': 48, 'backspace': BACK_SPACE, 'delete': 117,
-    'escape': 53,
+    'escape': 53, 'clear': 71,
 
     'up': 126, 'down': 125, 'left': 123, 'right': 124, 'page_up': 116,
     'page_down': 121, 'home': 115, 'end': 119,
@@ -304,7 +304,7 @@ class KeyboardEmulation(object):
     RAW_PRESS, STRING_PRESS = range(2)
 
     def __init__(self):
-        pass
+        self._layout = KeyboardLayout()
 
     @staticmethod
     def send_backspaces(number_of_backspaces):
@@ -351,7 +351,7 @@ class KeyboardEmulation(object):
 
         last_modifier = None
         for c in characters(s):
-            for keycode, modifier in mac_keycode.KeyCodeForChar(c):
+            for keycode, modifier in self._layout.char_to_key_sequence(c):
                 if keycode is not None:
                     if modifier is not last_modifier:
                         # Flush on modifier change.
@@ -408,7 +408,7 @@ class KeyboardEmulation(object):
             if code is not None:
                 return code
             char = KEYNAME_TO_CHAR.get(name, name)
-            code, mods = mac_keycode.KeyCodeForChar(char)[0]
+            code, mods = self._layout.char_to_key_sequence(char)[0]
             return code
         # Parse and validate combo.
         key_events = parse_key_combo(combo_string, name_to_code)
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2016 Open Steno Project
+# See LICENSE.txt for details.
+
+"""Tests for misc.py."""
+
+import unittest
+
+import plover.misc as misc
+
+
+class MiscTestCase(unittest.TestCase):
+    def test_popcount_8(self):
+        for n in range(256):
+            self.assertEqual(misc.popcount_8(n),
+                             format(n, 'b').count('1'))
+        with self.assertRaises(AssertionError):
+            misc.popcount_8(256)
+        with self.assertRaises(AssertionError):
+            misc.popcount_8(-1)
+
+    def test_characters(self):
+        self.assertEqual(list(misc.characters('STKPWHR')),
+                         ['S', 'T', 'K', 'P', 'W', 'H', 'R'])
+        self.assertEqual(list(misc.characters('')), [])
+        self.assertEqual(list(misc.characters(u'✂⌨')), [u'✂', u'⌨'])
--- a/plover/oslayer/osxkeyboardlayout.py
+++ b/plover/oslayer/osxkeyboardlayout.py
@@ -9,6 +9,8 @@ import ctypes
 import ctypes.util
 import struct
 import unicodedata
+import re
+from plover.key_combo import CHAR_TO_KEYNAME
 
 try:
     unichr
@@ -22,7 +24,6 @@ CONTROL = u'⌃'
 CAPS = u'⇪'
 UNKNOWN = u'?'
 UNKNOWN_L = u'?_L'
-HEADER = u'KEYCODE | {}   | {}      | {}      | {}      | {}     | {}     | {}     | {}'.format('None', SHIFT, CAPS, OPTION, COMMAND + OPTION, SHIFT + OPTION, CAPS + OPTION, CONTROL).expandtabs(9)
 
 carbon_path = ctypes.util.find_library('Carbon')
 carbon = ctypes.cdll.LoadLibrary(carbon_path)
@@ -251,38 +252,33 @@ def KeyCodeForChar(character):
             result = result,
         return result
 
+def format_modifier_header():
+    modifiers = (u'| {}\t'.format(modstr(mod)).expandtabs(8)
+                 for mod in sorted(modmapping.values()))
+    header = u'Keycode\t{}'.format(''.join(modifiers))
+    return '%s\n%s' % (header, re.sub(r'[^|]', '-', header))
 
-def print_keycode_keys(keycode):
+def format_keycode_keys(keycode):
     """ Prints all the variations of the Keycode with modifiers """
     keys = (u'| {}\t'.format(printify(revmapping[keycode, mod])).expandtabs(8)
             for mod in sorted(modmapping.values()))
 
-    print(u'{}\t{}'.format(keycode, ' '.join(keys))).expandtabs(8)
+    return u'{}\t{}'.format(keycode, ''.join(keys)).expandtabs(8)
 
 if __name__ == '__main__':
-    import sys
-    for arg in sys.argv[1:]:
-        try:
-            arg = arg.decode(sys.stdin.encoding)
-        except AttributeError:
-            pass
-        try:
-            keycode = int(arg)
-        except ValueError:
-            for ch in arg:
-                result = mapping.get(ch, (None, 0))
-                if result is None or result[0] is None:
-                    print(u"{}: not found".format(printify(ch)))
-                else:
-                    if not isinstance(result[0], tuple):
-                        result = result,
-                    print(u"{}: {}".format(printify(ch),
-                                           ', '.join('{}{}'.format(modstr(mod), keycode)
-                                                     for keycode, mod in result)))
-        else:
-            print_keycode_keys(keycode)
-    if len(sys.argv) < 2:
-        print HEADER
-        print '-' * len(HEADER)
-        for keycode in range(127):
-            print_keycode_keys(keycode)
+    print format_modifier_header()
+    for keycode in range(127):
+        print format_keycode_keys(keycode)
+    unmapped_characters = []
+    for char, keyname in sorted(CHAR_TO_KEYNAME.iteritems()):
+        sequence = []
+        for code, mod in KeyCodeForChar(char):
+            if code is not None:
+                sequence.append(u'{}{}'.format(modstr(mod), printify(revmapping[code, 0])))
+            else:
+                unmapped_characters.append(char)
+        if sequence:
+            print u'Name:\t\t{}\nCharacter:\t{}\nSequence:\t‘{}’\n'.format(keyname, char, u'’, ‘'.join(sequence))
+    print u'No mapping on this layout for characters: ‘{}’'.format(u'’, ‘'.join(unmapped_characters))
+
+
--- a/plover/oslayer/osxkeyboardlayout.py
+++ b/plover/oslayer/osxkeyboardlayout.py
@@ -11,6 +11,7 @@ import struct
 import unicodedata
 import re
 from plover.key_combo import CHAR_TO_KEYNAME
+from plover.misc import popcount32
 
 try:
     unichr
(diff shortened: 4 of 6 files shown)


morinted:
@Achim

Could you try this branch with your setup?

Achim63:
The accented characters work now. There's only one key left that I cannot access as expected: the accent key left of the 1 (the French accent circonflexe). "TR/TKR": "{^}^{^}" produces nothing, "SH/TKR": "{#Shift_L(asciitilde)}" produces "N" instead of the expected "°" (so I just use "{^}°{^}")). Alt+^ should produce "„" and Alt+Shift+^ "“" – the standard German quote signs. I'd prefer not to "hard-code" those signs into my modifier dictionary so that it's not bound to German layout.
Interestingly, "{#Alt_L(asciitilde)}" produces the ~ which waits for another character (Alt+n on qwertz), so I can use it to produce ñ and Ñ. This is rather exciting, since it shows that it's possible to type an accent from the steno machine and then the letter that should get the accent (like Alt+u and then u gives ü, which works now also in steno!). But I couldn't figure out how to get the ´, ` and ^ that wait for the next character typed. Having a stroke for those would make producing accented characters much easier.

morinted:
You could run the test code to get a better idea of what the program is
doing with your edge cases, and also see where the dead keys are on your
layout:

`python -m plover.oslayer.osxkeyboardlayout`

morinted:
Okay, so for the circum symbol, I get:

```
Name:		asciicircum
Character:	^
Sequence:	‘⌘dk ^’
Base codes:	‘10’
```

Which is wrong because of the command. I think I need to favor non-command...

For asciitilde, I have:

```
Name:		asciitilde
Character:	~
Sequence:	‘⌃⌥n’
Base codes:	‘45’
```

which is actually valid to produce a ~. If you want a degree symbol from a keyboard shortcut, use ⌥⇧ü, or `{#option(shift(udiaeresis))}`

Best way to get your quotes is ⌥⇧w for „ and ⌥2 for “

I'll have to study a little bit to see how best to handle dead keys... I may try to never return them by default, and provide a special way to request a deadkey. As for other dead keys...

- option(u) is the dk ¨
- option(n) is the dk ~
- option(shift(6)) is the dk ^
- Right now, I believe `{#acute}` is the acute deadkey, and shift(acute) is the ` deadkey.

To produce just a caret... seems like I need to use Control option shift 6: ^

I think the only thing the script actually did *wrong* was selecting the character that had cmd as a modifier, I should adjust my modifier-favoring code to strongly dislike the command modifier, as it never seems appropriate (and in Qwertz, unlike Qwerty, control is often used).

You can see the dump of the layout here:

https://gist.github.com/morinted/7c2c4ae1a9b3aaf386d72dd1883f110a

The big reason that you are running into deadkeys is because we default to the *base* keycode without modifiers, and on German you have deadkeys in the layout. You just need to work around them.

You keep referring to an asciicircum key... I don't have one on the Macbook, I see some in pictures of old keyboards... do you have a photo? Is it a deadkey? Is it keycode 10? You can use this tool to get keycodes: https://itunes.apple.com/ca/app/key-codes/id414568915?mt=12

I tried your German quotes, the program sends option(asciicircum) and option(2); they seem to work fine over here.

-------

Overall, it all seems to work fine other than ^ not outputting.

morinted:
Fixed it up locally... only problem that I can find is that ` isn't a valid
key on Qwertz... can't find a workaround for that.

Achim63:
`is Shift+´ (on a US keyboard, this would produce +).

morinted:
That's a deadkey for accenting, I see no way to just output the character. Therefore, Plover won't be able to output "`" on a Qwertz layout (because there are entries for ` in the keymap, but they all don't actually work)

The workaround would be `{#shift(acute) space}`

Achim63:
I was wrong, they actually all work now, e.g. "TR/TA": "{#acute}" actually gets the accent and waits for the next stroke. Same for `and ^. I don't know what I did wrong, but it's all good now!

Achim63:
A moment later: my modifier key dictionary is complete for a German Apple keyboard now. So whatever strange keystroke might be defined for a menu item in any app, I should be able to call it from the steno machine. The only key I didn't care about is the fn key (but I think that's special anyway and only works together with the F-keys.).

morinted:
I updated again, just now, I had some success by also selecting deadkeys with fewer modifiers, and the grave accent now works :)

Overall I'm *much* more confident in the code now that it's been reorganized. Amazing how proper variable naming makes a difference in readability. I'll have to resquash but for now this is good.

Achim63:
I tried this, and now it's broken again for me (the "asciitilde" doesn't give the "dead" one that waits for the next stroke). It worked in the previous commit.

morinted:
You should be using `{#option(n)}` for dead asciitilde I believe.

morinted:
The table tells all:


Keycode	|       | ⌘     | ⇧     | ⇧⌘    | ⇪     | ⌥     | ⌥⌘    | ⌥⇧    | ⌥⇧⌘   | ⌥⇪    | ⌃     | ⌃⇧    | ⌃⌥    | ⌃⌥⇧   
--------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------
0       | a     | a     | A     | a     | A     | å     | å     | Å     | Å     | Å     | Home  | Home  | å     | Å     
1       | s     | s     | S     | s     | S     | ‚     | ‚     | Í     | Í     | ‚     | \x13  | \x13  | ‚     | Í     
2       | d     | d     | D     | d     | D     | ∂     | ∂     | ™     | ™     | ∂     | End   | End   | ∂     | ™     
3       | f     | f     | F     | f     | F     | ƒ     | ƒ     | Ï     | Ï     | ƒ     | \x06  | \x06  | ƒ     | Ï     
4       | h     | h     | H     | h     | H     | ª     | ª     | Ó     | Ó     | ª     | Bksp  | Bksp  | ª     | Ó     
5       | g     | g     | G     | g     | G     | ©     | ©     | Ì     | Ì     | ©     | \x07  | \x07  | ©     | Ì     
6       | y     | y     | Y     | y     | Y     | ¥     | ¥     | ‡     | ‡     | ¥     | \x19  | \x19  | ¥     | ‡     
7       | x     | x     | X     | x     | X     | ≈     | ≈     | Ù     | Ù     | ≈     | \x18  | \x18  | ≈     | Ù     
8       | c     | c     | C     | c     | C     | ç     | ç     | Ç     | Ç     | Ç     | \x03  | \x03  | ç     | Ç     
9       | v     | v     | V     | v     | V     | √     | √     | ◊     | ◊     | √     | \x16  | \x16  | √     | ◊     
10      | dk^   | ^     | °     | °     | dk^   | „     | „     | “     | ’     | „     | \x1e  | °     | „     | “     
11      | b     | b     | B     | b     | B     | ∫     | ∫     | ‹     | ‹     | ∫     | \x02  | \x02  | ∫     | ‹     
12      | q     | q     | Q     | q     | Q     | «     | «     | »     | »     | «     | \x11  | \x11  | «     | »     
13      | w     | w     | W     | w     | W     | ∑     | ∑     | „     | „     | ∑     | \x17  | \x17  | ∑     | „     
14      | e     | e     | E     | e     | E     | €     | €     | ‰     | ‰     | €     | Help  | Help  | €     | ‰     
15      | r     | r     | R     | r     | R     | ®     | ®     | ¸     | ¸     | ®     | \x12  | \x12  | ®     | ¸     
16      | z     | z     | Z     | z     | Z     | Ω     | Ω     | ˇ     | ˇ     | Ω     | \x1a  | \x1a  | Ω     | ˇ     
17      | t     | t     | T     | t     | T     | †     | †     | ˝     | ˝     | †     | \x14  | \x14  | †     | ˝     
18      | 1     | 1     | !     | 1     | 1     | ¡     | ¡     | ¬     | ¬     | ¡     | 1     | !     | ¡     | ¬     
19      | 2     | 2     | "     | 2     | 2     | “     | “     | ”     | ”     | “     | 2     | "     | “     | ”     
20      | 3     | 3     | §     | 3     | 3     | ¶     | ¶     | #     | #     | ¶     | 3     | §     | ¶     | #     
21      | 4     | 4     | $     | 4     | 4     | ¢     | ¢     | £     | £     | ¢     | 4     | $     | ¢     | £     
22      | 6     | 6     | &     | 6     | 6     | ]     | ]     | dk^   | ^     | ]     | 6     | &     | \x1d  | ^     
23      | 5     | 5     | %     | 5     | 5     | [     | [     | ﬁ     | ﬁ     | [     | 5     | %     | Esc   | ﬁ     
24      | dk´   | ´     | dk`   | `     | dk´   | '     | '     | ˚     | ˚     | '     | ´     | `     | '     | ˚     
25      | 9     | 9     | )     | 9     | 9     | }     | }     | ·     | ·     | }     | 9     | )     | }     | ·     
26      | 7     | 7     | /     | 7     | 7     | |     | |     | \     | \     | |     | 7     | /     | |     | \x1c  
27      | ß     | ß     | ?     | /     | ß     | ¿     | ¿     | ˙     | ˙     | ¿     | ß     | ?     | ¿     | ˙     
28      | 8     | 8     | (     | 8     | 8     | {     | {     | ˜     | ˜     | {     | 8     | (     | {     | ˜     
29      | 0     | 0     | =     | 0     | 0     | ≠     | ≠     | ¯     | ¯     | ≠     | 0     | =     | ≠     | ¯     
30      | +     | +     | *     | *     | +     | ±     | ±     |      |      | ±     | +     | *     | ±     |      
31      | o     | o     | O     | o     | O     | ø     | ø     | Ø     | Ø     | Ø     | \x0f  | \x0f  | ø     | Ø     
32      | u     | u     | U     | u     | U     | dk¨   | ¨     | Á     | Á     | dk¨   | \x15  | \x15  | ¨     | Á     
33      | ü     | ü     | Ü     | ü     | Ü     | •     | •     | °     | °     | •     | \x1d  | \x1d  | •     | °     
34      | i     | i     | I     | i     | I     | ⁄     | ⁄     | Û     | Û     | ⁄     | \t    | \t    | ⁄     | Û     
35      | p     | p     | P     | p     | P     | π     | π     | ∏     | ∏     | ∏     | \x10  | \x10  | π     | ∏     
36      | \r    | \r    | \r    | \r    | \r    | \r    | \r    | \r    | \r    | \r    | \r    | \r    | \r    | \r    
37      | l     | l     | L     | l     | L     | @     | @     | ﬂ     | ﬂ     | @     | PgDn  | PgDn  | @     | ﬂ     
38      | j     | j     | J     | j     | J     | º     | º     | ı     | ı     | º     | \n    | \n    | º     | ı     
39      | ä     | ä     | Ä     | ä     | Ä     | æ     | æ     | Æ     | Æ     | Æ     | Esc   | Esc   | æ     | Æ     
40      | k     | k     | K     | k     | K     | ∆     | ∆     | ˆ     | ˆ     | ∆     | PgUp  | PgUp  | ∆     | ˆ     
41      | ö     | ö     | Ö     | ö     | Ö     | œ     | œ     | Œ     | Œ     | Œ     | \x1c  | \x1c  | œ     | Œ     
42      | #     | #     | '     | ^     | #     | ‘     | ‘     | ’     | ’     | ‘     | #     | \x1e  | ‘     | ’     
43      | ,     | ,     | ;     | ;     | ,     | ∞     | ∞     | ˛     | ˛     | ∞     | ,     | ;     | ∞     | ˛     
44      | -     | -     | _     | _     | -     | –     | –     | —     | —     | –     | -     | \x1f  | –     | —     
45      | n     | n     | N     | n     | N     | dk~   | ~     | ›     | ›     | dk~   | \x0e  | \x0e  | ~     | ›     
46      | m     | m     | M     | m     | M     | µ     | µ     | ˘     | ˘     | µ     | \r    | \r    | µ     | ˘     
47      | .     | .     | :     | :     | .     | …     | …     | ÷     | ÷     | …     | .     | :     | …     | ÷

Achim63:
It's not about asciitilde, it's about asciicircum that now stopped working. And as I said: in the first commit that I tried it worked just fine.

morinted:
Right, right, because ^ is a dead key on your base layout, you can't talk about it directly (consequence of asciicircum coming from the 6 key). If you want to use the base asciicircum you need to use a key that uses that keycode (10), e.g. degree.

So, just `{#degree}` will give you a dead circumflex. If you look at the above table, you have a couple options:

Keycode	|       | ⌘     | ⇧     | ⇧⌘    | ⇪     | ⌥     | ⌥⌘    | ⌥⇧    | ⌥⇧⌘   | ⌥⇪    | ⌃     | ⌃⇧    | ⌃⌥    | ⌃⌥⇧   
--------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------|-------
10      | dk^   | ^     | °     | °     | dk^   | „     | „     | “     | ’     | „     | \x1e  | °     | „     | “  
22      | 6     | 6     | &     | 6     | 6     | ]     | ]     | dk^   | ^     | ]     | 6     | &     | \x1d  | ^     

To use dk^, just do `{#degree}` or `{#option(shift(6))}`

Achim63:
You could even try this while having the US layout. "TR/TKR": "{#asciicircum}" outputs 6 there as well.

morinted:
As expected, because the deadkey circumflex on the US layout is `{#option(i)}`

morinted:
> Also "SH/TKR": "{#Shift_L(asciicircum)}" doesn't produce the ° anymore as it should. I now get & instead.

`{#shift(degree)}` will do. So will `{#option(shift(udiaeresis))}`

Achim63:
Whatever, I think {#asciicircum} should give ^ (dead or alive, I'd prefer dead), independent of layout. That's what it's all about – if it worked as in the first commit, people using French or Spanish keyboards could use my dictionary without changes and rummaging in key tables.

morinted:
It shouldn't. Key combos are layout-dependent. We've been over this.

I'm not sure if there's any interest in supporting a layout-agnostic dead key feature, but it's not what we've intended to do nor what we've built.

In actuality, the system already knows when it can use dead keys to produce, say, ü. Plover should make it so that you don't have to think about key strokes (or accent-key strokes for that matter). Right now if you use ü in a translation, it will work across Windows, Linux, and OS X, and supports layout switching on Windows and OS X. It will smartly use the u diaeresis key on Qwertz and the appropriate deadkey on Qwerty.

I'm sure we can invent a use case where it's useful to have a dead key stroke, but I think for general usage it's just not a thing.

@benoit-pierre I believe this is ready for review when you have time, I can squash in the fixup commits after.

Achim63:
I thought the key names were there for that exact purpose. {#plus} gives +, independent of layout. {#space} gives a space etc. So I really don't get why {#asciicircum} shouldn't give ^ independently.

Having the dead keys would make fingerspelling of names and places so much easier – just one stroke to remember for each kind of accent, not one for each vowel combined with each accent (ten for each accent áéíóúÁÉÍÓÚ). Málaga, Setúbal, Besançon, Épinal, Lac Léman, just a few locations that come to mind. I'm quite sure Mexico and Canada also have a lot of proper names using accents.

I'll certainly keep the files and build where all the dead keys and everything else work – the one you asked me to test in the first place.

morinted:
Basically, originally Plover used {#plus} to mean + cross platform, but
since then we've implemented Unicode support. So "+" in a translation works
layout independent.

Then recently, Benoit refractored the key combo code so that it was the
same on all platforms. Some common issues came up in the process: people
having difficulty knowing when to use a vs A, down not working when Down
does, the behavior of, say, Shift(plus a B c). In some implementations,
that would output +aBc instead of the expected +ABC. The reason is that
dealing with keys that have their own modifiers is difficult.

So we looked at what the key combos are actually useful for. Should it be
the main way to output symbols? No, because Unicode extends far beyond our
listed subset. So for symbols, we recommend using them in translation. Key
combos are most useful for keyboard shortcuts. It's the only way you can
force modifiers like command or control. So, what is a keyboard shortcut?
It is a modifier and a base key. Modifiers are easy, shift, control,
command, alt/option. We know what those are and they stay the same across
layouts.

The issue comes from the base key. Again, what should control(C) do? Is it
control shift c or control c? We decided to remove complication by opting
to say that no symbol name (a, b, c, 1, plus, backspace) should come with
hidden modifiers. It's simply too hard to predict cross layout and there
may be lots of unexpected behavior.

Unfortunately the alternative is what you've noticed, where we only use key
names to refer to physical keys. We don't include the modifiers for you.
"plus" is probably a real key on some layouts, but as you know on Qwerty it
is Shift = , so the plus name refers to that physical key. If you try to
send plus but don't send modifiers, you get equals. So if you want to use
plus, if for some reason you can't use it as a translation, then you have
to use the modifier. On Qwerty that means Shift(plus)… there is no question
about the modifiers. Shift(plus a B c) will send +ABC because they are all
under the shift modifier.

I am considering adding dead key support to this PR because Qwertz has
deadkeys as base keys, so it might be useful to refer to your "^" key as
dead_asciicircum, but that will still only point to whatever physical key
holds the dead asciicirum on other layouts.

I'm also not convinced that that older commit will work across different
layouts, it is a side effect of Qwertz having a deadkey in its layout,
which US does not.

If you want to have a generic deadkey, the best I can suggest is to create
a dictionary like the modifier dictionaries you have created previously.

Sort of like fingerspelling, but you prefix with a symbol stroke.

Say, SKPH* = {#}, SKPH*/*E = é, SKPH*/*EP = É and so on. Yes, that is a
little labor intensive, but it's infinitely more compatible cross-OS and
cross-layout than trying to use dead keys (regardless of Plover's design)

I'm also pretty certain that using the above method will let your strokes
be sensitive to {-|} and the like.

I get what you are trying to accomplish with the key combos, but I just
think that it's not achievable while still working well and consistently
across platforms, which is a very important goal for Plover. You've had the
benefit of working on OSX where the key combo code was overpowered in its
own special way, but this new solution is wayyy ahead of the miniscule
piece of the pie that Windows users had to deal with previously.

I do understand that we've really disrupted your use of the key combos, but
I think you may have been the only one using them so extensively. I hope
the new syntax and behavior makes it easier for people who aren't so great
at syntax parsing, as most common commands like {#alt(tab)} are easier to
write than they were: {#Alt_L(Tab)}

As always, if you don't like it, you can fork it and revert it, you can
concoct new solutions and run them by the community on Discord, you can
submit GitHub issues and PRs. This is a decision that's been made for what
I believe are sane reasons, and it wasn't meant to make your life
difficult, so I am sorry for the trouble. Hopefully particular issues you
come up with I can help you solve.

Excuse typos or terseness, tapping away on a touchscreen.

Ted

On Jun 8, 2016 6:29 PM, "Achim" <notifications@github.com> wrote:
>
> I thought the key names were there for that exact purpose. {#plus} gives
+, independent of layout. {#space} gives a space etc. So I really don't get
why {#asciicircum} shouldn't give ^ independently.
>
> Having the dead keys would make fingerspelling of names and places so
much easier – just one stroke to remember for each kind of accent, not one
for each vowel combined with each accent (ten for each accent áéíóúÁÉÍÓÚ).
Málaga, Setúbal, Besançon, Épinal, Lac Léman, just a few locations that
come to mind. I'm quite sure Mexico and Canada also have a lot of proper
names using accents.
>
> I'll certainly keep the files and build where all the dead keys and
everything else work – the one you asked me to test in the first place.

Achim63:
Thank you for the long explanation. I have no idea how all of this is different on Windows, so I wouldn't be much help there. I was just so glad that the accent keys were working – I'm using a slightly adapted version of the already existing diacritics dictionary from stenodict. So I hoped the three stroke definitions could be shortened to two strokes for the most common accents. One of the first things I stumbled upon when starting out with Plover was the difficulty I had to type "à la". This went to my dictionary as A/LA, so no problem there. But what to do with proper nouns that you probably will only type once in a lifetime so stick to fingerspelling?

Don't forget that people with a mother tongue of Spanish and French taken together exceed the number of people with English as their mother tongue. So they also probably have accents on their keyboards, and if typing English steno, they'd want to be able to finger spell proper names as fast and easy as possible.

Those accent keys on German, Spanish and French keyboards are really convenient to have if you're used to them. I can type Spanish or French on qwertz without having to learn new key combinations for all the accented vowels. I'm used to those keys since I learned to type (i.e. for over 40 years), so that's probably why I have such difficulty to have to live without them in steno.

And I'm probably over-sensitive regarding keyboard shortcuts – the conventions introduced by Apple as given in their Human Interface Guidelines (which I more or less learned by heart at one point) were one of the reasons to pick Macs all of my life. As an example: I always hated Adobe for "stealing" Cmd+h (hide application) – just because they used it on Windows. I often moved apps to the trash that had their origin on Windows, just because they had their own idea of shortcuts (and requesters with "ok" and "cancel" in the wrong places, close buttons in the upper right corner etc.), not consistent with the Apple system.
Sorry for the digression, I forgot what this has to do with steno myself by now ;-). Probably "consistency" – no matter what keyboard layout I use, I expect a steno dictionary entry to give the same results when sent from my steno machine. I see you're well on your way to that, but having to make compromises because Windows can't handle some things is just sad.

morinted:
Maybe try this in a translation?

http://www.fileformat.info/info/unicode/char/0302/index.htm

morinted:
I've updated the branch with these additional key names (for OS X):

- `dead_asciicircum`
- `dead_asciitilde`
- `dead_acute`
- `dead_grave`
- `dead_diaeresis`

You can now name the key you want if it's a dead key, like `dead_asciicircum` on Qwertz

morinted:
@benoit-pierre I've cleaned up the history some

benoit-pierre:
FYI, the dead circumflex/tilde names in the Linux and Windows code are `dead_circumflex` and `dead_tilde` (named after the X11 keysyms).

plover/oslayer/osxkeyboardcontrol.py
```diff
@@ -289,7 +293,7 @@ def _event_handler(self):
 
 # "Narrow python" unicode objects store characters in UTF-16 so we
 # can't iterate over characters in the standard way. This workaround
-# let's us iterate over full characters in the string.
+# lets us iterate over full characters in the string.
 def characters(s):
```
benoit-pierre: We should probably move this to the newly added `plover.misc` (and remove the duplicate in `winkeyboardcontrol`).

morinted:
To work with `dead_circumflex/tilde` I had to move some things around, I think I like it more now.

test/test_misc.py
```diff
@@ -0,0 +1,34 @@
...
+        self.assertEqual(misc.popcount_8(0), 0)
+        self.assertEqual(misc.popcount_8(1), 1)
+        self.assertEqual(misc.popcount_8(2), 1)
+        self.assertEqual(misc.popcount_8(3), 2)
+        self.assertEqual(misc.popcount_8(7), 3)
+        self.assertEqual(misc.popcount_8(15), 4)
+        self.assertEqual(misc.popcount_8(31), 5)
+        self.assertEqual(misc.popcount_8(63), 6)
+        self.assertEqual(misc.popcount_8(127), 7)
+        self.assertEqual(misc.popcount_8(255), 8)
```
benoit-pierre: We could just check the whole range like this:
```python
        for n in range(256):
            self.assertEqual(misc.popcount_8(n),
                             format(n, 'b').count('1'))
```

test/test_misc.py
```diff
@@ -0,0 +1,34 @@
...
+            misc.popcount_8(256)
+        with self.assertRaises(AssertionError):
+            misc.popcount_8(-1)
+
+    def test_characters(self):
+        def len_generator(generator):
+            return reduce(lambda x, y: x + 1, generator, 0)
+        self.assertEqual(len_generator(misc.characters('STKPWHR')), 7)
+        self.assertEqual(len_generator(misc.characters('')), 0)
+        self.assertEqual(len_generator(misc.characters(u'✂⌨')), 2)
```
benoit-pierre: Wouldn't it be better to actually check the elements of the generated list?
```python
        self.assertEqual(list(misc.characters('STKPWHR')), ['S', 'T', 'K', 'P', 'W', 'H', 'R'])
        self.assertEqual(list(misc.characters('')), [])
        self.assertEqual(list(misc.characters(u'✂⌨')), [u'✂', u'⌨'])
```

plover/misc.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) 2016 Open Steno Project
+# See LICENSE.txt for details.
+
+
+def popcount_8(v):
+    """Population count for an 8 bit integer"""
+    assert(0 <= v <= 255)
```
benoit-pierre: -> `assert 0 <= v <= 255`

morinted:
Thanks for the test feedback, definitely makes them more meaningful.

PR #541, opened 2016-06-13
Fix x server backspace keycode.

nimble0:
This corrects the keycode for backspace on X server, previously it was set to left control's keycode.


benoit-pierre:
It was set to Capslock, because I use Colemak... My mistake, thanks!

nimble0:
Oh, I made the same mistake, having swapped capslock and control. I didn't realise software rebindings affected the keycode.

PR #542, opened 2016-06-17
Improve macOS keyboard watching support

morinted:
@Achim63 noticed that keyboard watching wasn't working on his system. I believe the notification I was using was not fired in all cases, whereas the legacy Carbon API is always fired.

I also noticed that I forgot to set the visibility of some internal functions and that has been cleaned up.


Achim63:
It works now, z and y don't swap any more when I switch layouts. So I think this can be merged.

Issue #527, opened 2016-05-24
Port look up window to use suggestions window reverse look up


morinted:
## Summary

The look up window should act more like the suggestions window, with the difference being that the look up window operates on text in a text box and the suggestions window operates on the current text buffer.

## Actual Results

The reason this would be useful is that the suggestions window works across cases and also supports suffixes and prefixes, which are very useful. Personal, I don't need the suggestions window open most of the time, having it as the look up window is convenient because I only look up things maybe once or twice a day.

morinted:
Probably a good idea to also pull in the Add Translation dialogue's no-space-and-lowercase default state change

PR #546, opened 2016-06-25
Use alternate suggestions in lookup window

morinted:
Fixes two pet peeves of mine: #527 lookup window tooks things too literally and made finding translations a chore.

Now the lookup and suggestions windows will find text like " Pre-" as "{pre-^}"

Lookup window also can horizontally scroll (needs to be tested for behavior on Linux)

### DIFF
--- a/plover/gui/lookup.py
+++ b/plover/gui/lookup.py
@@ -9,72 +9,52 @@ import plover.gui.util as util
 from plover.translation import escape_translation, unescape_translation
 
 TITLE = 'Plover: Lookup'
-
+CANCEL_BUTTON = 'Close'
 
 class LookupDialog(wx.Dialog):
 
     BORDER = 3
-    TRANSLATION_TEXT = 'Text:'
-    
+
     other_instances = []
-    
+
     def __init__(self, parent, engine, config):
-        pos = (config.get_lookup_frame_x(), 
+        pos = (config.get_lookup_frame_x(),
                config.get_lookup_frame_y())
-        wx.Dialog.__init__(self, parent, title=TITLE, pos=pos)
+        wx.Dialog.__init__(self, parent, title=TITLE, pos=pos,
+                           style=wx.DEFAULT_DIALOG_STYLE|wx.RESIZE_BORDER)
 
         self.config = config
 
         # components
         self.translation_text = wx.TextCtrl(self, style=wx.TE_PROCESS_ENTER)
-        cancel = wx.Button(self, id=wx.ID_CANCEL, label='Cancel')
-        self.listbox = wx.ListBox(self, size=wx.Size(210, 200), style=wx.LB_HSCROLL)
-        
+        self.listbox = wx.ListBox(self, size=wx.Size(210, 200),
+                                  style=wx.LB_HSCROLL)
+        cancel = wx.Button(self, wx.ID_CANCEL, CANCEL_BUTTON)
+        cancel.Bind(wx.EVT_BUTTON, self.on_close)
         # layout
         global_sizer = wx.BoxSizer(wx.VERTICAL)
-        
-        sizer = wx.BoxSizer(wx.HORIZONTAL)
-        label = wx.StaticText(self, label=self.TRANSLATION_TEXT)
-        sizer.Add(label, 
-                  flag=wx.TOP | wx.LEFT | wx.RIGHT | wx.BOTTOM | wx.ALIGN_CENTER_VERTICAL, 
-                  border=self.BORDER)
-        sizer.Add(self.translation_text, 
-                  flag=wx.TOP | wx.RIGHT | wx.BOTTOM | wx.ALIGN_CENTER_VERTICAL, 
-                  border=self.BORDER)
-        sizer.Add(cancel, 
-                  flag=wx.TOP | wx.RIGHT | wx.BOTTOM | wx.ALIGN_CENTER_VERTICAL, 
-                  border=self.BORDER)
-        global_sizer.Add(sizer)
-        
-        sizer = wx.BoxSizer(wx.HORIZONTAL)
-        sizer.Add(self.listbox,
-                  flag=wx.ALL | wx.FIXED_MINSIZE,
-                  border=self.BORDER)
 
-        global_sizer.Add(sizer)
-        
+        global_sizer.Add(self.translation_text, 0, wx.ALL | wx.EXPAND,
+                         self.BORDER)
+        global_sizer.Add(self.listbox, 1, wx.ALL & ~wx.TOP | wx.EXPAND,
+                         self.BORDER)
+        global_sizer.Add(cancel, 0, wx.ALL & ~wx.TOP | wx.EXPAND, self.BORDER)
+
         self.SetAutoLayout(True)
         self.SetSizer(global_sizer)
         global_sizer.Fit(self)
         global_sizer.SetSizeHints(self)
         self.Layout()
         self.SetRect(AdjustRectToScreen(self.GetRect()))
-        
-        # events
 
-        # The reason for the focus event here is to skip focus on tab traversal
-        # of the buttons. But it seems that on windows this prevents the button
-        # from being pressed. Leave this commented out until that problem is
-        # resolved.
-        #button.Bind(wx.EVT_SET_FOCUS, self.on_button_gained_focus)
-        cancel.Bind(wx.EVT_BUTTON, self.on_close)
-        #cancel.Bind(wx.EVT_SET_FOCUS, self.on_button_gained_focus)
+        # events
         self.translation_text.Bind(wx.EVT_TEXT, self.on_translation_change)
         self.translation_text.Bind(wx.EVT_SET_FOCUS, self.on_translation_gained_focus)
         self.translation_text.Bind(wx.EVT_KILL_FOCUS, self.on_translation_lost_focus)
         self.Bind(wx.EVT_CLOSE, self.on_close)
+
         self.Bind(wx.EVT_MOVE, self.on_move)
-        
+
         self.engine = engine
         
         # TODO: add functions on engine for state
--- a/plover/gui/lookup.py
+++ b/plover/gui/lookup.py
@@ -6,10 +6,10 @@ from wx.lib.utils import AdjustRectToScreen
 
 from plover.steno import normalize_steno
 import plover.gui.util as util
-from plover.translation import unescape_translation
-
+from plover.translation import escape_translation, unescape_translation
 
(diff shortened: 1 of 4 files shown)


benoit-pierre:
That's great!

How about moving the code to it's own suggestions module? Right now it's still pretty basic, but if we gradually improve it, I think it deserves it's own module.

I get a traceback when I clear the pattern in the lookup dialog (with backspace or `*`) because the list is empty and we try to scroll to an invalid index:

```
  File "/home/bpierre/progs/src/plover/plover/gui/lookup.py", line 121, in on_translation_change
    self.listbox.EnsureVisible(0)
  File "/usr/lib64/python2.7/site-packages/wx-3.0-gtk2/wx/_controls.py", line 1231, in EnsureVisible
    return _controls_.ListBox_EnsureVisible(*args, **kwargs)
wx._core.PyAssertionError: C++ assertion "IsValid(n)" failed at ./src/gtk/listbox.cpp(744) in DoScrollToCell(): invalid index
```

The horizontal scrolling behavior of the lookup window is fine, but we should really fix how the layout behave on resizing.

morinted:
@benoit-pierre I tried my hand at a class here... let me know what you think.

benoit-pierre:
I think the suggestions instance should live in the engine, no point in putting it in the translator.

plover/suggestions.py
```diff
@@ -0,0 +1,52 @@
+import collections
+
+Suggestion = collections.namedtuple('Suggestion', 'text steno_list')
+
+
+class Suggestions(object):
+    def __init__(self, dictionary):
+        self.dictionary = dictionary
+
+    def list(self, translation):
```
benoit-pierre: Could another name be used? Shadowing a builtin can lead to horrible bugs.

plover/suggestions.py
```diff
@@ -0,0 +1,52 @@
...
+
+        for t in possible_translations:
+            for modded_translation in [mod % t for mod in mods]:
+                strokes_list = self.dictionary.reverse_lookup(modded_translation)
+                if not strokes_list:
+                    continue
+                # Return suggestions, sorted by fewest strokes, then fewest keys
+                strokes_list = sorted(
+                    strokes_list,
+                    key=lambda x: len(x) * 32 + sum(map(len, x))
```
benoit-pierre: Kinglet has 38 keys, how about using this instead:
```python3
key=lambda x: (len(x), sum(map(len, x)))
```

benoit-pierre:
Resizing now works fine, and so does clearing the text field. :+1:

plover/gui/lookup.py
```diff
@@ -106,17 +85,17 @@ def on_close(self, event=None):
...
+        translation = event.GetString()
         self.listbox.Clear()
         if translation:
-            d = self.engine.get_dictionary()
-            strokes_list = d.reverse_lookup(unescape_translation(translation))
-            if strokes_list:
-                entries = ('/'.join(x) for x in strokes_list)
-                for str in entries:
-                    self.listbox.Append(str)
+            suggestions = self.engine.get_suggestions().list(translation)
```
benoit-pierre: The engine API could be simplified to be used like this:
```python
suggestions = self.engine.get_suggestions(translation)
```

plover/suggestions.py
morinted: Is this shadowing? It will only be accessible by self.list and on instances. Curious if there's a case where you'd actually get a conflict with the built in.

benoit-pierre: And for the rest of the class declaration:
```diff
diff --git i/plover/suggestions.py w/plover/suggestions.py
--- i/plover/suggestions.py
+++ w/plover/suggestions.py
@@ -50,3 +50,7 @@ class Suggestions(object):
                 suggestions.append(suggestion)

         return suggestions
+
+    CONSTANT = collections.defaultdict(list)
+
+print Suggestions.CONSTANT['foo']
```
```shell
> python2 -m plover.suggestions
  File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/bpierre/progs/src/plover/plover/suggestions.py", line 56, in <module>
    print Suggestions.CONSTANT['foo']
TypeError: list() takes exactly 2 arguments (0 given)
```

morinted: That makes sense, thanks!

morinted:
Okay, updated once more with your feedback.

benoit-pierre:
OK, what about moving the instance out of the `Translator`? Adding it as a member there, only to be used externally does not make sense. Keeping it in the `Engine` does: it can be updated automatically when the list of dictionaries changes, it will make adding locking to the `Engine` easier.

morinted:
Yes, that makes sense. Moved to the engine…now, is it still worth wrapping the suggestions in engine.get_suggestions when otherwise it would be engine.suggestions.find? Alternatively I could make the suggestions member private.

benoit-pierre:
` Moved to the engine…now, is it still worth wrapping the suggestions in engine.get_suggestions when otherwise it would be engine.suggestions.find?

Yes, I think it is: this will make locking concurrent accesses to the engine easier.

benoit-pierre:
OK, one last thing (promised ;)), a regression:

```diff
diff --git i/plover/gui/lookup.py w/plover/gui/lookup.py
index 4bd84e7..6c7ad01 100644
--- i/plover/gui/lookup.py
+++ w/plover/gui/lookup.py
@@ -6,6 +6,7 @@ from wx.lib.utils import AdjustRectToScreen

 from plover.steno import normalize_steno
 import plover.gui.util as util
+from plover.translation import unescape_translation


 TITLE = 'Plover: Lookup'
@@ -88,7 +89,7 @@ class LookupDialog(wx.Dialog):
         translation = event.GetString()
         self.listbox.Clear()
         if translation:
-            suggestions = self.engine.get_suggestions(translation)
+            suggestions = self.engine.get_suggestions(unescape_translation(translation))
             if suggestions:
                 for text, strokes in suggestions:
                     self.listbox.Append(text)
```

Additionally, we should fix the display in both the lookup and suggestions dialog to call `escape_translation` before presenting a suggestion to the user. But this can be done in a follow-up PR if you want.

plover/gui/lookup.py
```diff
@@ -106,17 +85,17 @@ def on_close(self, event=None):
...
-            if strokes_list:
-                entries = ('/'.join(x) for x in strokes_list)
-                for str in entries:
-                    self.listbox.Append(str)
+            suggestions = self.engine.get_suggestions(translation)
+            if suggestions:
+                for text, strokes in suggestions:
+                    self.listbox.Append(text)
+                    entries = ('/'.join(x) for x in strokes)
+                    for str in entries:
```
benoit-pierre: `str` => `translation` to avoid shadowing a builtin.

morinted:
Updated the PR with the escaping. Had to also do a check to ensure that the stripped translation wasn't empty because `"\n".strip() == ""`

```diff
--- a/plover/gui/lookup.py
+++ b/plover/gui/lookup.py
@@ -6,7 +6,7 @@ from wx.lib.utils import AdjustRectToScreen

 from plover.steno import normalize_steno
 import plover.gui.util as util
-
+from plover.translation import escape_translation, unescape_translation

 TITLE = 'Plover: Lookup'

@@ -109,13 +109,15 @@ class LookupDialog(wx.Dialog):
         translation = event.GetString()
         self.listbox.Clear()
         if translation:
-            suggestions = self.engine.get_suggestions(translation)
+            suggestions = self.engine.get_suggestions(
+                unescape_translation(translation)
+            )
             if suggestions:
-                for text, strokes in suggestions:
-                    self.listbox.Append(text)
+                for suggestion, strokes in suggestions:
+                    self.listbox.Append(escape_translation(suggestion))
                     entries = ('/'.join(x) for x in strokes)
-                    for str in entries:
-                        self.listbox.Append('    %s' % (str))
+                    for entry in entries:
+                        self.listbox.Append('    %s' % (entry))
                 self.listbox.EnsureVisible(0)
             else:
                 self.listbox.Append('No entries')
diff --git a/plover/gui/suggestions.py b/plover/gui/suggestions.py
index 2b424f5..61595a2 100644
--- a/plover/gui/suggestions.py
+++ b/plover/gui/suggestions.py
@@ -9,6 +9,7 @@ from wx.lib.utils import AdjustRectToScreen
 from plover.gui.util import find_fixed_width_font, shorten_unicode
 from plover import system
 from plover.suggestions import Suggestion
+from plover.translation import escape_translation

 PAT = re.compile(r'[-\'"\w]+|[^\w\s]')
 TITLE = 'Plover: Suggestions Display'
@@ -142,7 +143,9 @@ class SuggestionsDisplayDialog(wx.Dialog):

         for suggestion in suggestion_list:
             self.listbox.SetDefaultStyle(self.word_style)
-            self.listbox.WriteText(shorten_unicode(suggestion.text) + u'\n')
+            self.listbox.WriteText(
+                shorten_unicode(escape_translation(suggestion.text)) + u'\n'
+            )
             if not suggestion.steno_list:
                 self.listbox.SetDefaultStyle(self.no_suggestion_style)
                 self.listbox.WriteText(self.no_suggestion_indent)
diff --git a/plover/suggestions.py b/plover/suggestions.py
index 704157c..38b1ccd 100644
--- a/plover/suggestions.py
+++ b/plover/suggestions.py
@@ -25,7 +25,7 @@ class Suggestions(object):
         possible_translations = set([translation])

         stripped_translation = translation.strip()
-        if stripped_translation != translation:
+        if stripped_translation and stripped_translation != translation:
             possible_translations.add(stripped_translation)

         lowercase_translation = translation.lower()
```

plover/suggestions.py
```diff
@@ -0,0 +1,52 @@
...
+            u'{^}%s{^}',
+            u'{%s^}',  # Suffix
+            u'%s{^}',
+            u'{&%s}',  # Fingerspell
+            u'{#%s}',  # Command
+        ]
+
+        possible_translations = set([translation])
+
+        stripped_translation = translation.strip()
```
benoit-pierre: The problem is entering `\n` in the lookup dialog will still match: `"KPA*": "{^}{-|}"`, maybe we should only strip spaces?
```python
        # Only strip spaces, so patterns with \n or \t are correctly handled.
        stripped_translation = translation.strip(' ')
```

morinted: I think `\n` stripped is `''` so it wouldn't get added to the possible translations set. However, I do agree that `.strip(' ')` is the proper way to handle this. Will update.

PR #553, opened 2016-06-30
formatting: handle escape characters when creating atoms

morinted:
Found that entries with \n \r \t were not doing anything while working on my machine, made a failing test and fixed it.


morinted:
@benoit-pierre so should we consider changing *these* strips to `strip(' ')`? I'm all for `{^}\t{^}` or `Mr.\u00a0{-|}{^}`, to me it's unexpected if we override that behavior silently.

morinted:
Since it seems the RTF parser delimits translation boundaries with newlines, I think this is good behavior to allow. Updated to only strip spaces.

PR #555, opened 2016-07-01
Improve default dictionaries copy

benoit-pierre:
Only copy a dictionary if it's in use by the current config and missing. The code also now use `shutil.copy` so different formats are supported without the need to know the internals of those formats.

Finally, I updated the `main.json` and `commands.json` assets with the result of saving with Plover: normalized strokes, non-escaped Unicode characters, and sorted entries.

Note: the `Posurdex` entry had an invalid trailing escaped Unicode character I stripped, but maybe we should take this opportunity to remove some of those medical entries?


morinted:
- Looks like the Posurdex character wasn't removed.
- As for removing the medical terms, I think that overall they are not useful but for students learning medical terminology they are... maybe we should extract them into a dictionary for Stenodict instead of having them in main. We could extract rare words from the main dictionary, maybe it's time to write a script that searches the words in Google and looks at # of results to gauge popularity? Posurdex: 4,600 results; acetaminophen: 10,100,000 results. Not sure what a valid threshold would be but <5,000 seems pretty darn niche.
- Would you mind also fixing the `"TA*BT": "{#Alt_L(Tab(Tab))}",` to `"TA*BT": "{#Alt_L(Tab Tab)}",` ?

benoit-pierre:
> - Looks like the Posurdex character wasn't removed.

Right, got confused by the diff output (since the character is invalid and does not show). It's removed now.

> - As for removing the medical terms, I think that overall they are not useful but for students learning medical terminology they are... maybe we should extract them into a dictionary for Stenodict instead of having them in main. We could extract rare words from the main dictionary, maybe it's time to write a script that searches the words in Google and looks at # of results to gauge popularity? Posurdex: 4,600 results; acetaminophen: 10,100,000 results. Not sure what a valid threshold would be but <5,000 seems pretty darn niche.

I think a dictionary without miss-strokes, medical terms / domain specific entries, is better for people starting with Plover. If the student is a beginner in a particular field, it make more sense for him to add the relevant Stenodict dictionary instead of having those entries as part of the default main dictionary.

> - Would you mind also fixing the `"TA*BT": "{#Alt_L(Tab(Tab))}",` to `"TA*BT": "{#Alt_L(Tab Tab)}",` ?

Done.

Issue #482, opened 2016-04-25
Linux: Send events for modifiers


morinted:
Classification: Enhancement

## Summary

Some applications don't properly receive modifier keys from Plover, likely because they handle shifted states and the like on their own. For example:

- Gnome terminal (Ubuntu)
- Konsole
- KRunner

e.g. `Email: someone@example.com` would come out as `email; someone2example.com`

Plover version: Tested with 2.5.8 & current HEAD

benoit-pierre:
* bad news: sending additional press/release events for modifiers with `send_events` does not work
* good news: switching to XInput and using XTest for sending all events work

Issue #556, opened 2016-07-12
Capitalization after a period doesn't work


Geremia:
Classification: Bug

Reproducibility: Always

## Summary

I can't get capitalization after a period to work. I put a period+space, and it doesn't auto-capitalize the first letter of the next word after the period+space. It doesn't matter if I check "Start Capitalized" in the prefs or not.

## Steps to Reproduce

{.} is my entry. I tried {^.}{-|}, too, but to no avail.

Also, Dvorak is set as my system's default keyboard layout. I changed it to the regular US layout. Still, it doesn't auto-capitalize.

## Version

Plover version weekly-v3.0.0+80.gdc7ccc3
Installed via GitHub source of weekly-v3.0.0+80.gdc7ccc3

## Notes

Describe any debugging steps you've taken yourself.

If you've found a workaround, provide it here.

## Configuration

OS:

- If it's OS X,
  go to Finder, Apple menu, "About This Mac".
  - Alternatively, if you're comfortable with the Terminal, run
    `sw_vers` and paste the output.

- Linux,
  Slackware-current (14.2)
  relevant to Plover's operation: KDE Plasma
**I do not get this bug running XFCE4.**
  - uname -a: Linux alan 4.4.14 #2 SMP Fri Jun 24 13:38:27 CDT 2016 x86_64 Intel(R) Core(TM) i3 CPU       U 330  @ 1.20GHz GenuineIntel GNU/Linux

I've satisfied these reqs from requirements.txt:

> pytest
> setuptools
> pyserial>=2.6
> appdirs>=1.3.0
> hidapi
> mock
> pyobjc-core>=3.0.3; "darwin" in sys_platform
> pyobjc-framework-Cocoa>=3.0.3; "darwin" in sys_platform
> pyobjc-framework-Quartz>=3.0.3; "darwin" in sys_platform
> wxPython>=3.0; "darwin" in sys_platform
> python-xlib>=0.14; "linux" in sys_platform
> wxPython>=3.0; "linux" in sys_platform
> pywin32>=219; "win32" in sys_platform

benoit-pierre:
Qt5 apps don't take into account the modifiers state in our simulated key press/release events, can you confirm this only happen with Qt5 applications?

Geremia:
@benoit-pierre It only happens when I use the KDE Plasma window manager, which is Qt5. I don't get the issue in other WMs like XFCE4. This issue is indeed specific to KDE Plasma.

[...]

Issue #520, opened 2016-05-12
Linux: when the machine is set to keyboard, using a key combo like "{#a b c}" does not work


benoit-pierre:
Classification: Bug

Reproducibility: Always

## Summary

It's not possible to use (no modifiers) suppressed keys in a key combo when the machine is set to keyboard.

## Steps to Reproduce

1. set the machine to keyboard, standard layout
1. add a brief whose translation is "#{a b c}"
1. focus a text field
1. stroke the brief

## Expected Results

1. "abc" appears in the text field

## Actual Results

1. no text is output

## Version

Plover version 3.0.0.

## Notes

* the problem is we use `xtest.fake_input` to simulate key events, and resulting events for keys that are suppressed will be filtered-out  by the X11 server (when no modifier other than`NumLock` is active)
* switching to `focus_window.send_event` won't work for global shortcuts (e.g. the window manager shortcuts)
* mix-matching the use of both `xtest.fake_input` and `focus_window.send_event` might be possible

PR #558, opened 2016-07-16
Improve Linux keyboard support

benoit-pierre:
### Switch to XInput

Use XInput to capture keyboard events; this allow filtering-out the XTest keyboard, and only capture other keyboards when suppressing; which in turn will make it possible to use XTest for all simulated events (and not just for sending key combos).

### Use XTest for all simulated key events

This improve compatibility with with Qt5 applications (modifiers are now correctly handled), window managers (e.g. Awesome run prompt), Xterm (it's not necessary to use allowSendEvents anymore), ...

### DIFF
--- a/plover/oslayer/xkeyboardcontrol.py
+++ b/plover/oslayer/xkeyboardcontrol.py
@@ -29,7 +29,6 @@ import threading
 from Xlib import X, XK, display
 from Xlib.ext import xinput, xtest
 from Xlib.ext.ge import GenericEventCode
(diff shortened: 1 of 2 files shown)


benoit-pierre:
Note: this fixes #482, #520, and should probably take care of #556 too.

benoit-pierre:
Another nice improvement with this PR: no more transient focus out/in events with keyboard suppression! Which mean the "Add translation" code could be simplified a little and it's now possible to edit entries in the dictionary editor with Plover on when using the keyboard. `\o/`

PR #562, opened 2016-07-26
Minor dictionary fixes, spaces before/after words

percidae:
There are several dictionary entries which have a leading or trailing space, which doesn't make sense. 
This commit fixes a few entries.

Also addresses #400


benoit-pierre:
LGTM.

PR #563, opened 2016-07-27
translation: tweak Translation.str/repr format

benoit-pierre:
- change output so it's unambiguous (None vs "None")
- escape \n, \r, and \t (prevent line breaks in strokes.log entries)

### DIFF
--- a/plover/translation.py
+++ b/plover/translation.py
@@ -101,7 +101,12 @@ class Translation(object):
         return not self.__eq__(other)
 
     def __str__(self):
-        return 'Translation(%s : %s)' % (self.rtfcre, self.english)
+        if self.english is None:
+            translation = 'None'
+        else:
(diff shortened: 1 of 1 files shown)


plover/translation.py
```diff
@@ -101,7 +101,12 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
     def __str__(self):
-        return 'Translation(%s : %s)' % (self.rtfcre, self.english)
+        if self.english is None:
+            translation = 'None'
+        else:
+            translation = escape_translation(self.english)
+            translation = '"%s" ' % translation.replace('"', r'\"')
```
morinted: Is the trailing space in `'"%s" '` intentional?

benoit-pierre: Nope!

PR #564, opened 2016-07-28
oslayer/winkeyboardcontrol: fix capture process termination

benoit-pierre:
The heartbeat thread needs to be stopped before exit, or with Python 3 the process will deadlock at exit trying to join it.


morinted:
Looks good, can you rebase for merge?

benoit-pierre:
Done.

PR #567, opened 2016-07-31
steno_dictionary: improve reverse lookup

benoit-pierre:
Don't stop at the first dictionary with matches: return all possible strokes in all dictionaries (ignoring strokes overridden in higher precedence dictionaries).


morinted:
Ah, I've been hankering for this sort of functionality since I started adding my own definitions for words. So many thanks.

PR #569, opened 2016-08-06
Changed 'E. Coli' and 'E coli' entries to 'E. coli'

percidae:
This pull request addresses the different writings of "E. coli" in the dictionary and unifies them.
#400


morinted:
Looks good to me, thanks for the consistency change.

PR #570, opened 2016-08-06
Last of the changes needed for core Python 3 support

benoit-pierre:
Note:

* configuration: the changes include making sure the configuration file is saved as UTF-8, and string options are returned as Unicode
* RTF/CRE dictionaries: for lack of better RTF file format support, assume the encoding of RTF/CRE dictionaries is cp1252
* Stentura: we can't just replace the use of `buffer` with `memoryview` in Python 3 because it does prevent us from resizing the original `bytearray` when needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_install:
     pip --disable-pip-version-check --timeout=5 --retries=2 install --upgrade --user 'setuptools>=19.6' &&
     # Manually install Cython if not already to speedup hidapi build.
     pip --disable-pip-version-check --timeout=5 --retries=2 install --user Cython &&
-    pip --disable-pip-version-check --timeout=5 --retries=2 install --upgrade --user -r requirements.txt &&
+    pip --disable-pip-version-check --timeout=5 --retries=2 install --upgrade --user -r requirements.txt -c requirements_constraints.txt &&
     # And now downgrade setuptools so py2app works correctly...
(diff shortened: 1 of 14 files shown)


benoit-pierre:
@morinted: shall the change to the call to `CGEventKeyboardSetUnicodeString` be included? It is compatible with an un-patched Python 2 version of pyobjc-core, right?

morinted:
Yes, it works with unpatched Python2 but breaks Python3 output. If we just leave it, it only causes a crash when you use extended characters on python3. I'm not sure whether we should change it now or when we get a patched version of PyObjc.

benoit-pierre:
I made a [build](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_2) with the change (including a patched pyobjc-core): can you check if it works?

morinted:
The Python 3 build works and outputs emoji, yup 👍 

But the wx Py2 version doesn't launch. Gives me some ambiguity errors: https://gist.github.com/morinted/17a44aa4000f72920819d4f840dadbb7 is this because you are using PyInstaller instead of py2app?

benoit-pierre:
Yes, that's probably because of PyInstaller.

benoit-pierre:
Updated to include the fix for full Unicode support with Python 3 on OSX.

morinted:
You've tested the Stentura changes with users on Discord, right?

benoit-pierre:
Only one user.

morinted:
Okay, a weekly will be necessary soon.

PR #571, opened 2016-08-07
[RFC] new UI

benoit-pierre:
The goal of this PR is get feedback on the new UI rewrite.

### Test builds

Note: the new UI is available for the Python 3 builds on Windows and OS X. ~~For Linux, the Debian package still uses Python 2, and provide both the (old) WX UI and the new PyQt5 based UI, launch Plover with `plover -g qt` to select the new UI (and of course Python 3 is supported if you run from source).~~ The Debian package now uses Python 3 too, and the new UI is automatically selected.

* [ui_preview_1](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_1)
* [ui_preview_2](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_2)
* [ui_preview_3](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_3)
* [ui_preview_4](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_4)
* [ui_preview_5](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_5)
* [ui_preview_6](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_6)

### Major differences

* The engine uses a dedicated thread.
* The UI is localized: for now only full localization for French is available. For testing purpose, the `LANGUAGE` environment variable can be set to override the locale detection. Note that part of some dialogs will already be automatically localized thanks to Qt own localizations (e.g. standard OK/Cancel buttons labels), or by the OS (e.g. on Windows a native file browser is used).
* window states (size and position, fonts used) are saved using Qt settings' API (instead of Plover's configuration); rational: make it easier to use the same configuration on different devices (using different fonts for different display DPI..., window sizes, etc...).

### Known issues

* ~~[OS X] there's no differentiation between a right and left click on the tray icon.~~
* ~~[OS X] on quitting with Ctrl+Q, the warning about closing the main window is shown.~~
* [OS X] the default font in the paper tape is not monospaced (not a major issue as it's now user configurable).
* [Windows] one report of various crashes when using the 64bits build on Windows 7 (e.g. when re-ordering in the dictionaries manager).
* [Linux] the tray icon does not work right on Ubuntu 14.04 Trusty (the PyQt5 version available is really old).
* [Linux] minimize on start does not work when tray icon support is not available

## Main window

main_window

Menu layout:
```
File
  Toggle output
  Reconnect machine
  -----------------
  Configure
  -----------------
  Show/Hide
  Quit Plover
Tools
  New translation
  Lookup
  Manage dictionaries
  -------------------
  Paper tape
  Suggestions
Help
  About
```

## Tray icon

The tray icon is automatically enabled when supported.

3 different state icons (reused from #263, as I'm not much of an artist, I hope it's all right @gcr) are used:
* machine disconnected:

state-disconnected
* machine connected, output disabled:

state-disabled
* machine connected, output enabled:

state-enabled

Context menu layout:
```
Toggle output
Reconnect machine
-----------------
Tools
  New translation
  Lookup
  Manage dictionaries
  -------------------
  Paper tape
  Suggestions
-----------------
Configure
-----------------
Help
  About
-----------------
Show/Hide
Quit Plover
```

## Configuration

configuration

## Dictionaries Manager

dictionaries_manager

Note:
* I kept the original ordering, about flipping it: highest priority (user) dictionary on top? This way the top (visible on open) of the list will contain the dictionaries the user is more likely to interact with, and the list will match the order the dictionaries are searched.
* Drag & dropping a dictionary file from an external file manager on to the list should add it to Plover.

## Dictionary Editor

dictionary_editor

## Add Translation

add_translation

## Lookup

lookup

## Paper Tape

paper_tape

## Suggestions

suggestions

## About

about


benoit-pierre:
The behavior of the old UI with regard to focusing back the previous window has been kept when a dialog was triggered through a translation command (for lookup, add translation, and configure), but I really think it should be possible to disable it. IMHO it's not needed on Linux; as any good window manager already does what's right, and our code as some annoying side effects (like raising the previous window when focusing it back, which can result in some previous visible windows getting hidden), how about a new interface setting?

Similarly, remembering window positions/sizes can be annoying when you want to use your window manager facilities to automatically layout Plover's windows in a set way.

morinted:
I noticed today that I can't write emoji into the Add Translation dialog box on Mac with Plover, but I can paste them in. Can you reproduce this on Windows or Linux?

benoit-pierre:
It works fine on Linux with `"SHR*UG/SHR*UG": "¯\\_(ツ)_/¯"`, but not with my Windows 10 VM...

benoit-pierre:
New release: [ui_preview_2](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_2): include a fixed pyobjc-core version so emojis work in OS X.

morinted:
So "¯\_(ツ)_/¯" works on OS X, but the emoji don't register at all. Also of
note, I can't get the character viewer to show up:
https://support.apple.com/en-ca/HT201586

This makes me think that maybe QT is missing a flag or something to allow
this behavior. On Windows it might just not handle attaching a character to
a string like we found with older applications.

benoit-pierre:
How about when running from source?

morinted:
The behavior is the same when running from source.

benoit-pierre:
OK, so the build is not missing a library.

morinted:
This is ancient: https://bugreports.qt.io/browse/QTBUG-8979

morinted:
Maybe we can also get @gcr to make some visually simpler tray icons, like a simplified steno layout so that it isn't as cramped as the full layout in the tray. I did this once before for a simplified hardware button on the open steno site: http://www.openstenoproject.org/images/icon-hardware-2x.png

Ideally the icons would be square as well but I acknowledge that that's a lot to ask 😁

benoit-pierre:
Yeah, I also wonder how they look on a white background: needs some border on all elements.

SyntaxBlitz:
Pretty bad! 
image

This, when I hold down my mouse on the icon: 
image

morinted:
In the machine configuration, the machine options are under an "options"
section. Would it be possible to simply place the options in the top-level
of the machine config? Something about the nesting looks particularly off
on Mac. I don't think the grouping is particularly valuable either since
the only other option is the key map which is an option anyway. And the
Tréal just shouldn't show anything other than the key map since it has no
options.

morinted:
I can get the character viewer to work on Mac by including an empty "Edit" menu entry next to File, Dictionaries, and Tools. This is discussed in this QT issue: https://bugreports.qt.io/browse/QTBUG-40514?jql=text%20~%20%22character%20viewer%22%20ORDER%20BY%20created%20DESC

Unfortunately I still can't write emoji with Plover into the application, but at least I can bring up the character selector to select them.

There are also many good human interface guidelines that I'll probably have to peruse in order to improve Plover's behavior according to Mac standards.

morinted:
Here's what it looks like:

edit menu

Of note, the edit menu doesn't show until OS X attaches the emoji & symbols item (about 1 to 2 seconds after start up)

morinted:
It's also strange to me how the output check box is just hanging out in the center of the window. I'm not quite sure what the solution is…but because the UI is somewhat malleable you can get into a weird state like:

I'm splitting hairs

I'm downloading the QT creator so I'll have a chance to play with some of this stuff.

morinted:
When editing dictionaries, the dictionary editor opens inline, and shares the dictionary manager's window. This means that the only way to close the dictionary editor is with the escape key.

screen shot 2016-08-08 at 3 00 52 pm
screen shot 2016-08-08 at 3 01 08 pm

morinted:
I dislike the gradient given to the toolbar all over the place. I've implemented a workaround.

- All the windows have default 12 padding which looks very big on OS X. I think a better default would be 6 if possible.

Next, for the toolbar:

```diff
diff --git a/plover/gui_qt/utils.py b/plover/gui_qt/utils.py
index b36273b..e68a0c0 100644
--- a/plover/gui_qt/utils.py
+++ b/plover/gui_qt/utils.py
@@ -1,12 +1,15 @@

-from PyQt5.QtCore import QSettings
+from PyQt5.QtCore import QSettings, QSize
 from PyQt5.QtWidgets import (
     QMainWindow,
     QMenu,
     QToolBar,
     QToolButton,
+    QStyleFactory,
 )

+import sys
+

 def ToolButton(action):
     button = QToolButton()
@@ -18,6 +21,9 @@ def ToolBar(*action_list):
     toolbar = QToolBar()
     for action in action_list:
         toolbar.addWidget(ToolButton(action))
+    if sys.platform.startswith('darwin'):
+        toolbar.setStyle(QStyleFactory.create('windows'))
+        toolbar.setIconSize(QSize(23, 16))
     return toolbar
```

screen shot 2016-08-08 at 6 20 04 pm

morinted:
I've made some changes for OS X, all of them bundled into this commit: https://github.com/morinted/plover/commit/933784ea7f4e801265feee4bcb498d5c76c801af

### Icons

- I changed to an SVG icon pack -- https://materialdesignicons.com/ which are MIT licensed (which is OK to use in our project, I believe since GPL is a *more* restrictive license) I think these are a little more modern than the set used previously and I gave some thought as to which icons to use.

To do that required a loader, which renders the icons at 2x the resolution required (Mac need 2x for retina, which is activated for Pixmaps with `self._app.setAttribute(Qt.AA_UseHighDpiPixmaps)`

screen shot 2016-08-08 at 10 43 32 pm
screen shot 2016-08-08 at 10 59 34 pm

- You can see the pin in/out icon change

screen shot 2016-08-08 at 10 59 39 pm
screen shot 2016-08-08 at 10 59 43 pm

- I also implemented coloring the SVG in case we decided to use different colors (I noticed the menu icons are black which isn't ideal)

screen shot 2016-08-08 at 11 01 29 pm

### Tray Icon Mask

In OSX you can made the icon look better on the light-bar by setting: `icon.setIsMask(True)`

### Padding

The padding has been adjusted for what I think looks good on Mac.

-------

The commit is really messy. Just let me know what you think before we continue.

stenoknight:
That new icon pack is a huge improvement!!! It looks really, really nice.

morinted:
I was thinking, too, that the icon color could be configurable for users who have dark themes. Not sure if we can decouple the icon color in menus from the one in the UI but that would also be useful on macOS

morinted:
The icons above are no good because they don't work well on light/dark themes. I really like icons8, but they are non-free and so even though they "support" open source projects, I don't see how we can release them with a GPL-licensed Plover.

Therefore, I'm going to try and use their MIT-licensed icons instead, and create new icons based off of those: https://github.com/icons8/flat-color-icons

It will take a couple days for me to get them to a point that I'm happy. In the meanwhile, here are a couple of them in use (I'm going to be adjusting some of them and swapping out others…)

screen shot 2016-08-10 at 3 25 13 pm

morinted:
1. Would it be possible for the machine options to be inlined instead of nested? Or is that two tough because one is a list and the other a grid?
2. In the dictionary editor, you cannot undo after deleting one entry → you can undo after deleting two entries, which then allows you to undo the first deletion as well.
3. Continuing to work with the icons, it feels good working with the MIT and GPL-licensed icons—I don't feel guilty for modifying and morphing them into something new. Here is what I've got so far, what's left is keep on top, choose font, clear, and save:
plover_icons

benoit-pierre:
> Would it be possible for the machine options to be inlined instead of nested? Or is that two tough because one is a list and the other a grid?

No, the machine options are one widget. Changing that would break the whole "programatically create/layout the configuration dialog".

> In the dictionary editor, you cannot undo after deleting one entry → you can undo after deleting two entries, which then allows you to undo the first deletion as well.

Fixed.

benoit-pierre:
New preview: [ui_preview_3](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_3)

morinted:
Tray icon attempts:

screen shot 2016-08-12 at 9 45 38 am

--------------------

I've completed my icon set:

screen shot 2016-08-12 at 9 47 54 am

--------------------

I was brain storming as to what we could do with the main window, if we decide to drop the minimalist interface in favor of the tray icon's menu: *please excuse poor proportions and penmanship*

screen shot 2016-08-12 at 9 38 46 am

> The machine drop down lets you change between configured machines, but the settings gear takes you to that machine's configuration in case it doesn't connect right away. The dictionary manager is below. The priority dictionary has a star next to it to say that it has priority (probably user configurable since "real" steno software always has things ordered opposite)
>
> The up/down arrows are there in case the user doesn't want to drag-and-drop to reorder.

lifton:
Great to see so much work on the UI. I'm all for whatever you decide, but I'm curious about the move from wx to QT, mostly for my own edification in case I want to make a similar move for other projects. Can someone point me to the thread where this was discussed?

benoit-pierre:
@lifton: I have used three different widget libraries for Python:
- GTK 2/3: PyGtk and PyGObject
- Qt: PyQt5
- wxWidgets: wxpython

Based on my experience, here is how I would rate them: Qt > GTK > wxWidgets.

Qt is better all around: Python 2 and 3 support, better documentation/examples, better tools, portable. Easier multi-threading support with its signals/slots system. Additionally, the Python 3 version can now also be installed through pip (at least on Windows and Linux). Note: that's when rating QWidgets, not QML's QtQuick, which is in my experience not as good: less mature, documentation/examples not as good, requires native graphic acceleration support (can't run without OpenGL on Linux, so not possible to test with Xephyr/Xnest)...

GTK is pretty good too. Documentation is not as good. Portability is good in my experience (but I've not used it on Mac OS X, so I don't know how it fares on that platform).

wxWidgets is the less portable alternative: the fact that it uses a different native toolkit on each platform can lead to differences in behavior; because each widget is (re-)implemented on top of a different native toolkit (unlike Qt/GTK, where there's one implementation on top of a (smaller) platform abstraction layer). It's also missing (official) Python 3 support.

lifton:
@benoit-pierre Thanks for the informative run down!

morinted:
When I ran with an unrecognized machine type, Plover gives the "unrecognized machine type" exception twice and then segfaults.

morinted:
Placed my icons online: https://tedmor.in/plover_icons

gcr:
I really like @morinted's tray icons. Much simpler than trying to show the entire steno machine, but the key design is still iconic.

morinted:
> TypeError: on_lookup() got an unexpected keyword argument 'manage_windows'
> Abort trap: 6

With UI preview 4 the lookup window causes a crash.

morinted:
Suggestion for what to put above the dictionary manager, dropping the machine connected status bar at the bottom:

screen shot 2016-08-25 at 2 49 50 pm

benoit-pierre:
New preview: [ui_preview_4](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_4)

morinted:
I've exported the above tray icons in svg format, I did a set for Mac which is black and white and a normal set in color with light outlines.

For the black-and-white Mac version, we'd want to use mask. For Windows and Linux, we can use the color icons à la carte.

[tray-icons.zip](https://github.com/openstenoproject/plover/files/444957/tray-icons.zip)

morinted:
Here are the disconnected and disabled icons with proper masking.

I might rework the icons so that they don't have empty space when there's no symbol showing. Maybe putting the X and bars on top of the key instead of off to the side.

[masked-icons.zip](https://github.com/openstenoproject/plover/files/445205/masked-icons.zip)

morinted:
So I reworked them a little so that the SVGs are set to be square…and then made the key itself a little bit smaller because I found it looked too big in my taskbar. Try this set out.
[plover-tray-square.zip](https://github.com/openstenoproject/plover/files/445527/plover-tray-square.zip)

benoit-pierre:
New preview: [ui_preview_5](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_5)

benoit-pierre:
New preview: [ui_preview_6](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_6)

morinted:
UI preview 6

Tiny notes:

- There's still doubled padding on the right side of the "output" section.
- Maybe in the toolbar the layout should be: `[Lookup] [Suggestions] [Paper Tape] | [Configure]`

For some reason having **configure** grouped with the 3 helper tools feels off.

Slotkenov:
I got an error when stroking DUPT:
> hook 'add_translation' callback <built-in method emit of PyQt5.QtCore.pyqtBoundSignal object at 0x...> failed...
No add translation dialogue came up.
I'm running preview 5 on OSX.
I see there is a preview 6 already, I'll give that a try.

morinted:
ui_preview_6:

1. Start Plover with fresh config
2. Remove a dictionary (in this case, `main.json`
3. Undo remove

Result: exception:

```
  File "plover\plover\gui_qt\dictionaries_widget.py", line 183, in on_undo
    assert self._states
AssertionError
```

benoit-pierre:
OK, fixed in `new_ui.incubation`. Will release a new preview when the builds stuff is merged, and I've rebased onto it.

benoit-pierre:
New preview: [ui_preview_7](https://github.com/benoit-pierre/plover/releases/tag/ui_preview_7)

benoit-pierre:
Closing this, a new PR will be prepared for merging to master.

PR #572, opened 2016-08-07
Dictionary update: Fixed the entry for the backslash (`\`) symbol.

Slotkenov:
Fix for the following issue in #400 

Incorrect entry for the backslash.

Old entry (results in `{^`):
```
"SPWHRAERB": "{^}\\{^}",
```
New entry (resulting in `\`):
```
"SPWHRAERB": "{^\\^}",
```


morinted:
Thanks! I believe `{^}\\\{^}` would also work but the infix operator is good too.

PR #577, opened 2016-08-21
Dictionary entry for lowercase "east"

davidkitfriedman:
Just changing this entry:

"AOEFT": "East",

to:

"AOEFT": "east",

so that there is an entry for lowercase "east".

(Issue #400: Dictionary errors)


morinted:
Thanks for the changes!

PR #582, opened 2016-09-01
Updated main dictionary to reflect most of the messages in #400

Slotkenov:
I took the liberty to update the main Plover dictionary according to most of the posts made in the issue #400 thread. I wasn't sure about everything, so please review.


Slotkenov:
The AS → as results in word boundary errors though. Or was it meant to be `as^`?

plover/assets/main.json
```diff
@@ -40939,7 +40942,7 @@
 "KP*P": "{&X}",
 "KP*P/PH*EPB": "X-Men",
 "KP*P/PWOBGS": "Xbox",
-"KP*P/PWOBGS/3-60": "Xbox 360",
+"KP*P/PWOBGS/306": "Xbox 360",
```
morinted: This is silly but…why not add one as "3-6T"? Like

`KP*P/PWOBGS/3-69` ?

plover/assets/main.json
```diff
@@ -49954,7 +49958,8 @@
 "PA*BG/STAPB": "pakistan",
 "PA*BGS": "PACS",
 "PA*BGT": "pact",
-"PA*D": "peripheral arterial disease",
+"PA*D": "{^pad}",
+"PA*D/WA": "Padua",
```
morinted: How about `PAD/WA*` instead? Otherwise you might get:

`-T/TOUFP/PA*D/WA` for "the touch Padua" instead of "the touchpad was"

plover/assets/main.json
```diff
@@ -100907,7 +100910,6 @@
 "SRER/SUS": "versus",
 "SRER/TABL": "veritable",
 "SRER/TAOES": "vertice",
-"SRER/TAOEUS": "verifies",
```
morinted: I'd leave this in as a misstroke.

plover/assets/main.json
```diff
@@ -142608,4 +142605,4 @@
 "WUT": "but",
 "WUZ": "was",
 "WUZ/KWREU": "wuzzy"
-}
+}
```
morinted: To prevent a diff here, ensure there's a new line at the end of the file.

plover/assets/main.json
Slotkenov: Nice one! And `KP*P/PWOBGS/3-60` actually doesn't need to be removed.

plover/assets/main.json
Slotkenov: Indeed

plover/assets/main.json
Slotkenov: O wait, `KP*P/PWOBGS/3-60` doesn't need its own stroke, right? That's just stroking Xbox and then the numbers 360 separately.

morinted: Well, it does, because it's impossible to stroke "3-60" (`#P-FO`)

Slotkenov:
How can a dictionary change mess up the build? Or is something else going on?

benoit-pierre:
Just some transient failure, I restarted the build and it went fine.

PR #583, opened 2016-09-02
add Python 3 (tests only) builds to AppVeyor and Travis CI

benoit-pierre:
Note: I removed the wxPython requirements in `setup.py` since it's not possible to require it on Windows, and we need to install it manually anyway on all platforms (because it's not installable through pip).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,31 +25,31 @@ matrix:
 
 before_install:
   - |
+    (
+    set -e
     case "$TRAVIS_OS_NAME" in
     osx)
-      brew update &&
-      brew install python --framework &&
-      brew link --overwrite python &&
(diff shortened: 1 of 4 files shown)


benoit-pierre:
Added some missing `;;` to case branches.

benoit-pierre:
Regarding the AppVeyor failure: seem to be a transient error, but I don't have a way to force a rebuild like  with other builds on my repositories, @morinted, can you?

benoit-pierre:
I removed the use of a custom built wheel for hidapi, now that official wheels are available through PyPi.

benoit-pierre:
Removed `six` from `setup_requires` (not needed anymore).

PR #587, opened 2016-09-09
Update license to GPLv2+

### LINKED ISSUE #336: Clarify license: strictly the GPLv2; or, GPLv2 or any later version?
Is Plover licensed under just the GPLv2, or "the GPLv2, or (at your option) any later version of the GPL"? The default GPLv2 license text includes both flavors, and that's what's in the current LICENSE.txt. All the in-source references I found seemed to just say, "See LICENSE.txt", which doesn't clear that up.

morinted:
### About

Update the license to explicitly state GPLv2+, instead of only GPLv2. This is an attempt to resolve our license conflicts; some of our libraries are GPLv3. For this change, we've received permission from all copyright holders in issue #336 

The wording was grabbed from https://pypi.python.org/pypi?%3Aaction=list_classifiers

### Issues

Closes #336

### DIFF
--- a/plover/__init__.py
+++ b/plover/__init__.py
@@ -14,7 +14,7 @@ __credits__ = ['Founded by stenographer Mirabai Knight.\n\nDevelopers:'
                'Benoit Pierre',
                '\nand many more on GitHub: \n<https://github.com/openstenoproject/plover>'
                ]
-__license__ = 'GNU General Public License v2 (GPLv2)'
+__license__ = 'GNU General Public License v2 or later (GPLv2+)'
 __description__ = 'Open Source Stenography Software'
 __long_description__ = """\
 Plover is a free open source program intended to bring realtime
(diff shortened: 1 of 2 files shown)


benoit-pierre:
The `setup.py` classifier needs to be updated too (to `License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)`).

morinted:
Thanks -- don't know why my search for GPL didn't catch that. Oh well. Updated.

benoit-pierre:
`LICENSE.txt` needs to be updated too.

morinted:
I thought that declaring it in the project applied the license without changing the LICENSE.txt:

> Each version is given a distinguishing version number.  If the Program
specifies a version number of this License which applies to it and "any
later version", you have the option of following the terms and conditions
either of that version or of any later version published by the Free
Software Foundation.  If the Program does not specify a version number of
this License, you may choose any version ever published by the Free Software
Foundation.

benoit-pierre:
Of course...

PR #590, opened 2016-09-12
Add ProCAT machine

morinted:
### About

Add the ProCAT machine, thanks to David Graper for sending a handy little diagram from his Blaze: [BlazeKeyouts.pdf](https://github.com/openstenoproject/plover/files/468399/BlazeKeyouts.pdf)

When contacting ProCAT, the representative told me that the Flash, Blaze, Stylus I & II, Impression, and Xpression all use the same protocol. Thankfully that should mean we get support for them all.

### Reason

While some ProCAT writers could emulate TX Bolt, it wasn't true of all of them. I believe the Blaze was the last to emulate another protocol.

### Release Notes

We should add ProCAT machines to the list of supported writers.

### Tested Platforms

Glen and I tested on Windows, but I'll leave further testing for the weekly -- I imagine it will work since we're just using the serial base class.

### DIFF
--- a/plover/machine/procat.py
+++ b/plover/machine/procat.py
@@ -1,0 +1,63 @@
+# Copyright (c) 2016 Ted Morin
+# See LICENSE.txt for details.
+
+"""Thread-based monitoring of a ProCAT stenotype machine."""
+
+# Python 2/3 compatibility.
+from six import iterbytes
+
+from plover.machine.base import SerialStenotypeBase
+
+"""
+ProCAT machines send 4 bytes per stroke, with the last byte only consisting of
+FF. So we need only look at the first 3 bytes to see our steno. The leading
+bit is 0.
+"""
+STENO_KEY_CHART = ('', '#', 'S-', 'T-', 'K-', 'P-', 'W-', 'H-',
+                   'R-', 'A-', 'O-', '*', '-E', '-U', '-F', '-R',
+                   '-P', '-B', '-L', '-G', '-T', '-S', '-D', '-Z',
+                   )
+
+BYTES_PER_STROKE = 4
+
+
+class ProCAT(SerialStenotypeBase):
+    """Interface for ProCAT machines.
+    """
+
+    KEYS_LAYOUT = '''
+        #  #  #  #  #  #  #  #  #  #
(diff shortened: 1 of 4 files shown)


plover/system/english_stenotype.py
```diff
@@ -158,6 +158,31 @@
...
+        '-R'   : '-R',
+        '-P'   : '-P',
+        '-B'   : '-B',
+        '-L'   : '-L',
+        '-G'   : '-G',
+        '-T'   : '-T',
+        '-S'   : '-S',
+        '-D'   : '-D',
+        '-Z'   : '-Z',
+    },
```
benoit-pierre: How about using:
```python
KEYMAPS['ProCAT'] = KEYMAPS['TX Bolt']
```
Since it's the same?

plover/machine/procat.py
```diff
@@ -0,0 +1,68 @@
...
+                   )
+
+BYTES_PER_STROKE = 4
+
+
+class ProCAT(plover.machine.base.SerialStenotypeBase):
+    """Standard stenotype interface for a ProCAT machine.
+
+    This class implements the three methods necessary for a standard
+    stenotype interface: start_capture, stop_capture, and
```
benoit-pierre: This paragraph is really unnecessary, no? And wrong too.

plover/machine/procat.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2016 Ted Morin
+# See LICENSE.txt for details.
+
+"""Thread-based monitoring of a ProCAT stenotype machine."""
+
+# Python 2/3 compatibility.
+from six import iterbytes
+
+import plover.machine.base
```
benoit-pierre:
```python
from plover.machine.base import SerialStenotypeBase
```
would be simpler.

plover/machine/procat.py
morinted: Yep, copy-pasting is not always the best idea, is it?

morinted:
Made those changes, also applied to geminipr which was my copy-paste.

morinted:
Travis Mac builds seem to be queuing indefinitely right now for me.

Issue #557, opened 2016-07-13
Fingerspelling deletions don't always work with "After Output" space placement


SyntaxBlitz:
Classification: Bug

Reproducibility: Always
## Summary

Using the asterisk key to delete letters in fingerspelled words occasionally causes problems when the space placement is set to "After Output"
## Steps to Reproduce
1. Set Space Placement to "After Output"
2. Type a few words with Plover
3. Fingerspell a word with a few letters (like "taco")
4. Press the asterisk `n - 2` times where n is the number of letters in the fingerspelled word
## Expected Results

The first two letters of the fingerspelled word remain.
## Actual Results

The fingerspelled word is replaced with a space, followed by the second letter of the word.

[Gif of the problem](https://i.imgur.com/nygJ5av.gif)

The relevant part of strokes.log:

```
2016-07-12 21:53:56,486 Translation(('EU',) : I)
2016-07-12 21:53:57,520 Stroke(H- R- A- O- -E -U -B -G)
2016-07-12 21:53:57,520 Translation(('HRAOEUBG',) : like)
2016-07-12 21:53:58,634 Stroke(T- *)
2016-07-12 21:53:58,635 Translation(('T*',) : {>}{&t})
2016-07-12 21:53:59,112 Stroke(A- *)
2016-07-12 21:53:59,112 Translation(('A*',) : {>}{&a})
2016-07-12 21:53:59,536 Stroke(K- R- *)
2016-07-12 21:53:59,536 Translation(('KR*',) : {>}{&c})
2016-07-12 21:53:59,928 Stroke(O- *)
2016-07-12 21:53:59,928 Translation(('O*',) : {>}{&o})
2016-07-12 21:54:00,223 Stroke(S- *)
2016-07-12 21:54:00,224 Translation(('S*',) : {>}{&s})
2016-07-12 21:54:00,890 Stroke(*)
2016-07-12 21:54:00,890 *Translation(('S*',) : {>}{&s})
2016-07-12 21:54:01,282 Stroke(*)
2016-07-12 21:54:01,282 *Translation(('O*',) : {>}{&o})
2016-07-12 21:54:01,729 Stroke(*)
2016-07-12 21:54:01,729 *Translation(('KR*',) : {>}{&c})
2016-07-12 21:54:03,151 Stroke(*)
2016-07-12 21:54:03,151 *Translation(('A*',) : {>}{&a})
2016-07-12 21:54:03,607 Stroke(*)
[...]

Issue #471, opened 2016-04-11
Repeat Last Stroke after typing two numbers outputs the numbers reversed for some combos


david-pitty:
Classification: Bug

Reproducibility: Always
## Summary

Repeat last stroke after typing two numbers outputs the reversed number two numbers.
## Steps to Reproduce
1. I type: 12
2. I use Repeat last stroke
## Expected Results
1. 1212
## Actual Results
1. 1221
## Version

Plover v2.5.8+567.g52ebe0b.

Installed via: download from https://github.com/openstenoproject/plover/releases/tag/weekly-v2.5.8%2B567.g52ebe0b
## Notes

Using Repeat last stroke several times keeps outputting the reversed number although EU was never typed. 
This could not be replicated with 1313, 1818, 1919, 4848, 4949, 5959
I was able to replicate with 0990, 1221, 1441, 1551, 2552, 3553, 4554
## Configuration

OS:
Windows 8.1

david-pitty:
Issue present with both space placement configurations.

morinted:
I get the feeling that #287, #470, and this one are all related and will require a rework of either repeat-last-stroke or the number bar...

benoit-pierre:
Fixed in master.

PR #592, opened 2016-09-17
add/fix some tests and related bugs

benoit-pierre:
* speed up Stentura tests (reduce timeout)
* fix possible failure when running the default dictionaries tests: a `UnicodeDecodeError` exception when running under tox
* improve default dictionaries tests to check for shadowed/overridden keys in lower precedence dictionaries (and fix `main.json` accordingly)
* add xfail test for #557 
* add a test for #471 and fix the bug: this exhibited another bug in the translator (handling of last translations, fixed too)


benoit-pierre:
Damn those number keys...

benoit-pierre:
Had to replace the use of `system.KEY_ORDER` by `system.KEYS` in a few places, now that the former also includes the virtual number keys...

morinted:
It would be nice to remake the number keys as a plugin or scripted dictionary so that it's not magic.

I've reviewed the changes and they look good as far as I can tell, but I'd prefer to try them out for a little bit to see if I can create any breaking cases. I'll soak them before Monday. Then I'm up for a weekly release.

PR #593, opened 2016-09-18
Strange Entry for "KR-FRPB"

davidkitfriedman:
This looks like it might have been for just some particular situation:

<code>
"KR-FRPB": "Pshhhkkkkkrrrrkakingkakingkakingtshchchchchchchchchdingdingding",
</code>

#400 Dictionary errors


benoit-pierre:
Thanks, but this PR should really contain one single commit with the aforementioned change, can you rebase it?

davidkitfriedman:
Maybe that rather small change could just be integrated into a bunch of other updates. I can just add that as a comment to the issue #400 thread.

PR #598, opened 2016-09-23
Ignore only KeyPress when a modifier is set

### LINKED ISSUE #597: Releasing a key with different modifiers causes the key to "stick"
Steps to reproduce this bug:

1. Press any key that is used for steno (ex. Q on qwerty, which maps to steno S).
2. Press any modifier key (shift, ctrl, alt, etc.).
3. Release the first key that was pressed.
4. Release the modifer key.

Since the key was released with the modifier held down, Plover doesn't see it as the same key and thinks the key was never released, continuing to wait for the chord to be completed.

This isn't an issue if you're only typing in steno since modifier keys aren't used anyway, but I frequently switch between steno and qwerty. Rapid typing in qwerty causes the situation described to occur fairly commonly (usually when failing to release the last letter or two of a previous word while typing a subsequent capitalized word), and then trying to chord the `{PLOVER:TOGGLE}` command doesn't work.

Reproduced in 3.0.0+253.g244f590 (the current latest commit), installed via AUR.

KeyboardFire:
Previously, both KeyPress and KeyRelease were ignored if there were any modifiers. This runs into the issue of a key "sticking" in a perpetually pressed state if the key is pressed without modifiers but released with modifiers.

This wouldn't be an issue in steno, but it creates problems when switching in and out of steno with `{PLOVER:TOGGLE}` and typing rapidly with a conventional keyboard layout. (See the GitHub issue referenced for further details.)

Fixes #597.

### DIFF
--- a/plover/oslayer/xkeyboardcontrol.py
+++ b/plover/oslayer/xkeyboardcontrol.py
@@ -213,16 +213,15 @@ class KeyboardCapture(threading.Thread):
             assert event.data.sourceid in self.devices
             keycode = event.data.detail
             modifiers = event.data.mods.effective_mods & ~0b10000 & 0xFF
-            # Ignore event if a modifier is set.
-            if modifiers != 0:
-                continue
             key = KEYCODE_TO_KEY.get(keycode)
             if key is None:
                 # Not a supported key, ignore...
                 continue
             # ...or pass it on to a callback method.
             if event.evtype == xinput.KeyPress:
-                self.key_down(key)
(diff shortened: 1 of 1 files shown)


morinted:
LGTM, @benoit-pierre ?

benoit-pierre:
I tested on Windows 10, and we seem to suffer from the same problem, what about OS X, can you test @morinted?

morinted:
Yes, also an issue on Mac.

PR #604, opened 2016-09-29
machine/procat: detect when packet is too small

### LINKED ISSUE #601: ProCAT Protocol Stops Working
Classification: Bug

Reproducibility: It has happened twice since installation, which was about two hours ago as I write this.
## Summary

With Plover set to the ProCAT protocol, I will be writing along, and suddenly Plover stops sending my text to whatever I'm writing in.
## Steps to Reproduce
1. I set Plover to use the new ProCAT protocol and write.
## Expected Results
1. This should happen: Plover should send my output to whatever I'm writing to for as long as I'm writing.
## Actual Results
1. Instead, this happen: Plover just stops sending what I'm writing to whatever I'm writing in at the time.  :(
   [plover_proCAT_error.txt](https://github.com/openstenoproject/plover/files/491974/plover_proCAT_error.txt)

morinted:
### Issues

Fixes #601 

### Tested Platforms

Glen tested this yesterday on his 10.9 installation and didn't run into his issue.

### DIFF
--- a/plover/machine/procat.py
+++ b/plover/machine/procat.py
@@ -41,6 +41,8 @@ class ProCAT(SerialStenotypeBase):
             raw = self.serial_port.read(BYTES_PER_STROKE)
             if not raw:
                 continue
+            if len(raw) < BYTES_PER_STROKE:
+                raw += self.serial_port.read(BYTES_PER_STROKE - len(raw))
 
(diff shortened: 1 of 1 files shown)


morinted:
He ran into it again! Will revisit.

PR #605, opened 2016-09-29
PRIORITY_DICT

KoiOates:
### Reason

Feature described in "Any update on Spanish input in Plover?" on the Google group.

### Release Notes

Uses the PLOVER:PRIORITY_DICT command to prioritize a dictionary or comma separated list of dictionaries, the first one being the highest priority.
e.g.
{
  "SP-RB": "{PLOVER:PRIORITY_DICT:spanish/main.json; commands.json, briefs.json}",
  "HR-RB": "{PLOVER:PRIORITY_DICT:main.json; commands.json, briefs.json}"
}
### Tested Platforms

Tested on Ubuntu 15.10

I'm getting less GitHub-illiterate by the hour, but this is my first time trying a pull request. Let me know if something's off.

### DIFF
--- a/plover/dictionary/prioritize_dictionaries.py
+++ b/plover/dictionary/prioritize_dictionaries.py
@@ -1,0 +1,27 @@
+from plover import log
+
+def prioritize_dictionaries(selections, old_file_names):
+    def match_partial_path(file_names, selection):
+        matches = []
+        for file_name in file_names:
+            if file_name.endswith(selection):
+                matches.append(file_name)
+
+        if matches:
+            matches.sort(key=len)
+            return(matches[0])
+
+    selections = [x.strip() for x in selections.split(";")]
+    selections.reverse()
+    new_file_names = old_file_names[:]
+    for selection in selections:
+        matching_file = match_partial_path(old_file_names, selection)
+        if matching_file:
+            ix = new_file_names.index(matching_file)
+            t = new_file_names.pop(ix)
+            new_file_names.append(t)
+        else:
+            log.error('No dictionary file matching "%s" found.' % selection, exc_info=True)
+        
+    return new_file_names     
+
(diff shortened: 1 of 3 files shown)


morinted:
Thanks for implementing this. I'll have some feedback on the code, but before that, two git things:

- Your email that you committed with isn't on your GitHub account, you can either add the email to your profile or change the author on the commits and repush.
- Since these changes aren't really distinct, it would be best to squash them into one and force push to update your branch, ideally with a new commit message.

morinted:
Ah, and the build failing is on our part → there have been some dependency updates, so don't worry about it.

It would be nice to see tests for this change, maybe you can pull out some logic into a testable function?

And in terms of implementation, I think it's possible to have commas in file names in at least some of the main OSes, maybe it would be wise to use a different delimiter (;, perhaps?)

KoiOates:
Thank you for the GitHub advice, Ted. I'm just excited to have something I can help with on this project! If nothing comes up to drag me away from the computer I'll make the changes you suggested later tonight.

morinted:
To fix the building issues on Travis you should be able to rebase against master.

KoiOates:
Here's a start at least for testing. I wasn't quite sure if this was what you had in mind (or where the appropriate place to put the new code would be) but it just verifies that a list of dictionary filenames (similar to what would be retrieved by config.get_dictionary_file_names) ends up in the expected order based on user input.

morinted:
There may be concurrency issues with an engine update mid-stroke, right @benoit-pierre ?

benoit-pierre:
Yes, I'd prefer if we hold on this kind of changes until the plugins system is available.

morinted:
So @KoiOates I like the feature and would like to get into it more once we are better settled with the plugin system -- does that make sense?

KoiOates:
Definitely. I can imagine people eventually swapping plugins for the features they need the way people are now sharing dictionaries, and meanwhile Plover stays clean and robust at the core. I like where this is going!

PR #607, opened 2016-09-29
rework builds

benoit-pierre:
* add a `bootstrap.sh` (for Arch Linux / Ubuntu / OS X) and `bootstrap.bat` for Windows
* rework Travis build to use `bootstrap.sh`: this will speed up the OS X builds thanks to the wheels cache (e.g. `pyobjc-core` won't need to be always built from source)
* fix Travis OS X jobs: use the latest (`xcode8`) image
* add Ubuntu package rules and helper (`linux/debuild.sh`), update the Travis build to create the package on tagged Python 2 build
* add Arch LInux package rules and helper (`linux/makepkg.sh`)
* update Linux / OS X / Windows documentation


benoit-pierre:
I noticed to Python 3 leftover changes in the Debian rules. Will amend.

morinted:
Looks great!

PR #613, opened 2016-10-02
New UI

benoit-pierre:
Slightly cleaned-up version of #571.

* remove the wxWidgets GUI
* switch to using Python 3
* rework the engine
* add 2 GUIs: headless mode, and PyQT5


benoit-pierre:
`plover/gui_qt/resources/states.xcf` is not needed anymore, will amend.

benoit-pierre:
OK, removed it, along with 2 unused imports.

PR #615, opened 2016-10-04
 Plugins support: GUIs, machines, systems

benoit-pierre:
Rework the registry to make it more powerful and use it to support plugins.

# Implementation

 `pkg_resources` is used for loading distributions from a `plugins` directory (same location as the configuration file). Entry points are used to declare and find supported plugins, e.g. for Plover itself:

```
        [plover.gui]
        none = plover.gui_none.main
        qt   = plover.gui_qt.main

        [plover.machine]
        Gemini PR = plover.machine.geminipr:GeminiPr
        Keyboard  = plover.machine.keyboard:Keyboard
        Passport  = plover.machine.passport:Passport
        ProCAT    = plover.machine.procat:ProCAT
        Stentura  = plover.machine.stentura:Stentura
        TX Bolt   = plover.machine.txbolt:TxBolt
        Treal     = plover.machine.treal:Treal

        [plover.system]
        English Stenotype = plover.system.english_stenotype
```

Note that on Linux this include all Python packages available through the standard mechanisms (installed in the site package directory, available in `PYTHONPATH`, ...).

Plugins are loaded as needed.

## GUI

GUI plugins are selectable using the `-g GUI_NAME` command line option. A valid plugin consist of a module providing 2 functions:

```python
def show_error(title, message):
    '''Display an error message box.

    <title>: title string.
    <message>: message string.
    '''
    pass

def main(config):
    '''Start the GUI.

    <config>: an instance of plover.config.Config (not yet loaded, will be saved on return).

    Can return an exit code (that will be passed to sys.exit).
    '''
    return 0
```

## Machine

A machine plugin will automatically be available through the usual configuration mechanisms. A valid plugin must be a class compatible with `plover.machine.base.StenotypeBase`. See https://github.com/benoit-pierre/plover_italian_stentura for a valid machine plugin: support for the Italian version of the Stentura.

## System

The system in use can now be configured in the new `System` tab of the configuration dialog. A valid system plugin is a module, similar to the one already provided for English Stenotype support in `plover.system.english_stenotype`. See https://github.com/benoit-pierre/plover_template_system for a sample template.

Note that it's not necessary for a system to directly provide keymaps for all supported machine types, and so to modify each compatible system when adding support for a new machine type: instead, it's possible to declare the new type as compatible with an existing model, e.g. for the Italian Stentura:

```python
class ItalianStentura(plover.machine.base.SerialStenotypeBase):

    KEYMAP_MACHINE_TYPE = 'Stentura'
```

# What's next?

Support for other plugin types:
* `plover.asset`: so a Plover plugin package can export assets (dictionaries, word lists). I already have POC for this.
* `plover.dictionary`: to support new dictionary formats (I already have POC implementations for Python dictionaries, as well as CSV and ODS formats).
* `plover.command`: custom commands. Use cases: a command will be called with an optional command line, and an instance of the current engine, so it will be possible to change the engine configuration (`PLOVER:PRIORITY_DICT` for example), call other Python modules, run external commands, implement a pause command (since this is done in the engine thread, the machine can keep queuing strokes in parallel)... I have a POC for this too, but 2 questions not to be addressed: is the command name case insensitive, and do we support those when the output is not enabled (like for `RESUME`, `TOGGLE`, and `QUIT`).
* `plover.extension`: a kind of general plugin, that must be explicitly enabled. Use case: register to engine hooks, to react on some events. Can monitor external events and change the engine configuration when they occur (for example, monitor window changes?).
* `plover.macro`: where a command is handled at the formatting stage, a macro would act on the translator stage. Use case: new retrospective commands. This part need to be formalized.


benoit-pierre:
Thinking more about case insensitive command names, I think it's not too much trouble, and we can make the configuration more robust too (case insensitive machine/system name) by changing the registry API like this:

```python
    def register_plugin(self, plugin_type, entrypoint):
        log.info('%s: %s (from %s)', plugin_type,
                 entrypoint.name, entrypoint.module_name)
        self._plugins[plugin_type][entrypoint.name.lower()] = entrypoint

    def get_plugin(self, plugin_type, plugin_name):
        return self._plugins[plugin_type][plugin_name.lower()]

    def list_plugins(self, plugin_type):
        return [
            entrypoint.name
            for entrypoint in self._plugins[plugin_type].values()
        ]
```

Thoughts?

Also, I removed the keyboard aliases (`Microsoft Sidewinder X4` and `NKRO Keyboard`), do we still want to support those?

KoiOates:
You're my hero for working on this plugin system. This is going to make my Plover based projects both easier and cleaner! I like the case insensitive naming of plugins. Could save hours of hunting for problems after making a hard to spot typo at three in the morning. As for command names, I'm personally used to them being all uppercase as a convention, so I could go either way on that.

Now that I think of it, the only command I've ever thought to use when output isn't enabled is RESUME to turn it back on. So far I haven't expected Plover to do anything otherwise. That being said, I could still see it being useful to have chords mapped to arbitrary chunks of Python whether or not output is enabled (as long as the chords chosen are long enough not to be accidentally triggered while typing in qwerty). I'm envisioning a power-user using Plover chords to launch applications or run scripts in the background. Something like {PLOVER:RUN_EXTERNAL:/usr/bin/kill_video_game_i_shouldnt_be_playing_at_work_and_look_busy.sh --now} :)

benoit-pierre:
I wend ahead and rewrote the code so plugin names are case insensitive.

> Now that I think of it, the only command I've ever thought to use when output isn't enabled is RESUME to turn it back on. So far I haven't expected Plover to do anything otherwise. That being said, I could still see it being useful to have chords mapped to arbitrary chunks of Python whether or not output is enabled (as long as the chords chosen are long enough not to be accidentally triggered while typing in qwerty). I'm envisioning a power-user using Plover chords to launch applications or run scripts in the background. Something like {PLOVER:RUN_EXTERNAL:/usr/bin/kill_video_game_i_shouldnt_be_playing_at_work_and_look_busy.sh --now} :)

The problems with commands when output is disabled are:
* they're worthless when not using a keyboard, since there's really no point in disabling the output in that case anyway
* when a keyboard is used, and output is disabled, suppressing the stroke after the fact is a hack, and only work if backspace can be used to undo the result of each keypress (like in a text field)

morinted:
+1 for case insensitive. User friendly is the name of the game.

For disabled machines, is it possible in the case of read only TX protocol
for multiple listeners to be going? Like Ecplise and Plover running and the
user switching between?

benoit-pierre:
> For disabled machines, is it possible in the case of read only TX protocol for multiple listeners to be going? Like Ecplise and Plover running and the user switching between?

The current machine is always in use by Plover (since it has to listen to strokes even if the output is disabled to handle things like the RESUME command).  The only way to "disable" it is to switch to another machine or quit Plover.

balshetzer:
It might be an interesting feature to multiplex or pipe through steno
machines.

For example, imagine if Plover or a plugin could create one or more virtual
serial ports and output to that in specified formats like tx bolt. Then,
you could input into Plover with whatever machine and then Plover could
send output to multiple other programs using whatever protocol they prefer.
You just set up the other programs to listen on the virtual port and they
think they're connected to a machine. This way you could use Plover and
eclipse at the same time if you needed to.

benoit-pierre:
Adding support for output plugins is certainly possible, but emulating a virtual serial port is not going to be possible on all platforms (Windows...).

morinted:
While serial may not be possible without some serious legwork on Windows, I
believe Eclipse can read from a number of sources, including a IP-based
protocol (Wi-Fi machines). It should be possible to run something on
localhost to support what Hesky suggests.

PR #616, opened 2016-10-04
* main.json: add 0D and missing n:00 strokes.

JoshuaGrams:
Added a bunch of number strokes which are mentioned in Learn Plover:
- Time strokes for hours from 13:00 to 23:00
- The strokes for hours up to 12 which use the same hand for the modifier.
- #0D for 00

The three strokes which get overwritten are mis-strokes for the versions with a star.


morinted:
Thanks so much! But as for the same-hand clock strokes, I'm not certain..
 @stenoknight what do you think? I'm mainly uncomfortable with `-B8G` and maybe `-7BG`.

JoshuaGrams:
Yeah, I wasn't sure about that, but they're in Learn Plover. Might be better to fix Learn Plover instead...

morinted:
Mirabai said A-OK! They are some weird strokes but I do remember being confused about Learn Plover being off about it. Thanks for your efforts!

benoit-pierre:
It's too late now, but this probably should have been targeted at the `stable` branch.

PR #617, opened 2016-10-06
Auto reconnect writer

### LINKED ISSUE #596: auto-detect writer connection state
Classification: Usability, Enhancement

Reproducibility: Not Applicable
## Summary

Plover currently does not automatically discover if a serial device like the Stenoboard or ErgoDox in TX Bolt mode is disconnected and reconnected again or if the laptop comes back from standby.
## Steps to Reproduce
1. Unplug the device and plug it in again
2. or close the notebook (goes into standby) and reopen it again
## Expected Results
1. In both cases, the steno device should work again
## Actual Results
1. When unplugged and plugged in again, nothing works, hitting the refresh-button in plover does not work. I have to unplug again, hit refresh (plover discovers the device is disconnected), plug the device in again, hit refresh again.
2. When the laptop comes back from standby, I have to hit the refresh button once to get it working again.
## Version

Plover commit 244f590 (current master)
## Notes

I tried adding a timer to call _reconnect every couple of seconds, but I don't know if this is the right approach. 

I think the problem in scenario 1 is that plover still has the device (/dev/ttyACM0 in my case) "open". When the device is plugged in again, it therefore gets assigned the next free device /dev/ttyACM1. Hitting refresh in plover releases the handle to /dev/ttyACM0 again, thus allowing that name to be taken on the next plug in of the device.
## Configuration

OS: Linuxmint 17.2 Cinnamon

tschulte:
### Auto reconnect writers after Suspend or temporary disconnects

Automatically discover writer disconnection e.g. due to OS suspend. 

### Issues

Fixes #596

### Release Notes

Disconnects of writers are now discovered more reliably. If the writer is plugged in again within 30 seconds, it will be auto-reconnected. This does allow to suspend a notebook running plover with a writer attached. After turning the notebook on again, the writer does work again without user interaction.

### Tested Platforms

Tested with Linuxmint 17.3
* Stenoboard in both TxBolt and GeminiPR: both manual disconnect and OS suspend work
* ErgoDox using TxBolt protocol: recovers from OS suspend, but not from manual disconnect. But it seems the ErgoDox generally has issues. The com port can not be reopened on master either.

Since I refactored `ThreadedStenotypeBase` and all it's subclasses, all writers should be tested before merging this PR.


benoit-pierre:
I wonder if it might not be better to keep the machine classes dumb, and instead handle automatic reconnection in the engine class.

tschulte:
I was not aware of the engine class yet. You might be right. The engine would have to handle the state changes. Once the state is `RUNNING`, the retry-counter would go back to 30. If the state changes to `ERROR` and the counter is greater than 0, decrement, wait a second and reinitialize the machine.

Is that what you mean?

benoit-pierre:
Yes, exactly.

morinted:
I think moving the reconnect code out of the machine classes (Treal) and into the engine class is probably a good idea. Maybe we can discuss what sort of configuration (if any) the user should have?

Closing this PR for now, but thanks for submitting it and I welcome further developments!

tschulte:
I would suggest basing further changes on my branch, because

* less duplication (loop code)
* more consistent Exception handling code (some machines did not have any Exception handing in their run method)

Where do you want to discuss this? In the issue or here?

morinted:
Issue would be best.

The history on this branch would need to be cleaned up to base changes on
it. Perhaps we can reapply the changes we like on a new branch for a
cleaner commit history.

Issue #599, opened 2016-09-25
TX Bolt timeout causes noticeable delay on some machines


morinted:
Classification: Performance

Reproducibility: Usually -- slower machines (and Macs) are more prone.
## Summary

With the 100ms delay, some users notice lag while writing.
## Expected Results

While writing, the delay should be short enough so that the user can't tell there's a delay.
## Actual Results

Right now, the 100ms delay is _on top of_ the processing time between reads which makes the actual delay longer.
## Proposed Solutions

@benoit-pierre has two steps to resolve this issue:
1. Incorporate the processing time into the timeout. Instead of waiting 100ms after starting to read, it should be 100ms minus whatever time it took in the last loop.
2. Allow the timeout to be configurable, given a sane default. Right now the 100ms is hardcoded, and we ignore the user setting. It would be nice to listen to the user setting and provide a sane default like 75ms.

morinted:
I think that, for now, TX Bolt works well enough and if one wants a faster protocol there are others available.

PR #619, opened 2016-10-11
machine/txbolt: improve timeout handling

benoit-pierre:
Rewrite handling so the code really wait at most 50ms (taking into account the time spent processing a stroke), and at least 25ms (to avoid splitting a stroke when using some machines).

Test release available [here](https://github.com/benoit-pierre/plover/releases/tag/txbolt_timeout_test).

### DIFF
--- a/plover/machine/txbolt.py
+++ b/plover/machine/txbolt.py
@@ -6,6 +6,8 @@
 # Python 2/3 compatibility.
 from six import iterbytes
 
+from monotonic import monotonic
+
 import plover.machine.base
 
 # In the TX Bolt protocol, there are four sets of keys grouped in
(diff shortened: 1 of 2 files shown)


benoit-pierre:
Should address (part of) #599.

benoit-pierre:
Thinking more about this, I realize that my reasoning is flawed. If we detected the end of the stroke, then the time we take processing it is irrelevant, as a timeout during the next read does not signal the end of the next stroke. So this patch would really only help if we take an inordinate amount of time unpacking the raw bytes.

morinted:
Well, it might help the case where a computer is experiencing a heavy case of throttling, though? I still seems valid to me.

morinted:
So I tried this, I still get the occasional split with the LightSpeed over Bluetooth on Mac. Any suggestions for debugging?

morinted:
I *still* had the occasional splitting issue with the LightSpeed over Bluetooth on the Mac with this build. I'm exacerbated; maybe the data is really flowing through *that* slowly on Mac. Perhaps we should just set a default of 75ms, and make it user-configurable and call it a day?

Issue #178, opened 2014-02-05
Add support for "glue" with a specified separator


balshetzer:
Right now Plover supports "glue" which will cause all output items with glue to be attached without spaces.

However, sometimes the user wants to glue with a specified separator such as a hyphen.

We will add this feature with the following syntax. Glue will continue to be indicated by & for example: {&A} {&B} (&C} will result in ABC.

To connect with a hyphen the user would use: {&-A} {&-B} {&-C} which will result in A-B-C.

morinted:
What is the reason that this doesn't work now? Is & limited to one character?

morinted:
@stenoknight Where does the use case for this come from? Can you provide an example stroke set that I'd use for things that I want to separate with a concern operator?

stenoknight:
People might want stitching (A-B-C-D) alphabets for spelling or initialized (A.B.C.D.) alphabets for acronyms or initials. But frankly, I just handle these cases by putting the hyphens or periods in manually when they come up. Very low priority, IMO.

PR #623, opened 2016-10-17
Add support for stitching mode

arxanas:
See also the commit message.

Add support for a stitching mode, similar to #178.

Stitching can be enabled with a dictionary entry like `{MODE:SET_STITCH:-}`. I have `SP*EL` set to `{MODE:CAPS}{MODE:SET_STITCH:-}` to transcribe someone spelling a word.

There's a test case, and I also tested it manually on macOS.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -280,7 +280,7 @@ class _Action(object):
         self.command = command
 
     @classmethod
-    def from_last(self, last_action, *args, **kwargs):
+    def from_last(cls, last_action, *args, **kwargs):
         """Construct a new action, persisting the appropriate variables."""
         persistent_state = {
             "case": last_action.case,
@@ -292,7 +292,7 @@ class _Action(object):
                 "explicit keyword arguments was a persistent state variable.")
         kwargs.update(persistent_state)
 
-        return _Action(*args, **kwargs)
+        return cls(*args, **kwargs)
 
     def copy_state(self):
         """Clone this action but only clone the state variables."""
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -12,6 +12,8 @@ from collections import namedtuple
 import re
 import string
 
+from six import viewkeys
+
 from plover import orthography
 
 
@@ -210,7 +212,7 @@ class _Action(object):
     CASE_NONE, CASE_UPPER, CASE_LOWER, CASE_TITLE = range(4)
 
     def __init__(self, attach=False, glue=False, word='', capitalize=False,
-                 lower=False, orthography=True, space_char=' ',
+                 lower=False, orthography=True, space_char=' ', stitch_char='',
                  upper=False, upper_carry=False,
                  case=CASE_NONE, text='', replace='', combo='',
                  command=''):
@@ -241,6 +243,9 @@ class _Action(object):
         space_char -- this character will replace spaces after all other
         formatting has been applied
 
+        stitch_char -- this character will be interspersed between each
+        character of output, except spaces
+
         case -- an integer to determine which case to output after formatting
 
         text -- The text that should be rendered for this action.
@@ -265,6 +270,7 @@ class _Action(object):
 
         # Persistent state variables
         self.space_char = space_char
+        self.stitch_char = stitch_char
         self.case = case
 
         # Instruction variables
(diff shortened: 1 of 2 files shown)


morinted:
Stitching in traditional software is within the strokes, e.g. a fingerspelling alphabet that uses hyphens. Right now if you were just to write a fingerspelling dictionary, you'd have a trailing stitch (entries like {&a-} and {&b-} would come out as "a-b-" instead of the intended "a-b".)

While a mode certainly sounds interesting, it would affect all text. I think it's never been done (at least I've never heard of it before!)

So I'd clarify your PR as a stitching "mode" as opposed to stitching support. A true stitching support would also handle the RTF syntax. I'm not sure how we would differentiate between fingerspelling and stitching for the regular user. Maybe {&a{-}}…? I don't know, it's a tough one.

And the set spaces command is a translation mode, so to use it you use: `{MODE:SET_SPACE:}` instead of just `{SET_SPACE:}`.

arxanas:
I've updated the description of the commit and undone the "fix" for `SET_SPACE`.

It looks like there's precedence for a stitching mode. [This CaseCATalyst document](http://www.phoenixtheory.com/wp-content/uploads/2016/03/Retroactive-Stitching-Macro.pdf) discusses a stitching command. This semantics are more complicated, but I think the discrepancy here isn't that important:

> For instance, you stroke the command for {Stitch On}[-], write the word "stitch," then stroke the command for {Stitch Off} and the result will be the stitching of the word between the commands:  s-t-i-t-c-h.  If you stroke {Stitch On}[-], write "stitch this" then stroke the command for {Stitch Off} the result will be stitch-this.  If there are two or more words between the commands, the hyphens will appear between the words.

morinted:
This is news to me, thanks for sharing. Yeah, I guess our recommendation to users trying to replicate the behavior would be to use set space and set stitch accordingly.

Would you be able to verify that the behavior works as intended on both spaces before and spaces after mode in Plover?

Seems also that the test cases are failing in (at least) Windows

screen shot 2016-10-17 at 2 51 10 pm

arxanas:
Tested with both before and after space output, it appeared to work:

screen shot 2016-10-17 at 11 54 06

The build seems to be stalled waiting for OS X on Travis, but other than that I believe it's fixed.

gdwaynewarner:
Usually during a deposition, the witness will be asked to spell his or her name.  The court reporter would write something like this:

Q. Please state your name and spell it for the record.  

A. My name is Glen Warner, and that's spelled G-L-E-N, and the last name is spelled W-A-R-N-E-R.  

Unfortunately, I have seen some folks using mixed case to spell things lately, so perhaps there is a need for a mixed-case stitching mode as well.  

Finally, here's the text from the 2010 digitalCAT manual on stitching, if that's any use:

"Stitch Word

Using this command will "stitch" or hyphenate each individual letter of a translation. For example, if a person named John spells his name, it could appear as "J-o-h-n."

This command accepts the following parameters: LASTWORD, OFF, ON, TOGGLE. LASTWORD instructs the program to stitch the previous translation. ON instructs the program to begin stitching. The program will continue to stitch the words or translations that are written until stitching is turned off. This can be useful if you are finger spelling. OFF will instruct the program to stop stitching at this point. TOGGLE will instruct the program to begin stitching if it is not currently stitching words. It will also instruct the program to stop stitching if it is currently stitching.

Entry: {@STITCHWORD(PARAMETER)} Possible Parameters: LASTWORD, OFF, ON, TOGGLE"

arxanas:
@gdwaynewarner I think I did a poor job of demonstrating how to actually use the stitching. On its own, the only thing that the stitching mode does is add characters in between other characters. Casing is handled by other commands, which means that it supports mixed-case stitching.

Suppose I had these entries:

```
"KPA": "{}{-|}",
"KA*PS": "{MODE:CAPS}",
"TAO*EULT": "{MODE:TITLE}",
"ST*EUFP": "{MODE:SET_STITCH:-}",
"KP-PL": "example",
"TEGT": "text",
```

Then here are some example ways to use the stitching mode:

```
KP-PL
example

ST*EUFP KP-PL
e-x-a-m-p-l-e

KA*PS ST*EUFP KP-PL
E-X-A-M-P-L-E

TAO*EULT ST*EUFP KP-PL TEGT
E-x-a-m-p-l-e T-e-x-t

ST*EUFP KP-PL KPA TEGT
e-x-a-m-p-l-e T-e-x-t
```

So you can see that the casing isn't affected by the stitching. In my case, I made an entry `SP*EL` which mapped to `{MODE:CAPS}{MODE:SET_STITCH:-}`, which did the job of both entering stitching mode and also capitalizing the text. But this is entirely optional.

gdwaynewarner:
@arxanas  Looks good to me! Looking forward to testing it in the next build!

plover/formatting.py
```diff
@@ -937,6 +972,20 @@ def _apply_space_char(text, space_char):
         return text
 
 
+def _apply_stitch_char(text, space_char, stitch_char):
+    if not text:
+        return text
+
+    new_chars = []
+    for i, j in zip(text, text[1:]):
```
morinted: Wondering why you used a loop instead of list comprehension:

```python
space_char.join(
    [stitch_char.join(list(word)) for word in text.split(space_char)]
)
```

morinted:
For testing: I'd like to see the edge cases where there is no space_char, where there is no stitch_char, where space_char == stitch_char, as well as trailing and leading space char, and maybe the test case where text is only a space char.

morinted:
Closing for now, we will likely add stitching after the formatting rework.

arxanas:
@gdwaynewarner Looks like @morinted implemented this in https://github.com/morinted/plover_stitching/!

PR #633, opened 2016-11-03
Plugins support: dictionaries

benoit-pierre:
Add support for dictionary plugins, e.g. [plover_python_dictionary](https://github.com/benoit-pierre/plover_python_dictionary) add support for Python dictionaries.


morinted:
This looks so awesome :)

PR #636, opened 2016-11-08
Plugins support: command

benoit-pierre:
Add support for custom commands to the plugin system. Note that those are only supported when output is enabled. See [here](https://github.com/benoit-pierre/plover/compare/plugins_support...benoit-pierre:priority_dict) for an example re-implementing #605.

### DIFF
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -266,6 +266,10 @@ class StenoEngine(object):
             self._trigger_hook('add_translation')
         elif command == 'LOOKUP':
             self._trigger_hook('lookup')
+        else:
+            command_args = command.split(':', 2)
+            command_fn = registry.get_plugin('command', command_args[0]).resolve()
+            command_fn(self, command_args[1] if len(command_args) == 2 else '')
         return False
 
     def _on_stroked(self, steno_keys):
--- a/plover/registry.py
+++ b/plover/registry.py
@@ -13,6 +13,7 @@ PLUGINS_DIR = os.path.join(CONFIG_DIR, 'plugins')
 class Registry(object):
 
     PLUGIN_TYPES = (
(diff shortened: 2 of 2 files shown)


morinted:
Is it correct that the plugins will share the same namespace as the top-level ones? e.g. `{PLOVER:LOOKUP}` is built in and `{PLOVER:PASTE}` is a plugin?

benoit-pierre:
Yes, same as for other plugins, so (new) builtin commands can be implemented the same way.

morinted:
It makes sense. I only wonder if the PLOVER prefix breaks down at a point. Originally it made sense with suspend, focus, etc., but even "lookup" is blurring the lines and then custom things like `{PLOVER:VLC}` might not make sense.

On the other hand, having `{COMMAND:VLC}` would be more confusing from a user's perspective, so I am okay with the PR as written.

KoiOates:
@benoit-pierre You wouldn't happen to still have that example plugin implementation, would you? I'm kicking myself for not copying it when you first made this pull request. Although my free time for tinkering with Plover has been and will be very limited for the next few months, I would like to continue working on plugin based projects in the future.

benoit-pierre:
Check [this commit](https://github.com/openstenoproject/plover/commit/9b24341e6075c4f85ba1d2f96f477b07bb24e7e0).

KoiOates:
@benoit-pierre Beautiful, thank you!

KoiOates:
Any reading you guys can recommend for getting a more complete understanding of setuptools? So far I've just been taking my time picking at the docs for setuptools and egg files (and reading through the diffs on these plugin pull requests). It's starting to make more sense, it's just a side of Python I've never tried to learn about before now.

PR #640, opened 2016-11-09
Release 3.1.0

morinted:


linux/README.md
```diff
@@ -60,15 +60,15 @@ To run from source, use: `./launch.sh`.
...
+`chmod +x dist/Plover-3.1.0-py2.7.egg && ./dist/Plover-3.1.0-py2.7.egg`
 
 Note: the egg file can be moved, but it cannot however be renamed (as the package version is encoded in the name).
 
 ### Using a wheel file with pip
 
 A wheel file can be created with: `./setup.py bdist_wheel`. The resulting file can then be installed using pip, e.g. for a user install:
 
-`pip2 install dist/Plover-3.0.0-py2-none-any.whl`
+`pip2 install dist/Plover-3.1.0-py2-none-any.whl`
```
benoit-pierre: The wheel name should be updated to `dist/plover-3.0.0-py2.py3-none-any.whl` (`Plover` -> `plover`, and universal wheel with both Python 2 and 3 support). Lower case plover for the egg too.

benoit-pierre: And the `app/dmg` for OS X too.

PR #651, opened 2016-12-02
Fix suggestions handling of empty text actions

### LINKED ISSUE #618: Suggestion dialog should handle corrections while fingerspelling
Classification: UI/Usability, Enhancement

Reproducibility: Always
## Suggestion dialog should handle corrections

When fingerspelling, the suggestion dialog does show the completed word and it's possible strokes. This can be used while learning to lookup a stroke. But when correcting a typo, the suggestion dialog does not show the complete word any more
## Steps to Reproduce
1. Stroke PW_, *E, TP_, O_, R_, R*
2. Stroke *
3. Stroke *E
## Expected Results
1. The Suggestion display should show "before" and all possible strokes
## Actual Results
1. The suggestion display shows only "beforr" and "e"
## Version

Current master (commit 709d9629536b15fb481135d8df3fb3ca8a7540fc)
## Configuration

OS:

Linuxmint 17.3

nimble0:
### About

Fix empty text actions erasing entire suggestions word string when being removed. There's a similar check in the loop a few lines down, so this was probably just missed whenever that change was made.

### Issues

Fixes #618

### DIFF
--- a/plover/gui_qt/suggestions_dialog.py
+++ b/plover/gui_qt/suggestions_dialog.py
@@ -112,7 +112,8 @@ class SuggestionsDialog(Tool, Ui_SuggestionsDialog):
     def on_translation(self, old, new):
         for action in old:
             remove = len(action.text)
-            self._words = self._words[:-remove]
+            if remove > 0:
+                self._words = self._words[:-remove]
             self._words = self._words + action.replace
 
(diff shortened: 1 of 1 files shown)


morinted:
Thanks so much for this fix!

PR #660, opened 2017-01-07
Better handle paragraphs in RTF dictionaries

morinted:
### About

I've noticed paragraph strokes don't behave the way stenographers expect them to. Now with the \n char, we can give a deletable paragraph, along with capitalization.

### DIFF
--- a/plover/dictionary/rtfcre_dict.py
+++ b/plover/dictionary/rtfcre_dict.py
@@ -93,13 +93,9 @@ class TranslationConverter(object):
         r'\\r|\\n'
         return JSON_PAR
 
-    def _re_handle_plover_combo(self, m):
-        r'{\\\*\\plover_combo ([^}]+)}'
-        return '{#%s}' % m.group(1)
-
-    def _re_handle_plover_command(self, m):
-        r'{\\\*\\plover_command ([^}]+)}'
-        return '{PLOVER:%s}' % m.group(1)
+    def _re_handle_plover_meta(self, m):
+        r'{\\\*\\plover_meta ([^}]+)}'
+        return '{%s}' % m.group(1)
 
     def _re_handle_infix(self, m):
         r'\\cxds (([^{}\\\r\n]|\\n|\\r|\\t)+)\\cxds ?'
@@ -326,24 +322,30 @@ def format_translation(t):
 
     t = t.replace(JSON_PAR, r'\par')
     t = t.replace(JSON_LINE, r'\line')
-    t = re.sub(r'{\.}', '{\\cxp. }', t)
-    t = re.sub(r'{!}', '{\\cxp! }', t)
-    t = re.sub(r'{\?}', '{\\cxp? }', t)
-    t = re.sub(r'{\,}', '{\\cxp, }', t)
-    t = re.sub(r'{:}', '{\\cxp: }', t)
-    t = re.sub(r'{;}', '{\\cxp; }', t)
-    t = re.sub(r'{\^}', '\\cxds ', t)
-    t = re.sub(r'{\^ \^}', '\\~', t)
-    t = re.sub(r'{\^([^^}]*)}', '\\cxds \\1', t)
-    t = re.sub(r'{([^^}]*)\^}', '\\1\\cxds ', t)
-    t = re.sub(r'{\^([^^}]*)\^}', '\\cxds \\1\\cxds ', t)
-    t = re.sub(r'{-\|}', '\\cxfc ', t)
-    t = re.sub(r'{>}', '\\cxfls ', t)
-    t = re.sub(r'{ }', ' ', t)
-    t = re.sub(r'{&([^}]+)}', '{\\cxfing \\1}', t)
-    t = re.sub(r'{#([^}]+)}', r'{\*\plover_combo \1}', t)
-    t = re.sub(r'{PLOVER:([a-zA-Z]+)}', r'{\*\plover_command \1}', t)
-    t = re.sub(r'\\"', '"', t)
+
+    substitutions = (
+        ('{\.}', '{\\cxp. }'),
+        ('{!}', '{\\cxp! }'),
+        ('{\?}', '{\\cxp? }'),
+        ('{\,}', '{\\cxp, }'),
+        ('{:}', '{\\cxp: }'),
+        ('{;}', '{\\cxp; }'),
+        ('{\^}', '\\cxds '),
+        ('{\^ \^}', '\\~'),
+        ('{\^([^^}]*)}', '\\cxds \\1'),
+        ('{([^^}]*)\^}', '\\1\\cxds '),
+        ('{\^([^^}]*)\^}', '\\cxds \\1\\cxds '),
+        ('{-\|}', '\\cxfc '),
+        ('{>}', '\\cxfls '),
+        ('{ }', ' '),
+        ('{&([^}]+)}', '{\\cxfing \\1}'),
+        (r'\"', '"'),
+        # Handle any other meta as an ignored group
+        (r'{([^}\\]+)}', r'{\*\plover_meta \1}'),
+    )
+    for (pattern, substitute) in substitutions:
+        t = re.sub(pattern, substitute, t)
+
     t = escape_translation(t)
     return t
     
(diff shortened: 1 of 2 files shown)


benoit-pierre:
`\n` must be handled on save too, otherwise some definitions will be mangled when saving back after a change.

test/test_rtfcre_dict.py
```diff
@@ -31,8 +31,8 @@ def test_converter(self):
         (r'\}', '}'),
         (r'\~', '{^ ^}'),
         (r'\_', '-'),
-        ('\\\r', '{#Return}{#Return}'),
-        ('\\\n', '{#Return}{#Return}'),
```
morinted: This covered the case of a line ending with a backslash, I'm not sure if this is in the RTF CRE spec. If it is, we can add this functionality back.

plover/dictionary/rtfcre_dict.py
```diff
@@ -307,8 +308,9 @@ def load_dictionary(filename):
           "{\\stylesheet{\\s0 Normal;}}\r\n")
 
 def format_translation(t):
-    t = ' '.join([x.strip() for x in META_RE.findall(t) if x.strip()])
```
morinted: Messing with the whitespace seemed odd. If whitespace doesn't matter, then why bother modifying it?

morinted:
Updated with comments about some assumptions I made, if you could review and confirm whether it all makes sense.

benoit-pierre:
My problem is:

```shell
> cat test.rtf
```
```
{\rtf1\ansi{\*\cxrev100}\cxdict{\*\cxsystem Plover}
{\stylesheet{\s0 Normal;}}
{\*\cxs AO*EPBG/EUFRPBLGTS}{\cxp, } Inc.? \{#Return\} \{#Return\} A \cxfc
{\*\cxs R}would you state your\~ name, please{\cxp.} \par
{\*\cxs STPH/STKPWHRAOE/-FPLT}{\cxp? } \{#Return\} \{#Return\} QYes.^ \cxfc
{\*\cxs TKORPBL}{\cxconf [{\cxc dorm}|{\cxc don't remember}]}
}
```
```shell
> ./convert.py test.rtf test.json && cat test.json
```
```json
{
"AO*EPBG/EUFRPBLGTS": "{,} Inc.? {#Return} {#Return} A {-|}",
"R": "would you state your{^ ^} name, please{.} {^\n\n^}{-|}",
"STPH/STKPWHRAOE/-FPLT": "{?} {#Return} {#Return} QYes.^ {-|}",
"TKORPBL": "don't remember"
}
```
```shell
> ./convert.py test.json test2.rtf && cat test2.rtf
```
```
{\rtf1\ansi{\*\cxrev100}\cxdict{\*\cxsystem Plover}{\stylesheet{\s0 Normal;}}
{\*\cxs AO*EPBG/EUFRPBLGTS}{\cxp, } Inc.? \{#Return\} \{#Return\} A \cxfc
{\*\cxs R}would you state your\cxds  \cxds  name, please{\cxp. } \cxds \n\n\cxds \cxfc
{\*\cxs STPH/STKPWHRAOE/-FPLT}{\cxp? } \{#Return\} \{#Return\} QYes.^ \cxfc
{\*\cxs TKORPBL}don't remember
}
```

The `R` definition should be unchanged, no?

test/test_rtfcre_dict.py
benoit-pierre: I think an escaped `\r` or `\n` is supposed  to be handled like `\par`.

morinted:
Do we want to convert back to \~ and to \par? And do \par's need a specified style like \par\s1?

morinted:
Ahh, just made some changes, head isn't clear right now so I'll come back to it later.

plover/dictionary/rtfcre_dict.py
morinted: Okay, but in this particular case this is just from JSON translation to RTF, so having the extra whitespace shouldn't be an issue (as \\n \\r and \\t are escaped)

test/test_rtfcre_dict.py
morinted: If we did this, then we wouldn't be able to save escaped newlines from Plover without it becoming paragraph strokes on next load.

morinted: Oh, I see what you mean, with `III.E Carriage Returns` in the spec.

benoit-pierre:
I think we should handle paragraphs, line breaks, and escaped `\n`/`\r` like this:
```
\line   {^\n^}
\par    {^~|\n\n^}
\n \r   same as \par
```
Corresponding patch:
```diff
diff --git i/plover/dictionary/rtfcre_dict.py w/plover/dictionary/rtfcre_dict.py
index 8ea0a9f6..faa15dd1 100644
--- i/plover/dictionary/rtfcre_dict.py
+++ w/plover/dictionary/rtfcre_dict.py
@@ -34,7 +34,8 @@ DICT_ENTRY_PATTERN = re.compile(r'(?s)(?<!\\){\\\*\\cxs (?P<steno>[^}]+)}' +
                                 r'(?=(?:(?<!\\){\\\*\\cxs [^}]+})|' +
                                 r'(?:(?:(?<!\\)(?:\r\n|\n)\s*)*}\s*\Z))')
 
-JSON_PAR = '{^\n\n^}{-|}'
+JSON_PAR = '{^~|\n\n^}'
+JSON_LINE = '{^\n^}'
 
 class TranslationConverter(object):
     """Convert an RTF/CRE translation into plover's internal format."""
@@ -90,7 +91,7 @@ class TranslationConverter(object):
 
     def _re_handle_escaped_newline(self, m):
         r'\\r|\\n'
-        return '\n'
+        return JSON_PAR
 
     def _re_handle_infix(self, m):
         r'\\cxds (([^{}\\\r\n]|\\n|\\r|\\t)+)\\cxds ?'
@@ -124,7 +125,10 @@ class TranslationConverter(object):
         if command == 'par':
             self.seen_par = True
             return JSON_PAR
-            
+
+        if command == 'line':
+            return JSON_LINE
+
         if command == 's':
             result = []
             if not self.seen_par:
@@ -310,9 +314,10 @@ HEADER = ("{\\rtf1\\ansi{\\*\\cxrev100}\\cxdict{\\*\\cxsystem Plover}" +
           "{\\stylesheet{\\s0 Normal;}}\r\n")
 
 def format_translation(t):
-    t = escape_translation(t)
     t = ''.join(META_RE.findall(t))
 
+    t = t.replace(JSON_PAR, r'\par')
+    t = t.replace(JSON_LINE, r'\line')
     t = re.sub(r'{\.}', '{\\cxp. }', t)
     t = re.sub(r'{!}', '{\\cxp! }', t)
     t = re.sub(r'{\?}', '{\\cxp? }', t)
```

Additionally maybe we should use an ignored control group to format things like key combos on save, e.g. something like `{#Return} {#Return}` would be saved back as `{\*\plover_combo #Return} {\*\plover_combo #Return}`.

morinted:
Okay, handled a few more thinks, I like the carry-capitalize for line breaking and used that extensively.

Also added commands and combos as ignored control groups.

morinted:
Handle all Plover metas as one instead of one for each type

morinted:
Updated with a more general meta catch-all

benoit-pierre:
More torture entries for round-tripping tests:
- `{^\}}`
- `{^}\}`
- `{\{^}`
- `\{{^}`

Note: those break in master too.

test/test_rtfcre_dict.py
```diff
@@ -210,6 +217,49 @@ def test_format_translation(self):
...
+                if states[-1] == states[-3]:
+                    settled = True
+                iterations += 1
+
+            self.assertEqual(expected, states[-1],
+                             err.format(initial=repr(initial),
+                                        end=repr(expected),
+                                        result=repr(states[-1]),
+                                        states=states))
+
```
benoit-pierre: I would prefer something simpler:
```python
    def test_round_trip_json(self):
        styles = {1: 'Normal', 2: 'Continuation'}

        convert = TranslationConverter(styles)

        err = (
            'Expected {initial} to turn into one of {expected} in round trip.\n' +
            'Instead, {result} was found. See process:\n{states}'
        )

        cases = (
            ('{^\n^}',),
            ('{#Return}',),
            ('{^^}', '{^}{^}'),
            ('{^test}',),
            ('{^toast^}',),
            ('Mr. {-|}',),
            ('Mr.{-|}',),
            ('state your{^ ^} name, please{^\n\n^}{-|}',),
            ('{PLOVER:ON}',),
            ('{^~|\n\n^}',),
            ('{^~|\n\n}',),
        )
        for case in cases:
            for translation in case:
                states = [translation]
                states.append(format_translation(states[-1]))
                states.append(convert(states[-1]))
                self.assertIn(states[-1], case,
                              err.format(initial=repr(states[0]),
                                         expected=repr(case),
                                         result=repr(states[-1]),
                                         states=states))
```

morinted:
Closing in favor of the RTF rework by Benoit Pierre.

PR #671, opened 2017-01-23
Tests don't fail if CONFIG_DIR is a symlink

### LINKED ISSUE #661: Test fails when Plover directory is a symlink
Classification: Bug

Reproducibility: Always

## Summary

When installing Plover (which involves running the tests), the installation fails because `MiscTestCase.test_dictionary_path` fails when `~/.local/share/plover` is a symlink.

## Steps to Reproduce

1. Run the tests

## Expected Results

1. Tests pass

## Actual Results

1. They don't

    =================================== FAILURES ===================================
    ______________________ MiscTestCase.test_dictionary_path _______________________

    self = <test.test_misc.MiscTestCase testMethod=test_dictionary_path>

        def test_dictionary_path(self):
            for short_path, full_path in (
                # Asset, no change.
                (ASSET_SCHEME + 'foo:bar',
                 ASSET_SCHEME + 'foo:bar'),
                # Absolute path, no change.
                (os.path.abspath('/foo/bar'),
                 os.path.abspath('/foo/bar')),
                # Relative path, resolve relative to configuration directory.
                (os.path.normpath('foo/bar'),
                 os.path.join(conf.CONFIG_DIR, 'foo', 'bar')),
                # Path below the user home directory.
                (os.path.normpath('~/foo/bar'),
                 os.path.expanduser(os.path.normpath('~/foo/bar'))),
            ):
                for input, function, expected in (
                    # Unchanged.
                    (short_path, 'shorten', short_path),
                    (full_path, 'expand', full_path),
                    # Shorten.
                    (full_path, 'shorten', short_path),
                    # Expand.
                    (short_path, 'expand', full_path),
                ):
                    function = '%s_path' % function
                    result = getattr(misc, function)(input)
                    self.assertEqual(result, expected, msg='%s(%r)=%r != %r'
    >                                % (function, input, result, expected))
    E               AssertionError: expand_path('/home/llama/.local/share/plover/foo/bar')='/home/llama/code/sh/dotfiles/.local/share/plover/foo/bar' != '/home/llama/.local/share/plover/foo/bar'

    test/test_misc.py:58: AssertionError
    =============== 1 failed, 132 passed, 1 xfailed in 1.87 seconds ================
    ==> ERROR: A failure occurred in check().
        Aborting...

The test fails because I have `~/.local/share/plover` symlinked to `~/code/sh/dotfiles/.local/share/plover`.

## Version

Plover version: Latest git (commit 6777999).

Installed via: AUR.

## Configuration

OS: Arch Linux.

tthurman:
MiscTestCase.test_dictionary_path() passes conf.CONFIG_DIR
through os.path.realpath() before using it, since the
functions it's testing do the same.

Fixes #661.

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -38,7 +38,7 @@ class MiscTestCase(unittest.TestCase):
              os.path.abspath('/foo/bar')),
             # Relative path, resolve relative to configuration directory.
             (os.path.normpath('foo/bar'),
-             os.path.join(conf.CONFIG_DIR, 'foo', 'bar')),
+             os.path.join(os.path.realpath(conf.CONFIG_DIR), 'foo', 'bar')),
             # Path below the user home directory.
             (os.path.normpath('~/foo/bar'),
              os.path.expanduser(os.path.normpath('~/foo/bar'))),


morinted:
We're currently reworking the dictionaries, would you test and see if the new tests pass with a symlinked config dir? https://github.com/openstenoproject/plover/pull/738

tthurman:
I confirm that the tests pass with this patch, and fail without it, both on upstream master and on the branch in  #738.

PR #666, opened 2017-01-09
Fix after spaces formatting for currency and fingerspelling

morinted:
### About

Three issues addressed:

- Capitalize needed same state persistence as lowercase for fingerspelling
- actions_to_spaces needed to handle the state persistence workaround
- The handling of the end of the word for currency needed a merge of word and text instead

### Issues

Fixes #606 and fixes #557

PR #679, opened 2017-01-31
Translation with more state

benoit-pierre:
Pass more state from the translator to the formatter so the output code can be simplified (by only taking into account an action text and replace fields). Supersede #666.


morinted:
I'm running with this patch for work today (spaces before). I might try spaces after tomorrow (but I feel as if any bugs I notice are par for the course…)

PR #682, opened 2017-02-07
add extension support to the registry

benoit-pierre:
Add support for extension plugins.

Minimal code for an extension:

```python
class Extension(object):

    def __init__(self, engine):
        pass

    def start(self):
        pass

    def stop(self):
        pass
```

An extension must be explicitly enabled by the user. It will be automatically initialized  and started/stopped when (re-)configuring the engine.

Use cases:
- monitoring external events (tracking application focus to automatically enable/disable Plover or maintain per application states)
- listening to a control socket for commands
- creating more complex plugins that are not state less (an extension could dynamically register other plugins)...

### DIFF
--- a/plover/config.py
+++ b/plover/config.py
@@ -83,6 +83,9 @@ SYSTEM_CONFIG_SECTION = 'System: %s'
 SYSTEM_KEYMAP_OPTION = 'keymap[%s]'
 SYSTEM_DICTIONARIES_OPTION = 'dictionaries'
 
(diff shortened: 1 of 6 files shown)


morinted:
Do you have a sample extension I could try? Doesn't have to be useful

benoit-pierre:
https://github.com/benoit-pierre/plover/compare/plugins_extension...benoit-pierre:extension_focus_track

morinted:
Thanks! Checking out the code now

morinted:
Works great, thanks!

PR #683, opened 2017-02-07
registry: add platform specific entrypoints

benoit-pierre:
- to allow for cleaner plugins code when implementations differ for each platform
- so a plugin can declare entrypoints for only supported platforms

Example: `plover.linux.command` instead of `plover.command` for declaring a Linux only command.

### DIFF
--- a/plover/registry.py
+++ b/plover/registry.py
@@ -1,4 +1,5 @@
 
+import sys
 import os
 
 import pkg_resources
@@ -8,6 +9,15 @@ from plover import log
 
 
 PLUGINS_DIR = os.path.join(CONFIG_DIR, 'plugins')
+
+if sys.platform.startswith('darwin'):
+    PLUGINS_PLATFORM = 'mac'
+elif sys.platform.startswith('linux'):
+    PLUGINS_PLATFORM = 'linux'
+elif sys.platform.startswith('win'):
+    PLUGINS_PLATFORM = 'win'
+else:
+    PLUGINS_PLATFORM = None
 
 
 class Registry(object):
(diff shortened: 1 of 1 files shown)


plover/registry.py
```diff
@@ -54,6 +55,17 @@ def update(self):
             entrypoint_type = 'plover.%s' % plugin_type
             for entrypoint in pkg_resources.iter_entry_points(entrypoint_type):
                 self.register_plugin(plugin_type, entrypoint)
+            if sys.platform.startswith('darwin'):
```
morinted: Wouldn't it be better to move this block out of the loop?

And perhaps we can stick with 'darwin' or 'mac' or 'macos' instead of 'osx', since that name is being retired. I'm undecided on which I like.

benoit-pierre: Done. I went with 'darwin' since that's the closest to `sys.platform`.

benoit-pierre: Although we do use 'osx' elsewhere (`osx/README.md`, `osxkeyboardcontrol.py`, ...).

morinted: Yeah, I've been thinking about that. I've wanted to replace our use of OS X throughout the code base.

macOS is kind of an obnoxious name, though. Especially the idea of `macoskeyboardcontrol`…but "mac" alone seems fine to me. Same character count as OS X and not going to be outdated.

benoit-pierre: "mac" it is then.

morinted:
Is the behavior undefined if the plugin exports both a bare entry point and a platform-specific one?

I'm reminded of our logging implementation where there is a default and then if a platform-specific one is found it overrides.

Wondering if you had considered this as a possibility (e.g., favor platform specific if it can be found)

benoit-pierre:
A platform-specific entry point will have priority, yes. However, the behavior is undefined if multiple plugins define the exact same entrypoint.

PR #696, opened 2017-02-21
gui_qt: update icons

morinted:
Visual refresh and selective renaming of icons. Summary of changes:

- All icons are same "size"
- Color palette reduced to subset of what it was, so settings icon, font selector, arrows, save are all more or less toned down.
- Arrow size changes to look less weird
- Add translation no longer features `STK` which was system-dependent and hard to make out at tiny sizes.

@benoit-pierre I'd love if you could test on your system to check:

- Size of icons (they might be too small now)
- That they all work (particularly the save icon that's given us trouble before)

Still TODO:

- For some reason some icons are huge and contain the sources for all the icons…


morinted:
screen shot 2017-02-21 at 9 23 07 am

benoit-pierre:
The dictionary edit/remove/add icons honestly do feel like a step back; the icons feel much less readable:
- they are (feel?) smaller
- the toned done palette and similar outlines make them harder to distinguish at a glance

Before:
2017-02-21-154802_325x445_scrot

After:
2017-02-21-154729_325x445_scrot

So far, no display errors (the save icon works fine).

morinted:
Thanks for the feedback, as you know the icons are huge on Mac so the readability wasn't so bad. I see your point.

For the green and red, maybe I could soften them a little? They are very vibrant right now.

I will try to make the icons a little bigger, too.

morinted:
Some improvements:

- Pencil icon has shading
- Icon sizes are bigger (back to normal)
- Fixed source sizes that caused jagged rendering (clearly visible upon zooming in on the OP)
- Less vibrant add/remove buttons
- Add translation changed to just "add" in the dictionary editor (makes more sense)

screen shot 2017-03-23 at 11 33 47 pm

plover/gui_qt/dictionary_editor.ui
```diff
@@ -164,7 +164,7 @@
  </widget>
  <layoutdefault spacing="5" margin="8"/>
  <resources>
-  <include location="resources/resources.qrc"/>
+  <include location="../../build/lib/plover/gui_qt/resources/resources.qrc"/>
```
benoit-pierre: Can those changes be reverted so the resource file is found when the build has not been run?

```bash
sed -e 's,../../build/lib/plover/gui_qt/resources/resources.qrc,resources/resources.qrc,' -i plover/gui_qt/*.ui
```

morinted:
Thank you for pointing that out -- fixed (with gsed because Mac's built in sed didn't work with your command :/ )

PR #702, opened 2017-02-27
drop Python 2 support on Windows / OS X

benoit-pierre:
Rational:
- Python 2 builds on those platform are narrow Unicode builds
- PyQt5 wheels are only available for Python 3
- those Python 2 builds are not really useful from a test standpoint, since the testsuite does not really test our platform specific code

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,10 +47,6 @@ before_install:
     set -ex
     # Bootstrap dev environment.
     bootstrap_cmd=(./bootstrap.sh)
-    if ! [ "$PYTHON" = 'python3' -a -n "$TRAVIS_TAG" ]
-    then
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -51,6 +51,13 @@ before_install:
     "${bootstrap_cmd[@]}"
     # List installed packages versions.
     $PYTHON -m pip list
+    if [ $TRAVIS_OS_NAME = osx ]
```
morinted: Could you comment on why this was necessary in case sometime in the future it's no longer necessary?

benoit-pierre: Done.

morinted:
LGTM!

PR #710, opened 2017-03-02
osxkeyboardcontrol: fix control key on macOS Sierra

morinted:
For some reason, after upgrading to Sierra, the control key doesn't act globally unless accompanied by flags that are present when pressing the physical control key.

Note: it does work locally (the current window catches events), but to use globals like Mission Control and Spectacle it needs this extra flag.

Also, as of Sierra having an output source seems to break control in globals too.

Not going to lie, it feels like magic.

TODO: try this on an older OS to make sure I didn't break anything.

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -107,6 +107,7 @@ DEADKEY_SYMBOLS = {
     'dead_tilde': '~',
 }
 
+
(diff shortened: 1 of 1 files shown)


morinted:
And to be clear, physically it seems impossible to send a control key without also having the secondary flag set.

morinted:
Confirmed working on 10.10 by Glen.

PR #711, opened 2017-03-04
Reset case action

Elktro:
### About

The code adds a new command {-!}, which cancels capitalize-next-word-command {-|}, and uppercasing {>} and lowercasing {<} commands. It also cancels capitalization action after stops such as the period {.}.

### Reason

Canceling case changing actions for the next word is very useful together with commands that for example result to cursor movement, such as {#Left}. This way moving cursor to different location to edit text will not result Plover capitalizing words in the middle of a sentence. Using {<} to achieve this, does not work as then a proper name, such as "Plover" will be decapitalized to "plover".

### Tested Platforms

The code has been tested with Ubuntu 16.04.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -384,6 +384,7 @@ META_CAPITALIZE = '-|'
 META_CARRY_CAPITALIZATION = '~|'
 META_LOWER = '>'
 META_UPPER = '<'
+META_RESET_CASE_ACTION = '-!'
 META_RETRO_CAPITALIZE = '*-|'
 META_RETRO_LOWER = '*>'
 META_RETRO_UPPER = '*<'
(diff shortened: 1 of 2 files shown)


benoit-pierre:
There's no need for a new action, as you can already do this by simply using `{}`.

Elktro:
> There's no need for a new action, as you can already do this by simply using {}.

Ha. That is true. No need for a new action. Sometimes the solution seems to be too simple to see.

Closing this pull request.

morinted:
@Elktro thanks for the PR. To be fair; it wasn't obvious to me that `{}` would reset the action in this way. Maybe you'd consider adding an to the **Dictionary Format** wiki page in the **Text Formatting** section, something like `Canceling formatting`

https://github.com/openstenoproject/plover/wiki/Dictionary-Format

If you'd prefer, I can write it up, just let me know 👍

Elktro:
@morinted Sure thing. I'll add a subsection about it as soon as I have more time.

Just wondering if there is then a way to define a dictionary entry that does absolutely nothing, not even reset actions? Adding an empty definition to a dictionary does not seem to be possible.

morinted:
The best I’ve come up with is to set it to {#}

That’s what I use for MOEJ in my emoji dictionary.

PR #712, opened 2017-03-05
Cherry picks for 3.1.1

benoit-pierre:
* lock down PyObjc version
* NKRO fix for Linux
* control key fix for macOS Sierra
* fix for missing modifiers in combos on macOS

Note: for some reason the Python 3 build on AppVeyor fails because of a permission error when installing pywin32... ¯\\_(ツ)_/¯

### DIFF
--- a/plover/gui/util.py
+++ b/plover/gui/util.py
@@ -6,9 +6,11 @@ import wx
 
 
 if sys.platform.startswith('win32'):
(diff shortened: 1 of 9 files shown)


benoit-pierre:
I went ahead and cherry picked the removal of pywin32 as a dependency too.

PR #714, opened 2017-03-06
Travis: update GitHub credentials

morinted:


morinted:
Note: pointed at master.

PR #722, opened 2017-03-10
log: initialize platform handler earlier

benoit-pierre:
Do it before loading plugins, so the user get some feedback on errors.

Note: this still leave Windows users without feedback if an error occurs during the initialization (no platform specific log handler).

### DIFF
--- a/plover/gui_none/main.py
+++ b/plover/gui_none/main.py
@@ -16,24 +16,6 @@ def show_error(title, message):
 
 def main(config):
 
-    handler_class = None
-    try:
-        if sys.platform.startswith('linux'):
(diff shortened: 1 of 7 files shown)


benoit-pierre:
It looks like we could use [plyer](https://pypi.python.org/pypi/plyer) for notifications on Windows:
2017-03-10-030616_1280x1007_scrot

benoit-pierre:
I pushed some commits to do just that.

benoit-pierre:
And it looks like the Windows build does not work... Looking into it.

benoit-pierre:
Using the new distribution works fine however, so I guess it's another PyInstaller issue?

morinted:
I love PyInstaller.

PR #723, opened 2017-03-10
windows: rework distribution creation

benoit-pierre:
Switch to using the official Python 3 embeddable zip file. Basically, instead of fighting with PyInstaller so everything we need is included, we just `pip install` Plover and all its dependencies, and then remove some things to make the final distribution lighter (mostly by removing some unnecessary Qt modules). This also mean that all standard library modules are present and accounted for, so they can be used by plugins.


benoit-pierre:
Updated `windows/README.md` section about building the distribution.

Issue #627, opened 2016-10-22
Custom Mac layout makes Plover crash


sammy-nes:
Classification: Crash/Hang/Data Loss

Reproducibility: Each time I open Plover with a custom keyboard layout made with Ukelele
## Summary

I have Plover crashing right after Plover icon bounces on the dock when **I have a custom keyboard layout enabled** (all in latin languages)
I have created a custom layout with Ukelele because I have my own keycaps arrangements on my keyboards. Same bug when when I launch it with some other pre installed non latin language layouts like Japanese, but some non latin language layouts works fine like Gurmukhi. 

When I enable any other latin pre installed layout it opens flawlessly.

<img width="805" alt="2016-10-22 17 40 37" src="https://cloud.githubusercontent.com/assets/22985030/19618213/bd66eeac-987e-11e6-86ce-37fd7d907cd0.png">

Thank you for your kind help
## Version

Plover version  v3.0.0+266.g9487982
## Configuration

OS: Mac Sierra 10.12

benoit-pierre:
Thanks, can you share the file for your custom layout?

sammy-nes:
Sure, here is the keyboard layout I am using. It is in the Library/Keyboard Layouts folder

[francemaster.keylayout.zip](https://github.com/openstenoproject/plover/files/545712/francemaster.keylayout.zip)

morinted:
Could you share this layout so that I can test it?

morinted:
Hi @sammy-nes -- I added code to fallback if we fail to read the keyboard layout (at the very least, prevents a crash!) -- would you try it? https://github.com/openstenoproject/plover/pull/724

I suspect key combos will fail to work (`{#command(c) command(v)}`). Maybe I can dig deeper and get something working.

morinted:
I think keyboard shortcuts will work now, too.

PR #724, opened 2017-03-11
osxkeyboardlayout: various fixes

### LINKED ISSUE #720: Deleting \_(ツ)_/ deletes too much text
Classification: Bug

Reproducibility: Always

## Summary

Trying to delete `\_(ツ)_/` also deletes a large amount of the preceding text. Also, my entry which reads as

```
"SHR*UG/SHR*UG": "¯\\_(ツ)_/¯",
```

In my dictionary file only renders as `\_(ツ)_/`, which is to say that the hands are missing for some reason. This might be related.

## Steps to Reproduce

1. Add entry for `\_(ツ)_/`. Mine is `SHR*UG/SHR*UG`.
2. Write some text.
3. Stroke `\_(ツ)_/`.
4. Stroke `*` to delete it.

## Expected Results

4. Only `\_(ツ)_/` is deleted. In my case, that would leave me with the word "shrug", which is the translation of `SHR*UG`.

## Actual Results

4. `\_(ツ)_/` is deleted, but also a great deal of the preceding text (usually the entire paragraph, but not necessarily the whole document, as far as I can tell). Then the word "shrug" is written out afterwards.

Here's a transcript from my `strokes.log`. In particular, note that the full form of the emoticon is written out in the log (with its hands) even though it isn't actually written out in the resulting output.

```
2017-03-09 15:20:05,198 Stroke(HEL : ['H-', '-E', '-L'])
2017-03-09 15:20:05,198 Translation(('HEL',) : "hell")
2017-03-09 15:20:05,508 Stroke(HROE : ['H-', 'R-', 'O-', '-E'])
2017-03-09 15:20:05,509 *Translation(('HEL',) : "hell")
2017-03-09 15:20:05,509 Translation(('HEL', 'HROE') : "hello")
2017-03-09 15:20:05,754 Stroke(-RGS : ['-R', '-G', '-S'])
2017-03-09 15:20:05,754 Translation(('-RGS',) : "{,}")
2017-03-09 15:20:06,032 Stroke(WORLD : ['W-', 'O-', '-R', '-L', '-D'])
2017-03-09 15:20:06,032 Translation(('WORLD',) : "world")
2017-03-09 15:20:06,828 Stroke(SHR*UG : ['S-', 'H-', 'R-', '*', '-U', '-G'])
2017-03-09 15:20:06,828 Translation(('SHR*UG',) : "shrug")
2017-03-09 15:20:07,058 Stroke(SHR*UG : ['S-', 'H-', 'R-', '*', '-U', '-G'])
2017-03-09 15:20:07,058 *Translation(('SHR*UG',) : "shrug")
2017-03-09 15:20:07,058 Translation(('SHR*UG', 'SHR*UG') : "¯\_(ツ)_/¯")
2017-03-09 15:20:07,695 *Stroke(* : ['*'])
2017-03-09 15:20:07,695 *Translation(('SHR*UG', 'SHR*UG') : "¯\_(ツ)_/¯")
2017-03-09 15:20:07,695 Translation(('SHR*UG',) : "shrug")
```

## Version

Plover version c3a4f96be7a9dec310641c2adea54b600f1be81f, built from source.

## Notes

The shrug rendered incorrectly in previous versions of Plover on OS X (which 7439d3a769e8855ebbb6c422e43dec13b46eaffd was presumably meant to fix). It now renders more correctly, in that there isn't a weird 0 in the sequence, but there are still no hands for the emoticon.

`git bisect` tells me that 6ff6a5912a55f1a9c42a83544c88af2f2980161f is the first bad commit. I can see it being a problem with control being held in some inappropriate scenario, since <kbd>ctrl+bksp</kbd> is "delete last word" in OS X. cc its author @morinted 

## Configuration

OS:

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.12.3
BuildVersion:   16D32
```

morinted:
### About

As usual, venturing into this file with a fresh mind allowed for some fixes.

- If a layout's Unicode mapping is None, like with Japanese, then get its ASCII representation. Fixes a crash.
- Keys with unknown deadkey states are ignored, allowing Hindi and sammy-nes to not crash...but the behavior may be undefined.
- Rework the favored key sequence logic to prefer longer sequences **if** the shorter sequence has a bad modifier compared to both of the other one.
- Fill in a missing piece of the layout reading that skipped some letters where the deadkey is explicitly set to nothing (odd but true?!).

### Issues

Fixes #720, #627 -- need to get confirmation from both authors.

### DIFF
--- a/plover/oslayer/osxkeyboardlayout.py
+++ b/plover/oslayer/osxkeyboardlayout.py
@@ -377,23 +377,28 @@ class KeyboardLayout(object):
                 if not isinstance(new_sequence[0], tuple):
                     new_sequence = new_sequence,
 
-                a_favored_b = favored_modifiers(
+                first_current_better = favored_modifiers(
                     current_sequence[-1][1], new_sequence[-1][1]
                 )
+                last_current_better = favored_modifiers(
+                    current_sequence[0][1], new_sequence[0][1]
+                )
 
+                # Favor key sequence with best modifiers (avoids a short sequence with awful modifiers)
+                # e.g. ABC Extended wants ¯ from (⌘⌥a,) instead of the saner (⌥a, space)
+                if first_current_better == last_current_better != 0:
+                    if first_current_better < 0:
+                        char_to_key_sequence[character] = new_sequence
                 # Favor shortest sequence (fewer separate key presses)
-                if len(current_sequence) < len(new_sequence):
-                    return
+                elif len(current_sequence) < len(new_sequence):
+                    pass
                 elif len(new_sequence) < len(current_sequence):
                     char_to_key_sequence[character] = new_sequence[0]
                 # Favor fewer modifiers on last item
-                elif a_favored_b < 0:
+                elif last_current_better < 0:
                     char_to_key_sequence[character] = new_sequence
                 # Favor lower modifiers on first item if last item is the same
-                elif (a_favored_b == 0 and
-                              favored_modifiers(
-                                  current_sequence[0][1], new_sequence[0][1]
-                              ) < 0):
+                elif last_current_better == 0 and first_current_better < 0:
                     char_to_key_sequence[character] = new_sequence
 
         def lookup_and_add(key, j, mod):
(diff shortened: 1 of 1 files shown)


morinted:
I've investigated better-supporting Hebrew, Hindi, and other vastly different languages.

My verdict: the behavior for now of writing QWERTY on these layouts is undefined. There's no easy way to detect them either, so Plover will just keep trying to work...

PR #728, opened 2017-03-15
Fix launch.sh to work on macOS

benoit-pierre:
No `readlink -e` on macOS.

### DIFF
--- a/launch.sh
+++ b/launch.sh
@@ -1,4 +1,15 @@
 #!/bin/sh
 
-cd "$(dirname "$(readlink -e "$0")")" &&
-exec ./setup.py launch -- "$@"
+set -e
+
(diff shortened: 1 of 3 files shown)


morinted:
We should probably apply a similar change to test.sh

benoit-pierre:
Maybe we should just assume `test.sh` is always run from the source directory? Unlike `launch.sh`, I don't see being able to symlink it somewhere in `$PATH` as a use-case we want to support.

morinted:
Agreed, so it'll just be the removal of the readlink?

benoit-pierre:
Yes, like `bootstrap.sh`, `test.sh` would need to be run from the source directory.

PR #731, opened 2017-03-21
Fix GUI tools' settings' persistence

nimble0:
### About

Add saving and restoring of ontop setting to suggestions dialog and paper tape using QSettings.

Fix restoring of paper tape style setting (was only restoring the text label).

Remove the config options for ontop and paper tape style from the main config.

### Reason

This restores behaviour lost when changing from wxWidgets to Qt.

### Tested Platforms

Linux Mint 17


nimble0:
Done.

PR #733, opened 2017-03-24
Fix Mac output edge cases

### LINKED ISSUE #732: Mac: sometimes enter does not dismiss Add Translation dialog
Sometimes, enter (both `\n` and `{#return}`) doesn't work for adding a translation.

Git bisect shows 6ff6a5912 as the culprit, probably the output source.

## Steps to Reproduce

So far only one case seems "reliable" to crop up this issue:

1. Make sure main plover window is open
2. In the Add Translation Translation field, enter a word
3. Press a stroke mapping to `{#left}`
4. Now you cannot use the Return key

When this state is enabled, tab actually produces a `\t` in the text box instead of changing fields. This all needs more investigation but I thought I'd write it down as a reminder.

morinted:
Fix #732 

I've given up trying to understand *why* this works the way it does, but through some cargo culting and testing I've found that:

- Command shortcuts *also* have the SecondaryFn Mask and global shortcuts require it (same as Control)
- The Output source None caused #732 as an issue -- I have no clue why. I reverted the output source, and mission control still works so I guess it was unnecessary in the first place.

### DIFF
--- a/plover/oslayer/osxkeyboardcontrol.py
+++ b/plover/oslayer/osxkeyboardcontrol.py
@@ -107,7 +107,6 @@ DEADKEY_SYMBOLS = {
     'dead_tilde': '~',
 }
 
-
 def down(seq):
     return [(x, True) for x in seq]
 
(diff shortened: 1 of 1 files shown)


morinted:
This seems to break Command-Shift-Tab

I think this points at the SecondaryFn flag *actually* being required for the arrows, not for the modifiers. I will further experiment.

PR #736, opened 2017-03-25
windows: add support for creating an installer

benoit-pierre:
The installer is created with [NSIS](http://nsis.sourceforge.net).

Features:
- install for all users or the current user only
- create a Plover group, with shortcuts for `plover.exe` and `plover_console.exe`
- register an uninstaller in the standard `Programs and Features`

Issues:
- no uninstallation of the previous version: the problem is I could not find a way to run the uninstaller and wait for it to finish... So if we want to support it, we'd need to use something clunky like: show an intermediate dialog, asking the user to press next only after the uninstaller as finished execution.


benoit-pierre:
Fixed the deployment on AppVeyor: deploy all generated artifacts.

morinted:
Something I noticed about the zip install is that the plover.ico doesn't have an extension on Windows, by default (which is a silly, but widespread, phenomenon)

Is is possible to move the ico into the data folder so that users using Plover portably will only be clicking on `plover` or `plover_console`?

benoit-pierre:
OK, done.

morinted:
Do we want to add the registry key fix for timeout to the un/installer?

morinted:
There should be a note in the windows README giving the command to build an installer, I think.

benoit-pierre:
> Do we want to add the registry key fix for timeout to the un/installer?

I don't have a patch for this, and I don't intend on spending more time on this.

> There should be a note in the windows README giving the command to build an installer, I think.

I have updated the documentation.

morinted:
🎊  Hurray!

PR #738, opened 2017-03-27
rework dictionaries handling

benoit-pierre:
* add support for enabling/disabling dictionaries
* change configuration entry, internal and default display order to be consistent: dictionaries are listed in the order they are searched
* add a configuration option to get back the legacy (higher priority dictionary last) display order
* add state icons to the dictionaries manager: a star for the "favorite" dictionary (first successfully loaded, enabled and writeable dictionary), an exclamation mark in case of load error (with a tool-tip detailing the exception), and a padlock for read-only dictionaries
* externally modified dictionaries are now automatically reloaded on engine configuration update


benoit-pierre:
Added workaround for icons/numbers placement on macOS.

plover/dictionary/loading_manager.py
```diff
@@ -44,12 +45,37 @@ def __init__(self, filename):
         self.result = None
         self.loading_thread.start()
 
+    def needs_reloading(self):
+        try:
+            new_timestamp = resource_timestamp(self.filename)
+        except:
+            # Bad resource name, permission denied, path
+            # does not exists, ...
```
morinted: *exist

plover/gui_qt/add_translation.ui
```diff
@@ -81,7 +81,7 @@
      <item row="0" column="1">
       <widget class="QComboBox" name="dictionary">
        <property name="editable">
-        <bool>true</bool>
```
morinted: This is a welcome change 👍

test/test_config.py
```diff
@@ -323,7 +358,7 @@ def test_as_dict_update(self):
         self.assertEqual(cfg.as_dict(), excepted_dict)
         update = {
             'auto_start'           : False,
-            'dictionary_file_names': [os.path.abspath('user.json')],
+            'dictionaries'         : [DictionaryConfig('user.json', False)],
```
morinted: would this be an array or a dict?

plover/gui_qt/add_translation.py
```diff
@@ -153,12 +151,61 @@ def _translation(self):
...
+        if reverse_order is not None:
+            self._reverse_order = reverse_order
+        iterable = self._dictionaries
+        if self._reverse_order:
+            iterable = reversed(iterable)
+        self.dictionary.clear()
+        for d in iterable:
+            item = shorten_path(d.get_path())
+            if not d.enabled:
+                item += ' [' + _('disabled') + ']'
```
morinted: I'm not sure I'm seeing this change...should it auto select the first enabled dictionary and label disabled ones?

With no selection and my first two dictionaries disabled I get this result

screen shot 2017-04-01 at 12 36 20 am

test/test_config.py
benoit-pierre: This how the configuration entry now looks like:
```
dictionaries = [{"enabled": true, "path": "user.json"}, {"enabled": false, "path": "commands.json"}, {"enabled": true, "path": "main.json"}]
```

plover/gui_qt/add_translation.py
benoit-pierre: That's because I messed up in the engine configuration update code when rewriting, so dictionaries are not actually disabled...

Note that the add translation dialog does not watch for configuration changes for the `dictionaries` variable, so the enabled/disabled state of dictionaries is not updated...

benoit-pierre:
Fixed typos, and now dictionaries are actually disabled by the engine when not enabled.

benoit-pierre:
I reworked the code:
- `StenoDictionaryCollection` is now more dict like, which simplifies the engine/GUI code a little
- a dictionary plugin is now is a StenoDictionary like class, most implementations will want to sub-class it and implement the `_load` and `_save` methods
- `StenoDictionary.casereverse` handling on deletion has been fixed
- `StenoDictionary.update` implementation is now a little faster

Things that still need to be improved: we should rework the dictionaries widget selection handling so some actions (edit, add translation, ...) are only available when at least one writable dictionary is selected.

morinted:
I'd expect that if I explicitly select a disabled dictionary from the Add Translation dialog, I should be able to add a stroke to it.

```
  File "/Users/ted/git/plover/plover/gui_qt/add_translation.py", line 247, in accept
    dictionary_path=dictionary)
  File "/Users/ted/git/plover/plover/engine.py", line 72, in _with_lock
    return func(self, *args, **kwargs)
  File "/Users/ted/git/plover/plover/engine.py", line 493, in add_translation
    self._dictionaries.set(strokes, translation, path=dictionary_path)
  File "/Users/ted/git/plover/plover/steno_dictionary.py", line 246, in set
    d = self[path]
  File "/Users/ted/git/plover/plover/steno_dictionary.py", line 268, in __getitem__
    raise KeyError(repr(path))
KeyError: "'/Users/ted/Library/Application Support/plover/main.json [disabled]'"
```

---------

This is feature creep, but thought I should voice it: Glen had requested, at one point, the ability to keep a dictionary open that will return results when looking up (e.g. in suggestion window), but would not be active. It *might* be worth having a reverse lookup that with a flag like `search_disabled=False` to support a feature like that -- and it wouldn't be default behavior. What do you think?

benoit-pierre:
> I'd expect that if I explicitly select a disabled dictionary from the Add Translation dialog, I should be able to add a stroke to it.

Fixed.

> This is feature creep, but thought I should voice it: Glen had requested, at one point, the ability to keep a dictionary open that will return results when looking up (e.g. in suggestion window), but would not be active. It might be worth having a reverse lookup that with a flag like search_disabled=False to support a feature like that -- and it wouldn't be default behavior. What do you think?

The core is already complicated enough, let's not add a new magic "searched but only in some cases" type of dictionary... If someone really want that feature, they can create a new suggestions tool plugin that does that.

morinted:
Great! This is such a fantastic PR.

PR #740, opened 2017-04-05
Mac embedded python distribution

morinted:
Use embedded Python in Mac App Bundle:
- No more PyInstaller
- Wider standard library support for plugins

Use dmgbuild to make the dmg:
- No more image mounting during build
- Image is available in 2x resolution
- Positioning works properly 10.10+ again

Probably has a blemish or two, but it's working and [here's a release](https://github.com/morinted/plover/releases/download/untagged-11ea740553b02848f749/plover-3.0.0.523.g6acfd93-py3.dmg)


morinted:
[Here's a zip](https://drive.google.com/file/d/0B_4v0_ER8UlwWnpGOGExV0VQNzA/view?usp=sharing) of the same app from Travis

osx/make_app.sh
```diff
@@ -0,0 +1,119 @@
...
+# $1: wheel name
+# $2: package name
+set -e
+
+. ./utils/functions.sh
+
+python='python3'
+osx_dir=$(dirname "$0")
+plover_dir=$(dirname $osx_dir)
+plover_wheel=$plover_dir/dist/$1.whl
```
benoit-pierre: It's better to quote variables to avoid unwanted expansion by the shell:
```bash
> mkdir 'unsafe name unsafe*'
> echo 'osx_dir=$(dirname "$0"); ls $osx_dir' > unsafe\ name\ unsafe\*/test.sh
> bash -x ./unsafe\ name\ unsafe\*/test.sh
++ dirname './unsafe name unsafe*/test.sh'
+ osx_dir='./unsafe name unsafe*'
+ ls ./unsafe name 'unsafe name unsafe*'
ls: cannot access './unsafe': No such file or directory
ls: cannot access 'name': No such file or directory
'unsafe name unsafe*':
test.sh
```

osx/make_app.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Remove broken symlinks
+rm $target_dir/Python.framework/Headers
+rm $target_dir/Python.framework/Python
+rm $target_dir/Python.framework/Resources
+# Trim superfluous content from app
+sed -e "s/\$python_version/$py_version/" -e "s/\$target_python/$target_python/" osx/app_resources/dist_blacklist.txt > build/dist_blacklist.txt
+$python -m utils.trim $target_dir/Python.framework build/dist_blacklist.txt
+# Make distribution source-less
+$python -m utils.source_less $target_libs */pip/_vendor/distlib/*
```
benoit-pierre: It's better to quote `*/pip/_vendor/distlib/*`.

setup.py
```diff
@@ -362,13 +357,18 @@ def finalize_options(self):
...
+        # Get dmgbuild
+        from utils.download import download
+        dmgbuild = download(
+          'https://github.com/morinted/dmgbuild/releases/download/v1.2.1%2Bplover/dmgbuild-1.2.1.pex',
+          '548a5c3336fd30b966060b84d86faa9a697b7f94'
+        )
+        subprocess.check_call(('chmod', '+x', dmgbuild))
         dmg = 'dist/%s.dmg' % PACKAGE
-        cmd = 'bash -x osx/app2dmg.sh %s %s' % (app, dmg)
+        cmd = '%s -s osx/dmg_resources/settings.py Plover %s' % (dmgbuild, dmg)
```
benoit-pierre: Why not use `cmd = 'python2 %s -s osx/dmg_resources...` so you don't have to chmod dmgbuild?

setup.py
```diff
@@ -324,29 +324,24 @@ def run_tests(self):
...
-            ('key'   , 'NSPrincipalClass'),
-            ('string', 'NSApplication'   ),
-        ):
-            element = Element(name)
-            element.text = value
-            element.tail = '\n'
-            dictionary.append(element)
-        tree.write(plist_info_fname)
+        self.run_command('bdist_wheel')
+        whl_cmd = self.get_finalized_command('bdist_wheel')
```
benoit-pierre: Why not use `whl_cmd.run()` so you don't have to call `self.run_command('bdist_wheel')`?

setup.py
```diff
@@ -324,29 +324,24 @@ def run_tests(self):
...
+        wheel_path = whl_cmd.get_archive_basename()
+        cmd = 'bash -x osx/make_app.sh %s %s' % (wheel_path, PACKAGE)
```
benoit-pierre: We probably don't need the `-x` anymore, no?

osx/make_app.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Strip 32-bit support
+ditto -v --arch x86_64 $app_dir $stripped_dir
+rm -rf $app_dir
+rm -rf $app_dist_dir
```
benoit-pierre: This should be done at the beginning, so only the results of the current script run are left (if the script failed, there should be no `$app_dist_dir`).

osx/make_app.sh
```diff
@@ -0,0 +1,119 @@
...
+# Absolute path to our python executable
+full_target_python=$target_dir/$target_python
+# The correct prefix for this Python (necessary because pip seems to get it wrong)
+local_sys_prefix=$($full_target_python -c 'import sys; print(sys.prefix)')
+# Install pip for this local python
+$python -c "
+from utils.download import download
+download('https://bootstrap.pypa.io/get-pip.py',
+         '3d45cef22b043b2b333baa63abaa99544e9c031d')
+"
```
benoit-pierre: Why not `$python -m utils.download 'https://bootstrap.pypa.io/get-pip.py' '3d45cef22b043b2b333baa63abaa99544e9c031d'`?

osx/make_app.sh
```diff
@@ -0,0 +1,119 @@
...
+$full_target_python $downloads/get-pip.py -f $wheels --prefix=$local_sys_prefix
+
+# Install Plover
+$full_target_python -m utils.install_wheels --ignore-installed --prefix=$local_sys_prefix $plover_wheel
+
+# Make launcher
+plover_executable=MacOS/Plover
+launcher_file=$app_dir/Contents/$plover_executable
+cp $osx_dir/app_resources/plover_launcher.sh $launcher_file
+sed -i '' "s/pythonexecutable/$target_python/g" $launcher_file
```
benoit-pierre: You could just use: `sed "s/pythonexecutable/$target_python/g" "$osx_dir/app_resources/plover_launcher.sh" >"$launcher_file"`

morinted:
Thanks for the great feedback! I've added a fixup commit -- if all is okay I will clean up the history.

setup.py
```diff
@@ -379,9 +379,9 @@ def run(self):
           'https://github.com/morinted/dmgbuild/releases/download/v1.2.1%2Bplover/dmgbuild-1.2.1.pex',
           '548a5c3336fd30b966060b84d86faa9a697b7f94'
         )
-        subprocess.check_call(('chmod', '+x', dmgbuild))
         dmg = 'dist/%s.dmg' % PACKAGE
-        cmd = '%s -s osx/dmg_resources/settings.py Plover %s' % (dmgbuild, dmg)
+        # Use Apple's built-in python2 to run dmgbuild
+        cmd = '/usr/bin/python %s -s osx/dmg_resources/settings.py Plover %s' % (dmgbuild, dmg)
```
morinted: Should I be concerned about spaces in `dmgbuild` or `dmg`?

osx/make_app.sh
```diff
@@ -0,0 +1,118 @@
...
+rm -rf "$stripped_dir"
+rm -rf "$app_dist_dir"
+
+# E.g. python3.5 (name of python executable)
+target_python="python${py_version}"
+# Main library folder in App
+target_dir="$app_dir/Contents/Frameworks"
+
+# Cleanup previous App and make skeleton
+rm -rf "$app_dir"
```
benoit-pierre: Already done above.

setup.py
benoit-pierre: Both `dmgbuild` and `dmg` are relative paths, and we (indirectly) set the basenames, so it should be okay.

osx/make_app.sh
morinted: Removed

setup.py
```diff
@@ -340,29 +340,24 @@ def run_tests(self):
...
-            element.text = value
-            element.tail = '\n'
-            dictionary.append(element)
-        tree.write(plist_info_fname)
+        whl_cmd = self.get_finalized_command('bdist_wheel')
+        whl_cmd.run()
+        wheel_path = whl_cmd.get_archive_basename()
+        cmd = 'bash osx/make_app.sh %s %s' % (wheel_path, PACKAGE)
+        log.info('running %s', cmd)
+        subprocess.check_call(cmd.split())
```
alexwlchan: Easier to write `cmd` as a list directly? i.e.

```python
cmd = ['bash', 'osx/make_app.sh', str(wheel_path), str(PACKAGE)]
...
subprocess.check_call(cmd)
```

Avoids any weirdness if one of those paths somehow ends up with a space in it.

osx/make_app.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+
+# Usage: ./setup.py bdist_app
+# $1: wheel name
+# $2: package name
+set -e
```
alexwlchan: Maybe at `set -o nounset`, so the script bails out if you don’t use it properly, or it hits another error? e.g. `greadlink` isn’t present on my machine, so `$python3_bin_dir` is empty but the script continues regardless.

benoit-pierre: `set -e` should already take care of that:
```bash
> bash <<\EOF; echo $?
#!/bin/bash
set -e
python3_bin_dir="$(dirname $(greadlink -f $(which python3)))"
echo 'next command'
EOF
bash: line 3: greadlink: command not found
dirname: missing operand
Try 'dirname --help' for more information.
1
```

morinted: I should account for the missing greadlink—thanks for the reminder

benoit-pierre: Would something like `python3 -c 'import os, sys; print(os.path.dirname(os.path.realpath(sys.executable)))'` work?

morinted: That works exactly. Thank you.

setup.py
morinted: We had this discussion: https://github.com/openstenoproject/plover/pull/740#pullrequestreview-33657417 but I can still change it if you insist

morinted: Hard to get a direct link, I think this one expands the outdated message: https://github.com/openstenoproject/plover/pull/740#discussion_r112349361

morinted:
History cleaned up.

setup.py
alexwlchan: If you’ve discussed it already, that’s fine – was just a suggestion. :)

PR #744, opened 2017-04-12
Fix inconsistent ordering of automatic suffix keys

nimble0:
### About

Fix system SUFFIX_KEYS getting put into a set, which loses all reliable ordering.

Order english stenotype SUFFIX_KEYS into reverse steno order (automatic suffixes which don't break steno order should be picked before those that do).

### DIFF
--- a/plover/system/__init__.py
+++ b/plover/system/__init__.py
@@ -2,6 +2,7 @@
 from io import open
 import os
 import re
+import collections
 
 from plover.oslayer.config import CONFIG_DIR, ASSETS_DIR
 from plover.registry import registry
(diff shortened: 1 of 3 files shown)


morinted:
The only concern I can see here is the chance of duplicated suffix keys from the system definition. Is that why we were using a set @benoit-pierre ?

benoit-pierre:
I don't see a reason for using a set, so the changes are fine. A test would be great too.

nimble0:
Are there existing stenotype system tests to add to?

nimble0:
Added a test and an assert that SUFFIX_KEYS is an ordered container.

morinted:
Thanks for figuring out this behavior and fixing it 👍

PR #747, opened 2017-04-19
A few Python tidies

alexwlchan:
### About

Some minor Python tidies:

* Remove unused imports and an unused variable
* Fix two places where `log` was being with `%`-formatting instead of comma-formatting.
* Fix a typo in a variable name.

### Reason

It doesn’t fix an issue, this is just general code hygiene.

### Tested Platforms

Just running a Python linter on OS X.

### DIFF
--- a/plover/dictionary/loading_manager.py
+++ b/plover/dictionary/loading_manager.py
@@ -3,12 +3,8 @@
 
 """Centralized place for dictionary loading operation."""
 
-import sys
 import threading
 import time
-
-# Python 2/3 compatibility.
-from six import reraise
 
 from plover.dictionary.base import load_dictionary
 from plover.exception import DictionaryLoaderException
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -57,7 +57,7 @@ def copy_default_dictionaries(dictionaries_files):
             continue
         # Check it's actually a default dictionary.
         basename = os.path.basename(dictionary)
-        if not basename in system.DEFAULT_DICTIONARIES:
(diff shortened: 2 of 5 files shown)


plover/log.py
```diff
@@ -9,7 +9,7 @@
 import traceback
 
 from logging.handlers import RotatingFileHandler
-from logging import DEBUG, INFO, WARNING, ERROR
```
benoit-pierre: This is so users of the API can do:
```python
from plover import log
log.set_level(log.DEBUG)
```

alexwlchan: Ah, gotcha. I’ll back that one out.

utils/zipdir.py
```diff
@@ -9,7 +9,7 @@ def zipdir(directory, compression=zipfile.ZIP_DEFLATED):
     zipname = '%s.zip' % directory
     prefix = os.path.dirname(directory)
     with zipfile.ZipFile(zipname, 'w', compression) as zf:
-        for dirpath, dirnames, filenames in os.walk(directory):
+        for dirpath, _, filenames in os.walk(directory):
```
benoit-pierre: I'd rather not use `_` for unused variables since we also use `gettext` to install `_()` as a builtin for translating strings. (There's only one use left of it for this in the Stentura code, other instances have been removed).

benoit-pierre:
The rest of the changes LGTM.

alexwlchan:
Okay, those two changes have been backed out.

morinted:
Thanks @alexwlchan and @benoit-pierre

PR #756, opened 2017-05-07
Add ability to receive plover commands from a named FIFO pipe

LukeSilva:
### About
Adds a command to enable, disable or toggle dictionaries as an engine command.
Adds a linux-only extension which reads from a named FIFO pipe, and runs each line written as an engine command.

### Reason
This feature allows other programs (eg: IDE / text editor) to change the state of plover, potentially enabling or disabling dictionaries.

A good example of this would be:
Text editor opens a .py file, tells plover to enable the python dictionary
Text editor opens a .c file, tells plover to disable the python dictionary, and enable the c dictionary

### Current issues
1. Does not allow the setting of the formatting MODE(s)
2. Git thinks I've merged in priority_dict when I haven't
3. The named pipe is always located at `/tmp/PLOVER_CMD` (is there some place where an extension can store it's configuration?)
4. There are no unit tests

### Tested on
The following shell script does force plover to run engine commands on Ubuntu 16.10
```
sleep 1
echo "RESUME" > /tmp/PLOVER_CMD
sleep 1
echo "ADD_TRANSLATION" > /tmp/PLOVER_CMD
sleep 1
echo "CONFIGURE" > /tmp/PLOVER_CMD
sleep 1
echo "LOOKUP" > /tmp/PLOVER_CMD
sleep 1
echo "FOCUS" > /tmp/PLOVER_CMD
sleep 1
echo "toggle_dict:-main.json" > /tmp/PLOVER_CMD
sleep 1
echo "toggle_dict:+main.json" > /tmp/PLOVER_CMD
sleep 1
echo 'toggle_dict:!main.json' > /tmp/PLOVER_CMD
sleep 1
echo 'toggle_dict:!main.json' > /tmp/PLOVER_CMD
sleep 1
echo "SUSPEND" > /tmp/PLOVER_CMD
sleep 1
echo "TOGGLE" > /tmp/PLOVER_CMD
sleep 1
echo "TOGGLE" > /tmp/PLOVER_CMD
```


benoit-pierre:
Thanks, but those 2 work-in-progress commits are not meant to be merged in Plover: they're not ready, and I plan to roll them into one plugin. Similarly, I invite you to create a dedicated plugin for you FIFO support. The goal is really to keep the core of Plover as minimal as possible (but we'll probably ship the next official with a set of default "blessed" plugins so you don't have to install all of them manually).

LukeSilva:
In that case, can I create another PR for the two engine commits squashed together? They're needed for the fifo_commands plugin to work.

benoit-pierre:
39d0970 is not right: `gui_qt.engine.Engine` and `gui_none.engine.Engine` already override `_in_engine_thread`

For a37d4e4: I am wondering if we need it. For commands triggered by strokes, we really do want to make sure they are run in the engine thread, so you can for example stroke in one language, switch system/dictionaries on the fly, and the new strokes are correctly handled with the new system/dictionaries, but in your case, just locking the engine could be enough:
```python
with engine:
   engine.send_engine_command('my command')
```
That is, once the suppression/pause/resume stuff has been reworked: so it's best to wait until that is done.

Issue #741, opened 2017-04-05
Uppercase last word also uppercases prefix of next word


arxanas:
Classification: Bug

Reproducibility: Always

## Summary

When using [the uppercase last word stroke](https://github.com/openstenoproject/plover/wiki/Dictionary-Format#uppercase-last-word) `{*<}`, if I uppercase the last word, and then type a prefix stroke for the next word, the prefix stroke also comes out as uppercase. Presumably, this should only happen for suffix strokes attached to the previous word.

## Steps to Reproduce

Consider my dictionary:

```json
"KPA*TS": "{*<}",
"TPAO": "foo",
"*ER": "{^er}",
"KAUPB": "{con^}",
```

Observe the following translations:

```
TPAO/KPA*TS/*ER/TPAO = FOOER foo      --> good, {^er} should be retroactively uppercased
TPAO/KPA*TS/KAUPB/TPAO = FOO CONfoo   --> bad, second word should be "confoo", not "CONfoo"
```

## Version

Plover version 4.0.0.dev0 @ 4c9567cdf41af70313d0b5c4fa5681070e72ec93 (current HEAD).

Installed via: source

benoit-pierre:
Fixed with #807.

Issue #719, opened 2017-03-08
Glue Strokes (&) not gluing when in "Spaces After" mode


mighele:
Classification: Bug

Reproducibility: Always

## Summary

When Plover is running with "Space placement: After output", glue strokes don't reliably glue together. Usually they will glue together 3 times and then not glue together (there will be a space in between items)

## Steps to Reproduce

1) My dictionary contains: "P*": "{&P}"
2) Set Plover to "Space placement: After output" in Output tab.
3) I press P* and release: letter "P" is output
Repeat step 3. Often it will correctly glue 3 P together but then will not glue correctly.
For example, after pressing P* 6 times your output will look like:
PPP P P P

## Expected Results

PPPPPP
(All Glue strokes are glued together with no spaces in between)

## Actual Results

PPP P P P
(Spaces in between some of the glue strokes)

## Version

Plover version 3.1.0.

Installed via: download from openstenoproject.org

## Notes

The glue functionality seems to work perfectly in "Spaces before output" mode. Maybe I'm one of a handful of users that prefer spaces after mode.

## Configuration

OS: Windows 10 Pro

morinted:
Thanks for the detailed bug report.

Benoit Pierre has been reworking the formatter code in a branch: https://github.com/benoit-pierre/plover/tree/formatter_reformatted

You can grab a test build off of AppVeyor: https://ci.appveyor.com/project/benoit-pierre/plover/build/1066/job/x48siycounn864m1/artifacts

Note: since it's a major version bump, I recommend you back up your plover.cfg

mighele:
That is a beautiful version of Plover! 👍 

Good work @benoit-pierre !

I've just checked and the glue strokes work fine on that version. Looking forward to it making it to the main branch!

benoit-pierre:
Already fixed on master.

PR #760, opened 2017-05-08
rework blackbox tests

benoit-pierre:
Rework blackbox tests to make it easier to add new ones. Also add tests for bug:

- #719:  pass on master, so it can be closed
- #741: xfail

### DIFF
--- a/test/test_blackbox.py
+++ b/test/test_blackbox.py
@@ -1,6 +1,9 @@
 
+import ast
+import functools
 import unittest
 
+from plover import system
 from plover.steno import Stroke
 from plover.formatting import Formatter
 from plover.translation import Translator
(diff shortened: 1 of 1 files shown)


test/test_blackbox.py
```diff
@@ -49,209 +103,210 @@ def setUp(self):
...
+        r'''
+        "KWEGS": "question",
+        "-S": "{^s}",
+        "TP-PL": "{.}",
+
+        KWEGS/-S/TP-PL
+
+        "questions. "
+        '''
+        print('arg')
```
morinted: Extra print statement?

test/test_blackbox.py
```diff
@@ -49,209 +103,210 @@ def setUp(self):
...
+        '''
+
+    @unittest.expectedFailure
+    @simple_replay
+    def test_bug741(self):
+        # Uppercase last word also uppercases next word's prefix.
+        r'''
+        "KPA*TS": "{*<}",
+        "TPAO": "foo",
+        "*ER": "{^er}",
```
morinted: Extra dictionary entry

test/test_blackbox.py
benoit-pierre: Oops...

test/test_blackbox.py
benoit-pierre: Fixed.

PR #761, opened 2017-05-17
Support Trisquel Linux in bootstrap.sh

hoijui:
works for me. :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,10 @@ lib64
 # Installer logs
 pip-log.txt
 
+# bootstrap.sh generated
+/requirements.txt
+/requirements_constraints.txt
+
 # Unit test / coverage reports
 .coverage
 .tox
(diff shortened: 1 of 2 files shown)


morinted:
Thanks for the PR. We are actually considering dropping the OS-specific handling of bootstrap.sh, as opposed to expanding it.

hoijui:
ok...
would it not still make sense to expand it first before removing it?
both for having it in history, and in case the removing is still some in the future, and thus it might be useful for someone in the meantime.
or do you expect merge errors with stuff you have already done in your code/repo?
either way, your choice of course.
not using it is also ok for me.

benoit-pierre:
The PR to remove `bootstrap.sh` has already been submitted (see #772), so the removal should happen sooner than later.

hoijui:
ahh i see, thank you, and i'll close this one then.
if i get a notification (which i should) on merge of your branch, i will check if it works with trisquel ,and add one of these custom version files if necessary.

PR #762, opened 2017-05-17
Mac: use circle CI

morinted:
### About

Use circle CI for Mac builds.

### Reason

- Artifacts are available
- No Travis queue

### TODO

- Cache GitHub releases tool
  - Update it, too…
- Remove Travis Mac build


morinted:
Not sure why it ran on morinted/plover instead of openstenoproject/plover…

This in progress PR was to test that openstenoproject actually had access to the Mac plan.

morinted:
Found a setting, will try retriggering it.

PR #765, opened 2017-05-21
machine/keyboard: fix a number of edge cases

benoit-pierre:
Instead of tracking which steno keys have been pressed/released, track the raw keys. This fix those edge cases (with + and - for press and release respectively):
- `+a +q -a` result in `S-` but a key is still held down
- `+a +r -a +a -r` result in `K-` but a key is still held down
- with arpeggiate mode enabled, `+a +space -space -a` does not work


benoit-pierre:
Added some tests.

benoit-pierre:
Labeled as "in-progress". We'll slightly rework tests.

PR #766, opened 2017-05-21
gui_qt/config: improve handling of machine specific options

benoit-pierre:
Look at the machine class, so it's possible to develop machine plugins that are not serial based, like the Tréal, or alternative keyboard machines.


benoit-pierre:
Yeah, right know, a tool is really the only way I can see to add a config dialog for another type of machine.

PR #767, opened 2017-05-21
main: add support for running console scripts

benoit-pierre:
This will allow access to plugins/packages scripts when running from a distribution.

### DIFF
--- a/plover/main.py
+++ b/plover/main.py
@@ -12,6 +12,8 @@ import sys
 import traceback
 import argparse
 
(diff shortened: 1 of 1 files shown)


morinted:
Looks good -- what's an example use case for this functionality?

benoit-pierre:
Access to internal Pover/plugin helper scripts (e.g. for converting dictionaries), or to other packages console scripts for debugging purposes (like pip).

PR #775, opened 2017-06-04
Separate Add Translation into widget and dialog

morinted:
Separate Add Translation into a widget and a dialog.

The dialog takes place of the previous tool, while the widget can be used by plugins who want to have the same functionality as Plover.

The most significant change is that `save_entry` on the AddTranslationWidget will return the change made.

This functionality would allow me to make my Dictionary Builder as a plugin.


benoit-pierre:
Not relying on `focusOut` events on the strokes/translation entries, and using instead loss of focus to the dialog window is no longer necessary, now that we're using Xinput/Xtest on Linux, and we don't suffer from transient focus loss/in events on output. So we can [simplify the code and keep all the focus/unfocus helper methods private to the widget](https://github.com/morinted/plover/compare/add_translation_break_apart...benoit-pierre:add_translation_break_apart).

PR #776, opened 2017-06-07
machine: drop Tréal support

benoit-pierre:
It has been moved to its own plugin.


requirements_plugins.txt
```diff
@@ -0,0 +1,7 @@
+# Plugins:
+plover-treal==1.0.0
+
+# Dependencies:
+hidapi==0.7.99.post20
```
morinted: Do we need to maintain this list manually?

benoit-pierre: I don't think there's another option: the list of requirements can vary — depending on the OS, Python version — when requirement markers are used... We can make it easier to find the list of requirements needed for each dependency however: https://github.com/benoit-pierre/plover/compare/no_treal...benoit-pierre:helper_for_updating_requirements

So to add a new standard plugin to the distributions:
1. create a new paragraph for in `requirements_plugins.txt`, only add a requirement for the plugin itself
2. push to GitHub, and check the output of listing the distribution packages

One issue: we don't build the Linux distribution when not deploying... And the standard plugins are not installed per default; because you can't do it at the same time as the other dependencies since they will depend on Plover... ¯\\_(ツ)_/¯

benoit-pierre: OK, so I added `utils/check_requirements.py`: it's used at the end of the distribution build to check the requirements (and print them, to make it easier to maintain them). This will load the plugins (without suppressing errors), so it can be used to detect some issues earlier (e.g. blacklisting some necessary PyQt5 files).

Note: our AppImage Docker image [needs to be updated](https://github.com/openstenoproject/plover-docker-images/pull/1).

osx/make_app.sh
```diff
@@ -108,6 +108,8 @@ sed -e "s/\$python_version/$py_version/" -e "s/\$target_python/$target_python/"
 "$python" -m utils.source_less "$target_libs" "*/pip/_vendor/distlib/*"
 
 # Strip 32-bit support
-ditto -v --arch x86_64 "$app_dir" "$stripped_dir"
+ditto -v --arch x86_64 "$app_dir" "$app_dist_dir"
```
morinted: No need to define $stripped_dir if it's not used

benoit-pierre: Amended.

PR #777, opened 2017-06-08
Added EclipseDictionary class for loading Eclipse .dix files.

tthurman:
### About

Added EclipseDictionary class to load .dix files, plus tests.

### Tested Platforms

Debian 9.0


morinted:
This is great, thanks for the PR!

We won't be able to merge it, though, as we are trying to move most new features into plugins, and this dictionary format is a great candidate for that. Not everyone will need this support so it makes sense to keep it as installable instead of bundled with Plover.

On the subject of the contents, I think having a working `save()` would be really great, even if we don't add much metadata.

arxanas:
Does anyone have an example dictionary file? I might try to port this to a Plover plugin (but no guarantees at all there).

PR #782, opened 2017-06-12
drop Python 2 support

benoit-pierre:
- apart from Linux, it does not make sense to use Python 2
  (we want at least Python 3.3, no narrow Unicode builds)
- this make life simpler for everybody (including plugins authors)


benoit-pierre:
Amended to drop unused `utils/metadata.py` instead of updating it.

benoit-pierre:
Amended again: stop creating a universal wheel.

PR #784, opened 2017-06-25
Added JetDictionary class for loading MS Access (.mdb, .dic) files.

tthurman:
### About

Adds the ability to load dictionaries from Microsoft Access files (.dic or .mdb).

Not tested with non-steno layouts; I don't have any sample dictionaries for such layouts.

### Tested Platforms

Debian stretch.


morinted:
I'm curious to see how your SQLite dictionary tests turn out.

I'd also like to help with making the read-only dictionary formats as plugins.

PR #792, opened 2017-08-04
Remove superfluous spaces to improve lookup.

Talljoe:
### About

Removed superfluous spaces from the dictionary.
### Issues

n/a

### Reason

There were several entries that were identical except for the spaces and these can be treated differently by Plover's lookup, etc. but are otherwise identical when outputting strokes (to my understanding).

### Release Notes

n/a

### Tested Platforms

Linux


plover/assets/main.json
```diff
@@ -721,7 +721,7 @@
 "-R": "are",
 "-RB": "{,}",
 "-RB/-RB/TPHR-RB": "{,}",
-"-RB/TPHR-RB/TK-RB": " -- ",
+"-RB/TPHR-RB/TK-RB": "--",
```
benoit-pierre: I think we should just remove this one instead of fixing it, since according to @stenoknight:
> And you can ignore all strokes that contain TPHR-RB. Those are legacy strokes from Eclipse's terrible time-based buffer system, but fortunately Plover doesn't have to deal with that.

plover/assets/main.json
```diff
@@ -734,7 +734,7 @@
 "-RBGS/SKWR-RBGS": "{^...}{-|}",
 "-RBGS/TK-RB": "--",
 "-RBGS/TOS/STPH": "{,}does it{?}",
-"-RBGS/TPHR-RB/TK-RB": " -- ",
+"-RBGS/TPHR-RB/TK-RB": "--",
```
benoit-pierre: Ditto.

plover/assets/main.json
Talljoe: While I was at it I removed all the other occurances as well.

PR #804, opened 2017-08-27
add plugins manager to the standard plugins

benoit-pierre:
* update PyQt5 to latest LTS version (5.9)
* add `plover-plugins-manager` to standard plugins: this means bundling-in the Qt web engine
* updated distributions Python version to 3.5.4
* PyQt5 system dependencies are no longer bundled, as this can cause issues on a number of distributions

Tested:
* AppImage on Arch Linux, Ubuntu Trusty/Xenial, and Fedora 22/25/Rawhide: starting the plugins manager on Fedora 25/Rawhide crash (SSL error, the distribution's pip does not work either...)
* Windows setup on Windows 10

### DIFF
--- a/ci/circle.sh
+++ b/ci/circle.sh
@@ -8,8 +8,8 @@ setup()
 {
   # Install Python.
   run mkdir -p "$downloads"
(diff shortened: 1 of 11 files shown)


morinted:
👏

PR #805, opened 2017-08-27
Rework blackbox/formatting tests

benoit-pierre:
Blackbox tests:
- make each test simpler
- improve error reporting
- shorten output size for failed tests
- support incremental checking of the output

Formatting tests: use pytest with parametrization so it's easier to select tests.

### DIFF
--- a/test/test_formatting.py
+++ b/test/test_formatting.py
@@ -3,1177 +3,1163 @@
 
 """Unit tests for formatting.py."""
 
-import unittest
+import inspect
+
+import pytest
 
 from plover import formatting
 
+from .test_blackbox import CaptureOutput
 
-# Add tests with partial output specified.
+
+class xfail(object):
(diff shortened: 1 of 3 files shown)


test/test_formatting.py
```diff
@@ -3,1177 +3,1163 @@
...
+
+class TestApplyCase(object):
+
+    test = ' some test '
+    test2 = 'test Me'
+    test3 = ' SOME TEST '
+
+    TESTS = (
+        # UNDEFINED
+        (test, '', Fa
```
morinted: Can we change the spelling to "optimization" because most of the code base is in American spelling?

morinted:
Looks good otherwise.

test/test_formatting.py
benoit-pierre: Done.

PR #807, opened 2017-08-30
rework formatting code

### LINKED ISSUE #554: Carry capitalization formatting edge case
Noticing this while coding lately...

``` python
    def test_carry(self):
        self.dictionary.set(('S-P',), '={^ ^}')
        self.dictionary.set(('PR-PB',), '{^~|()^}')
        for steno in (
            'S-P',
            'PR-PB',
        ):
            stroke = steno_to_stroke(steno)
            self.translator.translate(stroke)
        self.assertEqual(self.output.text, u' = ()')
```

```
>       self.assertEqual(self.output.text, u' = ()')
E       AssertionError: u' =()' != u' = ()'
E       -  =()
E       +  = ()
E       ?   +
```

benoit-pierre:
Greatly factorize the code, by changing the way spaces are handled: stitch actions at the output stage, so there's no need for 2 different code paths (space after and spaces before) when converting atoms to actions.

Also rework last word handling:
- for suffixes: store the raw un-cased word (sans prefix, as before)
- for other use-cases (retrospective commands): rely on an helper to generate the last n words from previous translations.

Fix #554 and #741.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -12,7 +12,216 @@ from collections import namedtuple
 import re
 import string
 
-from plover import orthography
+from plover.orthography import add_suffix
+
+
+CASE_CAP_FIRST_WORD = 'cap_first_word'
+CASE_LOWER = 'lower'
+CASE_LOWER_FIRST_CHAR = 'lower_first_char'
+CASE_TITLE = 'title'
+CASE_UPPER = 'upper'
+CASE_UPPER_FIRST_WORD = 'upper_first_word'
+
+SPACE = ' '
+META_ATTACH_FLAG = '^'
+META_CAPITALIZE = '-|'
+META_CARRY_CAPITALIZATION = '~|'
+META_COMMAND = 'PLOVER:'
+META_COMMAS = (',', ':', ';')
+META_GLUE_FLAG = '&'
+META_KEY_COMBINATION = '#'
+META_LOWER = '>'
+META_MODE = 'MODE:'
+META_RETRO_CAPITALIZE = '*-|'
+META_RETRO_FORMAT = '*('
+META_RETRO_LOWER = '*>'
+META_RETRO_UPPER = '*<'
+META_STOPS = ('.', '!', '?')
+META_UPPER = '<'
+MODE_CAMEL = 'CAMEL'
+MODE_CAPS = 'CAPS'
+MODE_LOWER = 'LOWER'
+MODE_RESET = 'RESET'
+MODE_RESET_CASE = 'RESET_CASE'
+MODE_RESET_SPACE = 'RESET_SPACE'
+MODE_SET_SPACE = 'SET_SPACE:'
+MODE_SNAKE = 'SNAKE'
+MODE_TITLE = 'TITLE'
+
+META_ESCAPE = '\\'
+RE_META_ESCAPE = '\\\\'
+META_START = '{'
+META_END = '}'
+META_ESC_START = META_ESCAPE + META_START
+META_ESC_END = META_ESCAPE + META_END
+
+META_RE = re.compile(r"""(?:%s%s|%s%s|[^%s%s])+ # One or more of anything
+                                                # other than unescaped { or }
+                                                #
+                                              | # or
+                                                #
+                     %s(?:%s%s|%s%s|[^%s%s])*%s # Anything of the form {X}
+                                                # where X doesn't contain
+                                                # unescaped { or }
+                      """ % (RE_META_ESCAPE, META_START, RE_META_ESCAPE,
+                             META_END, META_START, META_END,
+                             META_START,
+                             RE_META_ESCAPE, META_START, RE_META_ESCAPE,
+                             META_END, META_START, META_END,
+                             META_END),
+                     re.VERBOSE)
+
+# A more human-readable version of the above RE is:
+#
+# re.compile(r"""(?:\\{|\\}|[^{}])+ # One or more of anything other than
+#                                   # unescaped { or }
+#                                   #
+#                                 | # or
+#                                   #
+#              {(?:\\{|\\}|[^{}])*} # Anything of the form {X} where X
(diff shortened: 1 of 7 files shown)


benoit-pierre:
Known issues:
- [ ] set space placement to "after", `TEFT` -> "test ", `PHROPB` -> trailing space is briefly removed!

benoit-pierre:
The issue mentioned above is fixed. The code is ready for review (I'll squash the fixups once approved for merging).

benoit-pierre:
While working on a continuation of #809: adding 2 dedicated widgets for editing strokes/translation, it occurred to me that the new `Formatter.last_output_spaces_after` add state to the formatter, which means this need to be saved/restored when using a temporary state.

benoit-pierre:
Also, should a change to output settings (start attached/capitalized, space placement) impact a temporary state? Or be added to the previous (saved) running state and applied on restore?

morinted:
Can it apply to both?

On Sep 7, 2017 7:41 PM, "Benoit Pierre" <notifications@github.com> wrote:

Also, should a change to output settings (start attached/capitalized, space
placement) impact a temporary state? Or be added to the previous (saved)
running state and applied on restore?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/openstenoproject/plover/pull/807#issuecomment-327958124>,
or mute the thread

benoit-pierre:
I'm not sure they should impact the temporary state.

benoit-pierre:
Here's how the new engine's state API currently look in my `steno_edit` branch:

```python
    def push_new_state(self, start_attached=None, start_capitalized=None,
                       space_placement=None, dictionary_filter=None):
        ...

    def push_state(self, state):
        ...

    def pop_state(self, state):
        ...
```
When a steno edit widget gets the focus for the first time, `push_new_state` is called, and the returned state is saved. On focus loss, `pop_state` is called with the saved state (the reason for passing it back is detecting incorrect usage: someone trying to pop someone else state), and `push_state` when getting focus back. If one of the `start_attached/start_capitalized/space_placement` settings is not set, it's saved as such in the state object. When applying a state, a setting that is not set will get it's value from the current configuration. This way:
* the default running state will use `None`, so when enabled the current configuration settings will always be used
* the states used by steno edit widgets will override `start_attached/start_capitalized` to `True/False`
* a focus follow plugin could use both: states that use the default settings, and more specialized states for certain contexts

So without more specific changes to the code, modifying the current settings through the configuration dialog would indeed impact both the current state and the default running state.


Thoughts?

morinted:
Okay, you're proposing that the changing of settings should only affect the
saved state (original state?), and *new* temporary states will respect the
new defaults if they read from them normally.

But if you are changing from spaces after to before and have something like
7 windows open with the focus plugin, only the first state will change to
spaces before? That seems odd to me... or will all of the saved states get
the setting change and only the current temporary one won't?

Or is there a need for a warning that changes may not apply until a restart

benoit-pierre:
I see 2 use cases for the focus track plugin:
* for regular users, with no dedicated configuration, each state created by the plugin will use `None` for the settings, so those states will always respect the current configuration when enabled
* for more advanced users: support for specifying custom settings for some of those states (based on window name/class); those will override the configuration when the state is activated

Note that the substitution with the current configuration value when a state' setting is `None` occurs at on activation, when a state is enabled, and the value is not saved back to the state object, so it will be done every time the state is enabled.

A change of settings through the configuration dialog will always impact the current state. I guess the real question is: what happens to an active state setting if the corresponding configuration entry is changed?

Example:
* a new terminal window is opened
* the focus track plugin detects it, apply a specific user configuration:
```python
state['terminal'] = engine.push_new_state(start_attached=True, start_capitalized=False)
```
* the user then change the `start_attached` setting to False through the configuration dialog (note that the focus track plugin will ignore Plover's windows)
* this change the current formatter setting for `start_attached`
* the user switch to another application, then back to the terminal
* what are the current values for `start_attached` (for the formatter and `state['terminal']`)?

benoit-pierre:
Rebased to squash fixups and resolve conflicts.

benoit-pierre:
OK, I added back some `CASE_xxx` constants. I also tweaked the way the suggestions dialog is splitting text into words; so now when something like `full-release` is entered, suggestions for "release" will be shown too.

plover/formatting.py
```diff
@@ -12,7 +12,183 @@
...
+#                                   #
+#              {(?:\\{|\\}|[^{}])*} # Anything of the form {X} where X
+#                                   # doesn't contain unescaped { or }
+#             """, re.VERBOSE)
+
+
+WORD_RX = re.compile(r'((\d+([.,]\d+)+|\w+[-\w]*|[^\w\s]+)\s*)', re.UNICODE)
+
+
+class RetroFormatter(object):
```
morinted: I think this class could use an overall comment to explain why we need it, and maybe comments on the functions as to how they work since they could be used by plugins

plover/formatting.py
```diff
@@ -12,7 +12,183 @@
...
+        if not size:
+            return text
+        for fragment in self.iter_last_fragments():
+            text = fragment + text
+            if len(text) >= size:
+                break
+        return text[-size:]
+
+
+class _Context(RetroFormatter):
```
morinted: This one could also use a brief description to say what it's for

plover/formatting.py
```diff
@@ -93,41 +270,122 @@ def format(self, undo, do, prev):
...
+            last_action = prev[-1].formatting[-1]
+        else:
+            last_action = None
+
+        OutputHelper(self._output, self.last_output_spaces_after,
+                     self.spaces_after).render(last_action, old, new)
+        self.last_output_spaces_after = self.spaces_after
+
+
+class TextFormatter(object):
```
morinted: This one too

benoit-pierre:
Added some documentation.

Issue #773, opened 2017-05-28
RTF dictionary fails round-trip


gdwaynewarner:
**Classification**:  Bug

**Reproducibility**: Always

## Summary

Essentially, when I quit Plover and start it up again, some of my dictionary entries that have "keep together" attributes set (you know, that include the "^" symbol) all have  been modified such that outlines that should look like this --

`{^}&lt;p class="body">{^} {-|}`

-- end up looking like this:

`{^ &lt;p class="body"> ^} {-|}`

## Steps to Reproduce

1. Test outlines
2. Quit and restart Plover

## Expected Results

3. This should happen:

 The outline that looked like this before --

**{^}&lt;p class="body">{^} {-|}**

-- should look the same.

## Actual Results

I have screenshots, but I posted this issue -- complete with screenshots -- in the Plover Google group here:

https://groups.google.com/forum/#!topic/ploversteno/qo_vXTcJLxY

## Version

Plover version: 4.0.0.dev1+20.g60a373f

## Notes

There were other issues with the new build (lengthy pauses between what was stroked and what appeared on the screen), so I dropped back to version 3.0.0+328.g07e17fe, where I found the same issue with the dictionaries.

There was a suggestion to just use .json dictionaries, but while that might work, if the user is, as I am, coming from a regular CAT software background, chances are good they're not going to have a .json version of their dictionary. I know I don't, and I don't know of anything on the Mac that will reliably convert an .rtf file into a .json file ... though in truth, I think it has been about a year since I last checked, and the crap I found online just plain didn't work.

**If you've found a workaround, provide it here.**

I thought I could get away with copying the fixed version of the dictionary to another folder and replacing that dictionary with the copy just before starting Plover, but I still had the same issue, which leads me to believe that the issue occurs during the start-up routines.

## Configuration

ProductName:	Mac OS X
ProductVersion:	10.10.5
BuildVersion:	14F2315

morinted:
Until we rewrite our RTF implementation, I would define problem strokes in a supplementary JSON dictionary.

To make an empty JSON dictionary on a Mac:

- Open TextEdit
- In the menu bar, select "Format > Make Plain Text"
- Enter `{}` into the window 

![screen shot 2017-06-04 at 1 05 10 pm](https://cloud.githubusercontent.com/assets/5840970/26763752/bd217314-4926-11e7-843a-008f50f0d2b3.png)

- File > Save...
- Enter a file name, e.g. `glenzspecial.json`

[...]

PR #814, opened 2017-09-16
gui_qt/add_translation: fix widget translation label

benoit-pierre:
- escape translations/strokes to prevent interpretation (see: https://github.com/openstenoproject/plover/issues/773#issuecomment-306220676)
- style translations/strokes to make the whole text easier to read

Example:
2017-09-16-031342_427x305_scrot

### DIFF
--- a/plover/gui_qt/add_translation_widget.py
+++ b/plover/gui_qt/add_translation_widget.py
@@ -1,5 +1,6 @@
 
 from collections import namedtuple
+from html import escape as html_escape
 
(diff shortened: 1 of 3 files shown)


morinted:
Looking good. Could we also sort the mapped to results by length like the paper tape?

benoit-pierre:
OK, done.

gdwaynewarner:
Nice! My surprise bug is no more with this fix!

PR #818, opened 2017-09-16
Add support for creating dictionaries

benoit-pierre:
2017-09-17-013156_1920x1200_scrot
2017-09-17-013209_1920x1200_scrot

Note: the *"Create a new dictionary"* option will open a file save dialog, selecting an existing dictionary will prompt a confirmation dialog (*"main.json already exists. Do you want to replace it?"*): if the user confirms, the selected dictionary will be overwritten with a new one.


benoit-pierre:
Looks like Travis is getting confused by the force push (to remove a debug print)...

morinted:
I think the options could be:

- Open dictionary
- New dictionary

Seems to be the common wording in several programs I just checked (usually they don't include the noun, just "open" and "new" but with the + button the context might be lacking so it is good to be explicit)

benoit-pierre:
OK, I updated the labels to:
- Open dictionaries
- New dictionary

morinted:
This works on my machine, but it doesn't provide dictionary type hints.

When we tried to implement this previously, we offered the types in the format "JSON Dictionary (*.json)", see https://github.com/Germanika/plover/commit/e8aab554bb9832570dfe5937ff4fca6611bfd4cc#diff-ff7dcca020bebd5f065d5e0217640bf5R314

I suppose we'd have to filter on dictionary types that are not readonly.

This way less technically-inclined users can select the format without needing to type it from memory.

gdwaynewarner:
I think it would be good for the user to be able to select either .rtf or .json for the new dictionary's format.

Of course, once Plover's dictionary manager can export a .json dictionary as an .rtf, that can probably be done away with.

benoit-pierre:
@gdwaynewarner: that's already supported by this PR; the format is determined by the extension of the new filename.

benoit-pierre:
@ted: the problem is some platforms treat their users as idiots, and hide some information from them...

On my Arch Linux machine:
2017-10-01-195502_608x404_scrot

Under Unity with my Ubuntu Xenial VM:
2017-10-01-200327_580x405_scrot

I assume yours looks closer to the later?

Adding multiple filters is a good thing, e.g.:
- `Dictionaries (*.json *.rtf)`
- `JSON dictionaries (*.json)`
- `RTF dictionaries (*.rtf)`

But using the extension to get the name of the format is a kludge, ideally we'd use `RTF/CRE dictionaries (*.rtf)`, and `Python dictionaries (*.py)`. So it's not a valid fix for helping the user enter the correct extension...

We could automatically add the extension if the user does not specify it (but then we need another custom overwrite confirmation prompt if the file does already exists).

An alternative could be to not use the native file dialog for saving:
```diff
 plover/gui_qt/dictionaries_widget.py | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git i/plover/gui_qt/dictionaries_widget.py w/plover/gui_qt/dictionaries_widget.py
index b26a4a42..aa9b5e37 100644
--- i/plover/gui_qt/dictionaries_widget.py
+++ w/plover/gui_qt/dictionaries_widget.py
@@ -335,7 +335,9 @@ class DictionariesWidget(QWidget, Ui_DictionariesWidget):
 
     def _create_new_dictionary(self):
         new_filename = QFileDialog.getSaveFileName(
-            self, _('New dictionary'), None, _dictionary_filters()
+            self, _('New dictionary'), None,
+            _dictionary_filters(), None,
+            QFileDialog.DontUseNativeDialog,
         )[0]
         if not new_filename:
             return
```
Again, on Ubuntu Xenial:
2017-10-01-201951_641x415_scrot

benoit-pierre:
I just tried the automatic suffix support:
```python
        dialog = QFileDialog(self)
        dialog.setAcceptMode(QFileDialog.AcceptSave)
        dialog.setDefaultSuffix('json')
        dialog.setNameFilters(_dictionary_filters().split(';; '))
        if not dialog.exec_():
            return
        new_filename = dialog.selectedFiles()[0]
```
and this of course does not work correctly when using the native dialog: the suffix is automatically added, but no confirmation dialog is show if the file exists! *sigh*

morinted:
I prefer the native dialog if possible. I see the problem with naming as something we could handle on the plugin level, as metadata on the format.

It is concerning that the behavior falls apart on different distros. I remember it working very smoothly on Mac wrt selecting file types, but an "all dictionaries" option is pretty smart. I wonder how it behaves on Windows.

benoit-pierre:
Updated.

gdwaynewarner:
I might have missed it, but I think it would be much better if the dialogue wouldn't overwrite the user's dictionary, but maybe save the new dictionary as main_2.json or something like that. Chances are good that somebody wouldn't see that they were overwriting their dictionary, or maybe even believe that "overwrite" means make a new copy of their existing dictionary with a new name.

That's just not the kind of behavior one would want with such an important file as a dictionary file.

morinted:
For the dictionary create PR, it works a lot better if you don't set default suffix. Otherwise even if you purposely switch to RTF, type a name, and save, you'll get a .json dictionary unless you uncheck "hide extension" then backspace the ".json" part. It's really unusable.

I know platform-specific code sucks...but I don't think we should setDefaultSuffix on Mac (and keep the native overwrite dialog on this platform too?)

I'd like to see the behavior on Windows too.

morinted:
I tried on Windows, and I found that it worked equally well with and without the default format setting. I did notice a bug on Windows, though, that doesn't happen on Mac:

You can open the same dictionary multiple times and it gets listed multiple times in the dictionary list (including on overwrite)

benoit-pierre:
Well, no automatic suffix then! This make the code simpler...

Regarding the issue with Windows, could be because of different case? I'll see if I can reproduce.

benoit-pierre:
I've pushed a fix for the Windows issue.

benoit-pierre:
Now with passing tests...

morinted:
🎊 Looking great

PR #819, opened 2017-09-17
gui_qt: add support for machine option plugins

benoit-pierre:
A machine option plugin is a Qt widget that should follow the interface for GUI configuration options. When configuring a machine, the machine class (and its ancestor) will determine the plugin names that will be searched for the widget to use.

For example, when configuring the keyboard machine, the following list of names will be used to look for a compatible configuration widget:
- `plover.machine.keyboard:Keyboard`
- `Keyboard`
- `plover.machine.base:StenotypeBase`
- `StenotypeBase`
- `builtins:object`

Beside the 2 builtin plugins, another example use can be found in the [plover_midi](https://github.com/benoit-pierre/plover_midi/compare/machine_option) plugin.


plover/gui_qt/machine_options.py
```diff
@@ -0,0 +1,190 @@
...
+        self._update('timeout', timeout)
+
+    def on_xonxoff_changed(self, value):
+        self._update('xonxoff', value)
+
+    def on_rtscts_changed(self, value):
+        self._update('rtscts', value)
+
+
+class SerialOption(QWidget, Ui_SerialWidget):
```
morinted: This class is here twice in a row

benoit-pierre: Fixed.

PR #822, opened 2017-09-30
Update some requirements

benoit-pierre:
* PyObjC: 4.0 has been released
* pyserial: better handling of blocking errors and EINTR
* python-xlib: performance improvements
* wheel


benoit-pierre:
Ping!

benoit-pierre:
* update hidapi requirement: to 0.7.99.post21 so we can use pre-built wheel for Python 3.6.

benoit-pierre:
* update docutils requirement: fix a possible issue with restart on macOS.

PR #824, opened 2017-10-28
windows: fix distribution creation

benoit-pierre:
Update Python version to 3.6.3. This fix the AppVeyor build.

### DIFF
--- a/plover/main.py
+++ b/plover/main.py
@@ -13,7 +13,7 @@ import subprocess
 import traceback
 
 # This need to be imported before pkg_resources.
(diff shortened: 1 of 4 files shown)


benoit-pierre:
Reworked:
* fix missing dependencies (size is back to normal, instead of ~6MB...)
* fix plugins directory handling: method to set it was not compatible with the Python 3.6 embedded distribution

Note: I tested the plugins are working with a Windows 10 VM and on my Arch Linux install, this need to be tested on macOS too. Starting with `-l info` will now show the configuration and plugins directories for easier debugging.

PR #847, opened 2017-11-18
geminipr: flush input if stroke isn't valid

### LINKED ISSUE #846: Gemini PR protocol can become out of sync
## Summary

Sometimes, while @stenoknight is writing on her new Gemini PR machine, it starts producing inaccurate strokes, as if the number bar were hit. I'm guessing this is because the steno somehow gets offset and maybe S- becomes #, T- becomes S-, etc. I asked her to record more detail if it ever happens when she's not on a realtime job.

## Reproducing

The bug is intermittent, but happens with a newer machine over Bluetooth. She says there's a difference between the Plover versions:

- Plover 3 actually stops all output, and the reconnect button must be hit to start using the machine again.
- Plover 4 keeps going, but gets the wrong steno output.

I think this behavior is expected if you consider [this change](https://github.com/openstenoproject/plover/commit/39198e98b772914f189c51199e00da08ce73f11f#diff-0225703b0a36e618bf2b4357b9b5f730)

Perhaps a fix would be reintroducing the removed code in a state that doesn't produce an exception.

## Plover Version

- Latest weekly
- Plover 3.1.1

## System

Mirabai is running an Infinity Ergonomic Gen O on a Windows computer with the Toshiba Bluetooth stack.

morinted:
Fix #846 

Revert 39198e98b77291 and fix the code to not crash when it occurs.

I haven't tested that this works (Mirabai's problem is intermittent) but think it would be good to put in a weekly for testing. Alternatively we can hand her the CI build before merging.

### DIFF
--- a/plover/machine/geminipr.py
+++ b/plover/machine/geminipr.py
@@ -3,7 +3,10 @@
 
 """Thread-based monitoring of a Gemini PR stenotype machine."""
 
+import binascii
+
 from plover.machine.base import SerialStenotypeBase
+from plover import log
 
 # In the Gemini PR protocol, each packet consists of exactly six bytes
 # and the most significant bit (MSB) of every byte is used exclusively
@@ -44,6 +47,17 @@ class GeminiPr(SerialStenotypeBase):
             # Grab data from the serial port.
             raw = self.serial_port.read(BYTES_PER_STROKE)
             if not raw:
+                continue
+
+            # Make sure this is a valid steno stroke.
+            if (
+                len(raw) != BYTES_PER_STROKE
+                or not raw[0] & 0x80
+                or sum(1 for byte in raw if byte & 0x80) != 1
+            ):
+                log.error('invalid packet: %s (in waiting: %s)',
+                          binascii.hexlify(raw),
+                          binascii.hexlify(self.serial_port.read(self.serial_port.in_waiting)))
                 continue
 
             # Convert the raw to a list of steno keys.


benoit-pierre:
I don't think we should do that (blindly revert the code). I'd like some traces to understand what is happening exactly.

benoit-pierre:
This also won't fix the underlying issue, at best some strokes will be lost.

morinted:
It would probably be appropriate to try and resync by finding 0x80 and
getting that chord.

benoit-pierre:
But how can we lost `synchronization`? From what I understand, the machine should always send 6 bytes, are we loosing some? Are we loosing `synchronization` because of a short read? That's what I would like a trace to know what is happening exactly. Also it looks like the method was renamed: https://pythonhosted.org/pyserial/pyserial_api.html?highlight=flush#serial.Serial.reset_input_buffer

benoit-pierre:
How about adding this instead:
```diff
 plover/machine/geminipr.py | 12 ++++++++++++
 1 file changed, 12 insertions(+)

diff --git i/plover/machine/geminipr.py w/plover/machine/geminipr.py
index 80c4d1f1..179796d6 100644
--- i/plover/machine/geminipr.py
+++ w/plover/machine/geminipr.py
@@ -3,7 +3,10 @@
 
 """Thread-based monitoring of a Gemini PR stenotype machine."""
 
+import binascii
+
 from plover.machine.base import SerialStenotypeBase
+from plover import log
 
 # In the Gemini PR protocol, each packet consists of exactly six bytes
 # and the most significant bit (MSB) of every byte is used exclusively
@@ -46,6 +49,15 @@ class GeminiPr(SerialStenotypeBase):
             if not raw:
                 continue
 
+            if (
+                len(raw) != BYTES_PER_STROKE
+                or sum(n+1 for n, byte in enumerate(raw) if byte & 0x80) != 1
+            ):
+                log.error('invalid packet: %s (in waiting: %s)',
+                          binascii.hexlify(raw),
+                          binascii.hexlify(self.serial_port.read(self.serial_port.in_waiting)))
+                continue
+
             # Convert the raw to a list of steno keys.
             steno_keys = []
             for i, b in enumerate(raw):
```

morinted:
We aren't using pyserial 3.x though, right?

benoit-pierre:
We're using 3.4.

morinted:
Okay, updated with your patch, I'll send the AppVeyor build to Mirabai and hopefully we can get some data.

benoit-pierre:
So the problem is we read from the serial port with a timeout:
```
2017-11-22 09:18:22,619 [Thread-10-machine] ERROR: invalid packet: b'80' (in waiting: b'04200d2200')
2017-11-22 09:27:38,567 [Thread-10-machine] ERROR: invalid packet: b'80' (in waiting: b'')
2017-11-22 09:27:38,902 [Thread-10-machine] ERROR: invalid packet: b'000000040092' (in waiting: b'')
2017-11-22 09:27:39,235 [Thread-10-machine] ERROR: invalid packet: b'200000044480' (in waiting: b'')
2017-11-22 09:27:39,357 [Thread-10-machine] ERROR: invalid packet: b'00100c404480' (in waiting: b'')
2017-11-22 09:27:39,553 [Thread-10-machine] ERROR: invalid packet: b'000000020080' (in waiting: b'')
2017-11-22 09:27:40,687 [Thread-10-machine] ERROR: invalid packet: b'140000500080' (in waiting: b'')
2017-11-22 09:27:41,276 [Thread-10-machine] ERROR: invalid packet: b'000000040080' (in waiting: b'')
2017-11-22 09:27:41,448 [Thread-10-machine] ERROR: invalid packet: b'000000040080' (in waiting: b'')
2017-11-22 09:27:41,560 [Thread-10-machine] ERROR: invalid packet: b'000000040080' (in waiting: b'')
2017-11-22 09:27:41,692 [Thread-10-machine] ERROR: invalid packet: b'000000040080' (in waiting: b'')
2017-11-22 09:27:42,042 [Thread-10-machine] ERROR: invalid packet: b'000000040080' (in waiting: b'')
2017-11-22 09:27:42,290 [Thread-10-machine] ERROR: invalid packet: b'3b700e5e0080' (in waiting: b'')
2017-11-22 09:27:42,313 [Thread-10-machine] ERROR: invalid packet: b'3f700e5e00' (in waiting: b'')
```

So after a short read we can get out of order with the machine.

```diff
 plover/machine/geminipr.py | 19 +++++++++----------
 1 file changed, 9 insertions(+), 10 deletions(-)

diff --git i/plover/machine/geminipr.py w/plover/machine/geminipr.py
index bb8bc485..feff8f55 100644
--- i/plover/machine/geminipr.py
+++ w/plover/machine/geminipr.py
@@ -45,19 +45,18 @@ class GeminiPr(SerialStenotypeBase):
         while not self.finished.isSet():
 
             # Grab data from the serial port.
-            raw = self.serial_port.read(BYTES_PER_STROKE)
+            raw = self.serial_port.read(1)
             if not raw:
                 continue
 
-            # Make sure this is a valid steno stroke.
-            if (
-                len(raw) != BYTES_PER_STROKE
-                or not raw[0] & 0x80
-                or sum(1 for byte in raw if byte & 0x80) != 1
-            ):
-                log.error('invalid packet: %s (in waiting: %s)',
-                          binascii.hexlify(raw),
-                          binascii.hexlify(self.serial_port.read(self.serial_port.in_waiting)))
+            if not (raw[0] & 0x80):
+                log.error('invalid packet start: %s', binascii.hexlify(raw))
+                continue
+
+            raw += self.serial_port.read(BYTES_PER_STROKE - 1)
+
+            if len(raw) != BYTES_PER_STROKE or sum(b & 0x80 for b in raw[1:]):
+                log.error('invalid packet: %s', binascii.hexlify(raw))
                 continue
 
             # Convert the raw to a list of steno keys.
```

PR #859, opened 2017-11-24
add support for meta plugins

benoit-pierre:
Example use case: [timestamp support](https://github.com/benoit-pierre/plover_vlc_commands/blob/2409e978ededdaf628dbd3bd972a356d9873e7a9/plover_vlc_commands.py#L60) for the plover_vlc_commands plugin.


morinted:
Can you rebase this for merging?

benoit-pierre:
Done.

Issue #838, opened 2017-11-14
All Caps Last Word Behavior 


Wilfried84:
Classification: Usability

Reproducibility: Always

## Summary

When I write out a raw steno outline, I fingerspell it in lower case, with / and *, followed by an all caps stroke. In the previous version of Plover, it would make the whole thing all caps. Now it'll only upper case up to the last / or *. So to get the whole thing all caps, one has to hit the caps stroke before each / and *. Could we have a way to get back the old behavior? (I used Pover v. 3 until the latest weekly.)

## Expected Results

TEFT/TH/AOT/LAOIN

## Actual Results

teft/th/aot/LAOIN

## Version

weekly-v4.0.0.dev5+9.gba958374

Wilfried84:
To add, I discovered that all caps also stops at an apostrophe. So if you try to all caps don't, you end you with don'T. Using a capitalize stroke after it also results in don'T.

morinted:
This seems to be in the same vein as #836

Wilfried84:
Ah, didn't see that.

benoit-pierre:
It would be more efficient to use the [`show_stroke.py` Python dictionary](https://github.com/benoit-pierre/plover_python_dictionary/blob/master/test/show_stroke.py) for outputting raw strokes.

morinted:
Uppercasing the entire last word (up to spacing, treating symbols as part of the word) could be a plugin. I do think that there are easier ways to write raw steno than fingerspelling, though.

morinted:
No plans to implement this behavior at the moment. The show_stroke.py python dictionary should be adequate for writing raw steno.

Issue #837, opened 2017-11-14
No way to all-caps just the beginning of a word


arxanas:
Classification: Bug

Reproducibility: Always

## Summary

Suppose you have the dictionary

```
"PEUD": "pid",
"TPAOEUL": "file",
"KPA*TS": "{*<}",
"H*PB": "{^-^}",
"TK-LS": "{^^}",
```

and you try to write PID-file:

```
PEUD/KPA*TS/H*PB/TK-LS/TPAOEUL
```

This actually produces

```
PID-FILE
```

Instead of

```
PID-file
```

In general there doesn't seem to be a way to make capitalizing the previous part of the word not affect the next part of the word.

## Version

Plover version 4.0.0.dev5 @ ff51971a25c570dcc80146d6062cc4e28ed11864
Installed via: source

## Configuration

```
ProductName:    Mac OS X
ProductVersion: 10.12.6
BuildVersion:   16G29
```

benoit-pierre:
This is an intentional change from previous versions, the question is do we want something like this to work:
```python
    def test_retro_upper7(self):
        r'''
        'TEFT': 'test',
        "W-G": "{^ing with}",
        'RUP': '{*<}',

        TEFT/RUP/W-G  " TESTING with"
        '''
```
?

arxanas:
@benoit-pierre: What's the recommended way of writing a word like "PID-file", then? I generally use retroactive capitalization instead of preemptive capitalization, and it doesn't seem like there's a good way to accomplish this anymore, since I can't even use `TK-LS` to cancel the capitalization.

morinted:
I think the example in the OP is more realistic than the "test^ing with" issue. I'd probably still expect 
"TESTing with" due to the location of RUP.

arxanas:
This also appears to be manifesting when I'm coding. I tried to write this

```
  return SomeEnum::SOME_VALUE;
}

public
```

but got

```
  return SomeEnum::SOME_VALUE;
}

PUBLIC
```

where I typed

```
SOM/RUND/VAOUL/KPA*TS # SOME_VALUE
STPH*FPLT  # ;
STPH*B/STPH*B  # down-arrow a few times
R*R/R*R
PLIK
```

where `R*R` is defined to be `{MODE:RESET}{^}\n{^}`. There's even an explicit mode-reset in there. I also tried issuing explicit mode-resets with my `KLAO*ER` stroke.

[...]

PR #860, opened 2017-11-24
formatting: change word handling

### LINKED ISSUE #836: Capitalize-last-word doesn't work with {^'s}
Classification: Bug

Reproducibility: Always

## Summary

Suppose you have this dictionary:

```
"TPAO": "foo",
"KPATS": "{*-|}",
"AES": "{^'s}
```

Then I expect the sequence

```
TPAO/AES/KPATS
```

to produce

```
Foo's
```

but it actually produces

```
foo'S
```

which is to say that it capitalizes the "s" in the suffix rather than the real beginning of the word.

## Version

Plover version 4.0.0.dev5 @ ff51971a25c570dcc80146d6062cc4e28ed11864
Installed via: source

## Configuration

```
ProductName:    Mac OS X
ProductVersion: 10.12.6
BuildVersion:   16G29
```

benoit-pierre:
* don't carry on retro upper
* simplify word: split on whitespace, so the behavior is closer to pre-reformatted version
* better detection of word boundaries: extend check to all whitespace characters, not just space

Fix #836, #837, and #838.

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -83,7 +83,7 @@ META_RE = re.compile(r"""(?:%s%s|%s%s|[^%s%s])+ # One or more of anything
 #             """, re.VERBOSE)
 
 
-WORD_RX = re.compile(r'(?:\d+(?:[.,]\d+)+|\w+[-\w]*|[^\w\s]+)\s*', re.UNICODE)
+WORD_RX = re.compile(r'[^\s]+\s*', re.UNICODE)
 
 
 class RetroFormatter(object):
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -731,7 +731,7 @@ def _atom_to_action(atom, ctx):
             case = CASE_UPPER_FIRST_WORD
         text = _apply_case(text, case)
         if case == CASE_UPPER_FIRST_WORD:
-            action.upper_carry = not ' ' in text
+            action.upper_carry = not _has_word_boundary(text)
         # Apply mode.
         action.text = _apply_mode(text, action.case, action.space_char,
                                   action.prev_attach, ctx.last_action)
@@ -760,7 +760,7 @@ def _apply_meta_attach(meta, ctx):
         last_word and
         not meta.isspace() and
         ctx.last_action.orthography and
-        begin and (not end or ' ' in meta)
+        begin and (not end or _has_word_boundary(meta))
     ):
         new_word = add_suffix(last_word, meta)
         common_len = len(commonprefix([last_word, new_word]))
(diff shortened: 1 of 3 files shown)


arxanas:
I checked out this branch and I think I'm now having this issue which I haven't having before. I'm trying to write:

```
PREN/FAO/KPATS
```

where `KPATS` is my uppercase-previous-word stroke. I expect it to produce

```
(Foo
```

but it actually produces

```
(foo
```

---

I think I've just now discovered that if I write

```
KPA*/FAO/LO*ERLD
```

it doesn't affect the casing. That is, I expect

```
foo
```

but actually get

```
Foo
```

arxanas:
It also uppercased the wrong thing in this `/` expression:

```
Foo/bar
```

became

```
Foo/bar
```

instead of

```
Foo/Bar
```

This seems more arguable than other cases, but I imagine that it uppercased "bar" in the previous version or else I wouldn't have thought the behavior changed.

benoit-pierre:
OK, so the problem is previous to the big code changes in #807, the code for those retroactive commands would use `action.word`, which sort of is the current word, except prefixes are not included: `{prefix^}word{^ing}` -> the word is `wording`. The reason prefixes are not included is because `action.word` is also used by the orthographic rules (so `{prefix^}cancel{^ing}` will yield `prefixcanceling` even though it's not in the word list, because `canceling` is).

Obviously, when using for example retro-upper, the prefix should be included too, right? So the code does not use `action.word` anymore, and instead split the previously translated text into words with a regular expression; hence the changes in behavior compared to older versions.

So the big question is how do we define a word with respect to the formatter? People clearly expect different results (compare #838 with your expectations above). The reality is we can't make the default behavior work the way people want it in all cases (and fix the problem with prefixes not being included), but:
- #859 will allow for custom metas
- the current code should make it easier to have a retro-upper that can act on more than just the last word (.e.g.: `{:retro_upper:4}` to retro-upper the last 4 words)

I've made #865, with an alternative, which might be better if English (and not programming) is the main case.

morinted:
I do get the impression that the #838 use case is not a valid one; the last word wouldn't go past the slash. I think #865 makes more sense and that sharing the behavior will make for a more consistent experience. I still need to pull the PR and try it for a little.

arxanas:
So far this PR is doing the right thing for me. I earlier stated I wanted retro-uppercase on `foo/bar` to produce `Foo/bar`, but now I think that's the wrong move. This PR does `foo/Bar` which seems like the most reasonable option.

arxanas:
I commented on the wrong PR! Good thing you merged the correct one anyways.

PR #865, opened 2017-12-01
formatting: change word handling, take 2

### LINKED ISSUE #836: Capitalize-last-word doesn't work with {^'s}

benoit-pierre:
Alternative too #860.

* don't carry on retro upper
 * more consistent handling of words: use the same word definition for `_rightmost_word`
 * better detection of word boundaries: extend check to all whitespace characters, not just space

Fix #836 and #837, but **not** #838 (with this implementation the behavior is expected).

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -731,7 +731,7 @@ def _atom_to_action(atom, ctx):
             case = CASE_UPPER_FIRST_WORD
         text = _apply_case(text, case)
         if case == CASE_UPPER_FIRST_WORD:
-            action.upper_carry = not ' ' in text
+            action.upper_carry = not _has_word_boundary(text)
         # Apply mode.
         action.text = _apply_mode(text, action.case, action.space_char,
                                   action.prev_attach, ctx.last_action)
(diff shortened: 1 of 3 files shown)


morinted:
Also has conflicts with translator rework PR

benoit-pierre:
Rebased.

arxanas:
Trying to retro-uppercase `YERBGS` = `yes{,}` results in `yes,` instead of `Yes,`. Is this expected behavior under this PR?

benoit-pierre:
@arxanas: yes, since the last word is `,`.

PR #866, opened 2017-12-01
Improve Gemini PR, Passport, and Procat support

### LINKED ISSUE #846: Gemini PR protocol can become out of sync

benoit-pierre:
Add an `_iter_packets(packets_size)` helper to `SerialStenotypeBase`: yield packets of `packets_size` bytes until the machine is stopped. To workaround the fact that the Toshiba Bluetooth stack on Windows does not correctly handle the read timeout setting (returning immediately if some data is already available):
- the effective timeout is re-configured to `timeout/packet_size`
- multiple reads are done (until a packet is complete)
- an incomplete packet will only be discarded if one of those reads return no data (but not on short read)

Update the Gemini PR code to use this new method and reject invalid packets. Fix #846.

Note: the Procat code should probably be updated to do the same.

### DIFF
--- a/plover/machine/base.py
+++ b/plover/machine/base.py
@@ -6,6 +6,7 @@
 
 """Base classes for machine types. Do not use directly."""
 
+import binascii
 import threading
 
 import serial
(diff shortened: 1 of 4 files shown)


benoit-pierre:
And the Passport code should probably do the same as the Tx Bolt, i.e.:
```python
raw = self.serial_port.read(max(1, self.serial_port.inWaiting()))
```

benoit-pierre:
Updated to improve support for the following machines:
* Procat: use the new method for getting packets, discard invalid packets
* Passport: do not hog the CPU constantly trying to read 0 bytes of data from the serial port when no data is available....

PR #886, opened 2017-12-31
Config rework

benoit-pierre:
Right now, configuration settings are only validated for correctness when loading from the configuration file. While this is not really an issue when those are changed through our GUI, plugins can now update the configuration (e.g. the `plover_dict_commands` plugin).

This PR overhaul the way the configuration is handled so:
- all custom set/get accessors are removed in favor of a standard mapping API:
```
config.set_machine_type("Keyboard") -> config["machine_type"] = "Keyboard"
```
- all options are validated on set/update too
- updates are all or nothing: nothing is updated if one of the options is invalid

### DIFF
--- a/plover/config.py
+++ b/plover/config.py
@@ -3,17 +3,18 @@
 
 """Configuration management."""
 
-from collections import namedtuple
+from collections import ChainMap, namedtuple, OrderedDict
 import codecs
 import configparser
 import json
 import os
+import re
 
 from plover.exception import InvalidConfigurationError
 from plover.machine.keymap import Keymap
 from plover.registry import registry
 from plover.oslayer.config import CONFIG_DIR
-from plover.misc import expand_path, shorten_path
+from plover.misc import boolean, expand_path, shorten_path
 from plover import log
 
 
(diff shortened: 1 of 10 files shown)


plover/config.py
```diff
@@ -120,421 +63,340 @@ def from_dict(d):
...
+        self._config.write(writer)
 
+    def _set(self, section, option, value):
+        if not self._config.has_section(section):
+            self._config.add_section(section)
+        self._config.set(section, option, value)
 
     # Note: order matters, e.g. machine_type comes before
     # machine_specific_options and system_keymap because
     # the laters depend on the former.
```
morinted: "the latter depend on the former"

plover/misc.py
```diff
@@ -55,3 +55,13 @@ def normalize_path(path):
     if path.startswith(ASSET_SCHEME):
```
morinted: In 1e443041c18c6 commit message the grammar could be improved with "follows": `[...] convert a string as follows`

benoit-pierre:
Amended.

PR #906, opened 2018-02-28
Fix Arch Linux & Fedora Rawhide packages dependencies

benoit-pierre:


benoit-pierre:
Looks like AppVeyor does not like the branch name...

PR #918, opened 2018-03-12
Add euro sign to key_combo.py

mkrnr:


morinted:
We already support Unicode in the translation as documented in [Sending symbols](https://github.com/openstenoproject/plover/wiki/Dictionary-Format#sending-symbols)

To support {#EuroSign} we would need to import currency from [keysimdef.h](https://www.cl.cam.ac.uk/~mgk25/ucs/keysymdef.h) but for now I don't think we have a use case (unless @benoit-pierre can think of a good reason?)

In any case, thanks for the pull request!

Issue #907, opened 2018-03-02
Plover should have a safe mode to start without any external plugins enabled


morinted:
## Summary

To allow users to escape a broken or malicious plugin, Plover should have the ability to start without any external plugins.

benoit-pierre:
Fixed by #920: when running from a distributions, user plugins can be disabled by starting with `--no-user-plugins`.

PR #920, opened 2018-03-18
Rework plugins directory handling

benoit-pierre:
In order to support #907 and fix https://github.com/benoit-pierre/plover_plugins_manager/issues/4, this PR drop the custom plugins directory handling in favor of a dedicated launcher for distributions:
* when Plover is installed with pip or a distribution package manager, then user plugins are assumed to be installed in the standard user site (and can be installed/updated/removed from there with the plugins manager)
* when Plover is run through a distribution, then we want to be isolated from an existing Python installation: `PYTHONUSERBASE` is set according to Plover's configuration directory (so portable mode is respected). In this mode, user plugins can be disabled by launching with `--no-user-plugins` (the embedded distribution plugins will still be loaded).

This will require a new plugins manager (>= 0.5.11.dev0). Note that this new manager supports handling user plugins even if the user site is disabled (so broken plugins can still be updated/uninstalled).

One question though, when restarting after Plover was launched with `--no-user-plugins`, should user plugins be still disabled?

### DIFF
--- a/linux/appimage/apprun.sh
+++ b/linux/appimage/apprun.sh
@@ -59,7 +59,7 @@ appimage_uninstall()
 
 appimage_launch()
 {
-  exec "${APPDIR}/usr/bin/python3.6" -m plover.main "$@"
+  exec "${APPDIR}/usr/bin/python3.6" -s -m plover.dist_main "$@"
 }
 
 set -e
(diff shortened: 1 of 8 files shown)


benoit-pierre:
Note: the CI failures are expected, since the plugins manager need to be updated (and Plover development version incremented), except I have no idea why `check_requirements` cannot import `pkg_resources` on Circle CI...

benoit-pierre:
OK, now it fails as expected.

PR #924, opened 2018-03-22
Select a keyboard control module for OpenBSD.

tlevine:
Note that this change is not sufficient for running plover on OpenBSD;
one must also remove the PyQt5 dependency check.
https://groups.google.com/forum/#!topic/ploversteno/dR5OKtL1Lv8

I could not find a Github Issue corresponding to OpenBSD support.


benoit-pierre:
I'm going to wait and not merge this: I'll revisit OpenBSD support later when changes to output handling have been integrated. I've created an issue to track what's needed: #931

PR #936, opened 2018-04-10
Rework all tests to use pytest

benoit-pierre:
Note:
* the use of `lambda` to capture the line of each definition when parametrizing a test (so it's easier to locate a failing test)
* the way key combo tests are handled can seem a little complicated, but this is in prevision of updating the tests when support for locking keys is added


benoit-pierre:
* fix engine tests
* fix configuration tests random failures                                                                                                                                             
* fix `Config.__setitem__` implementation
* only create strokes/translations log when needed

test/__init__.py
```diff
@@ -1,6 +1,9 @@
```
morinted: I think the "tests: add parameterize helper" commit should go first, before it's used in other commits.

benoit-pierre: It does, but Github reorder commits by date... ¯\\_(ツ)_/¯

morinted: Oh, that's arbitrary, all right… guess it's all good then

PR #942, opened 2018-04-23
Fix Unicode characters outside of BMP on Windows.

zwagoth:
1. Windows requires that you split Unicode characters into surrogate pairs before submission to SendInput, if they are outside the BMP as the maximum size of the character point in SendInput is 16bits.

2. Split incoming Unicode characters into surrogate pairs if they are outside the BMP. This was done by hand based on the Unicode specification because Python appears to offer no means of splitting a surrogate pair without encoding to UTF-16, removing the BOM, and then manually cutting the bytes out and putting them back together.

### DIFF
--- a/plover/misc.py
+++ b/plover/misc.py
@@ -65,3 +65,15 @@ def boolean(value):
             return False
         raise ValueError(value)
     return bool(value)
+
+def to_surrogate_pair(char):
+    pairs = []
+    for code in char:
+        code_point = ord(code)
+        if code_point >= 0x10000:
+            high_part = (code_point - 0x10000) // 0x400 + 0xD800
+            low_part = (code_point - 0x10000) % 0x400 + 0xDC00
+            pairs += (high_part, low_part)
+        else:
+            pairs.append(code_point)
+    return pairs
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -24,6 +24,7 @@ from ctypes import windll, wintypes
 from plover.key_combo import parse_key_combo
 from plover.oslayer.winkeyboardlayout import KeyboardLayout
 from plover import log
+from plover.misc import to_surrogate_pair
 
 SendInput = windll.user32.SendInput
 LONG = ctypes.c_long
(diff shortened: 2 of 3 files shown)


morinted:
Would it be possible to write some unit tests against the splitting?

zwagoth:
I'll see what I can do in that regards, it may take me a few days. I don't have the right version of system python installed at the moment, I simply did this using the binary copy of Plover in order to get the emoji dictionary to work.

morinted:
What version of Windows did you run into issues with? I had tested my emoji dictionary on Windows with Plover 3+ and things were working, at the time at least.

zwagoth:
I am using windows 10 v1709 and the character that I was attempting to send is well above the 16bit range.

💦is listed as 0x1F4A6.

zwagoth:
So I've run into some issues while working on a test for this and am soliciting advice.

The code that needs to be tested is not platform specific, but it inside a platform specific module. There are no provisions for platform specific testing in the code, and the test would fail on other platforms if the module is imported. 

This means that the test either needs to be platform specific and lose most of its value as it will only get tested on a specific platform, or the code needs to be moved out into a non platform specific area, even though it will only ever be used on a specific platform, complicating all of the code for other platforms where it is not used at all.

morinted:
We have some utility functions stored and tested in https://github.com/openstenoproject/plover/blob/master/plover/misc.py

Maybe that would be a good place to include the splitting code, then just test that in isolation, and use it from the Windows side? Do you think that would work reasonably well?

plover/misc.py
```diff
@@ -65,3 +65,18 @@ def boolean(value):
             return False
         raise ValueError(value)
     return bool(value)
+
+
+def surrogate_pair(char):
```
morinted: Since this function isn't used apart from in to_surrogate_pair, could you inline it?

```python
code_point = ord(code)
if code_point >= 0x10000:
            high = (code_point - 0x10000) // 0x400 + 0xD800
            low = (code_point - 0x10000) % 0x400 + 0xDC00
            pairs += (high, low)
        else:
            pairs.append(code_point)
```

zwagoth:
Changes made.

zwagoth:
Rebased.

morinted:
Thanks so much!

PR #951, opened 2018-05-22
Dictionary consistency and redundancy changes/updates, E/ to AOE/

Greygal:
Changed most legacy multi-stroke definitions that began with E/ to being with AOE/, in order to be more consistent from a phonetic perspective, to reduce/eliminate conflicts with E/ as brief for "he", per discord conversation May 22, 2018 w/StenoKnight.
Removed redundant multi-stroke definitions that defined the word "he" followed by a second word, wherein the stroke(s) for the second word was identical to existing stroke(s) for the second word, as these are redundant entries.
Removed several duplicate entries.


plover/assets/main.json
```diff
@@ -28830,7 +28694,6 @@
 "HRAPS/-G": "lapsing",
 "HRAPS/-S": "lapses",
 "HRAPZ": "laps",
-"HRAR": "{^ular}",
```
morinted: I think this is not on purpose

Greygal: I've no idea how that happened.  I don't show that I deleted it, yet its gone.  Will add it back.

morinted:
Thanks for this great improvement

PR #962, opened 2018-06-03
gui_qt: Add tooltips to dictionary icons

### LINKED ISSUE #896: Create tooltips for dictionary status icons
## Summary

When using the mouse to hover over the dictionary status icons, there isn't an explanation about what each one is for:

image

(I'm personally aware of what each means by now, but I'm just writing this up from a usability perspective)

## Reproducing

Add multiple dictionaries. One with have the star icon.
Install a plugin like the plover-python-dictionary plugin and add a .py dictionary. It will have the lock icon.
Add a dictionary with an invalid format. It will have an error icon.
...

## Plover Version

## System

Windows 10

nsmarkop:
Fixes #896.

Added tooltips for the cases of

- loading
- favorite
- readonly

I left the error case as-is as it was showing a tooltip with the exception already. I wasn't sure on the exact messages to include in the tooltips, so I just took a shot at a short description of each state.

I ran `python setup.py extract_messages` to update the plover.pot file to include these new strings as well. It made some other changes outside of just line numbers for strings that have been added or no longer appear anywhere else; these appear fine at first glance to me.

### DIFF
--- a/plover/gui_qt/dictionaries_widget.py
+++ b/plover/gui_qt/dictionaries_widget.py
@@ -162,19 +162,24 @@ class DictionariesWidget(QWidget, Ui_DictionariesWidget):
             item = QTableWidgetItem(str(n + 1))
             if dictionary.path not in loaded_dictionaries:
                 icon = 'loading'
+                tooltip = _('This dictionary is being loaded.')
             else:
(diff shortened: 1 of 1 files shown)


nsmarkop:
Updated to remove the plover.pot changes.

PR #964, opened 2018-06-04
Add stitch mode

### LINKED ISSUE #178: Add support for "glue" with a specified separator

morinted:
`{MODE:STITCH}` to enable s-t-i-t-c-h-i-n-g, supporting custom delimiter with syntax `{MODE:STITCH:delim}`

This is a revisit of #623 and a mode to similarly support the functionality given by https://github.com/morinted/plover_stitching

This mode and that plugin together fix #178

### DIFF
--- a/plover/formatting.py
+++ b/plover/formatting.py
@@ -46,8 +46,10 @@ MODE_LOWER = 'LOWER'
 MODE_RESET = 'RESET'
 MODE_RESET_CASE = 'RESET_CASE'
 MODE_RESET_SPACE = 'RESET_SPACE'
+MODE_RESET_STITCH = 'RESET_STITCH'
 MODE_SET_SPACE = 'SET_SPACE:'
 MODE_SNAKE = 'SNAKE'
+MODE_STITCH = 'STITCH'
 MODE_TITLE = 'TITLE'
 
 META_ESCAPE = '\\'
@@ -498,9 +500,9 @@ class _Action:
                  # Previous.
                  prev_attach=False, prev_replace='',
                  # Current.
-                 glue=False, word=None, orthography=True, space_char=' ',
+                 glue=False, word=None, space_char=' ', stitch_char=None,
                  upper_carry=False, case=None, text=None, trailing_space='',
-                 combo=None, command=None,
+                 combo=None, command=None, orthography=True,
                  # Next.
                  next_attach=False, next_case=None
                 ):
@@ -522,11 +524,11 @@ class _Action:
 
         upper_carry -- True if we are uppercasing the current word.
 
-        othography -- True if orthography rules should be applies when adding
-                      a suffix to this action.
-
         space_char -- this character will replace spaces after all other
         formatting has been applied
+
+        stitch_char -- when set, this character is inserted between outputted
+        letters
 
         case -- an integer to determine which case to output after formatting
 
@@ -540,6 +542,9 @@ class _Action:
                  executed for this action.
 
         command -- The command that should be executed for this action.
+
+        othography -- True if orthography rules should apply when adding
+                      a suffix to this action.
 
         next_attach -- True if there should be no space between this and the next
                        action.
@@ -557,6 +562,7 @@ class _Action:
         self.next_case = next_case
         # Persistent state variables
         self.space_char = space_char
+        self.stitch_char = stitch_char
         self.case = case
         self.trailing_space = trailing_space
         # Instruction variables
@@ -571,9 +577,10 @@ class _Action:
             # Previous.
             prev_attach=self.next_attach,
             # Current.
-            case=self.case, glue=self.glue, orthography=self.orthography,
-            space_char=self.space_char, upper_carry=self.upper_carry,
-            word=self.word, trailing_space=self.trailing_space,
+            case=self.case, glue=self.glue,
+            space_char=self.space_char, stitch_char=self.stitch_char,
+            upper_carry=self.upper_carry, word=self.word,
+            trailing_space=self.trailing_space, orthography=self.orthography,
             # Next.
             next_attach=self.next_attach, next_case=self.next_case,
         )
@@ -583,8 +590,8 @@ class _Action:
             # Previous.
             prev_attach=self.next_attach,
             # Current.
-            space_char=self.space_char, case=self.case,
-            trailing_space=self.trailing_space,
+            space_char=self.space_char, stitch_char=self.stitch_char,
+            case=self.case, trailing_space=self.trailing_space,
             # Next.
         )
 
@@ -664,6 +671,7 @@ def _raw_to_actions(stroke, ctx):
                      case=ctx.last_action.case,
                      prev_attach=ctx.last_action.next_attach,
                      space_char=ctx.last_action.space_char,
+                     stitch_char=ctx.last_action.stitch_char,
                      trailing_space=ctx.last_action.space_char)
     ctx.translated(action)
     return [action]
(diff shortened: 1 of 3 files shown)


sp3ctum:
What is the relationship of the mode and the plugin? Is the plugin for adding some extra functionality? Or is it required?

morinted:
The plugin is for:

- Glue with operator meta
- Retroactive stitching meta

Modes can't be made with plugin, but a mode is like caps lock for stitching.

The mode and plugin are not dependent on each other.

plover/formatting.py
```diff
@@ -537,6 +539,9 @@ def __init__(self,
 
         command -- The command that should be executed for this action.
 
+        othography -- True if orthography rules should be applies when adding
```
benoit-pierre: `should be applies` -> `should apply`?

benoit-pierre:
With those 4 additional tests:
```python
    def test_mode_stitch_join_to_whitespace(self):
        r'''
        "PHO*D": "{MODE:STITCH}",
        "TEFT": "test",
        "S-P": "{^ ^}"

        PHO*D     ''
        TEFT      ' t-e-s-t'
        S-P       ' t-e-s-t '
        S-P/TEFT  ' t-e-s-t t-e-s-t'
        '''

    def test_mode_stitch_multiple_words_translation1(self):
        r'''
        "PHO*D": "{MODE:STITCH}",
        "TEFT": "test",
        "HR*U": "you will"

        PHO*D  ''
        TEFT   ' t-e-s-t'
        HR*U   ' t-e-s-t y-o-u w-i-l-l'
        '''

    def test_mode_stitch_multiple_words_translation2(self):
        r'''
        "PHO*D": "{MODE:STITCH}",
        "TEFT": "test",
        "HR*U": "you, will"

        PHO*D  ''
        TEFT   ' t-e-s-t'
        HR*U   ' t-e-s-t y-o-u, w-i-l-l'
        '''

    def test_mode_stitch_multiple_lines_translation(self):
        r'''
        "PHO*D": "{MODE:STITCH}",
        "TEFT": "test",
        "HR*U": "you\nwill"

        PHO*D  ''
        TEFT   ' t-e-s-t'
        HR*U   ' t-e-s-t\ny-o-u w-i-l-l'
        '''
```
Tests 1, 3 and 4 fail:
```python
E   AssertionError: test_mode_stitch_join_to_whitespace                                                  
E     
E     "PHO*D": "{MODE:STITCH}",
E     "TEFT": "test",
E     "S-P": "{^ ^}"
E     
E     PHO*D     ''
E     TEFT      ' t-e-s-t'
E   > S-P       ' t-e-s-t '
E     S-P/TEFT  ' t-e-s-t t-e-s-t'
E     
E      ' t-e-s-t- '
E   != ' t-e-s-t '

E   AssertionError: test_mode_stitch_multiple_words_translation2
E     
E     "PHO*D": "{MODE:STITCH}",
E     "TEFT": "test",
E     "HR*U": "you, will"
E     
E     PHO*D  ''
E     TEFT   ' t-e-s-t'
E   > HR*U   ' t-e-s-t y-o-u, w-i-l-l'
E     
E      ' t-e-s-t y-o-u-, w-i-l-l'
E   != ' t-e-s-t y-o-u, w-i-l-l'

E   AssertionError: test_mode_stitch_multiple_lines_translation
E     
E     "PHO*D": "{MODE:STITCH}",
E     "TEFT": "test",
E     "HR*U": "you\nwill"
E     
E     PHO*D  ''
E     TEFT   ' t-e-s-t'
E   > HR*U   ' t-e-s-t\ny-o-u w-i-l-l'
E     
E      ' t-e-s-t y-o-u-\n-w-i-l-l'
E   != ' t-e-s-t\ny-o-u w-i-l-l'
```

IMHO, the stitching code should use the same rules are other parts of the formatting code with respect to how words are delimited (and on that note, `_capitalize_all_words` is buggy`).

morinted:
Okay, I'll look at delimiting with the word regex later

morinted:
All right, I fumbled through this and got something that *works* but I'm not entirely satisfied with the cleanliness. Can you suggest improvements?

plover/formatting.py
```diff
@@ -741,7 +749,19 @@ def _atom_to_action(atom, ctx):
...
+        if (action.prev_attach
+                and action.stitch_char
+                # Don't add extra stitches if replacing
+                and not action.prev_replace
+                # Check that we are joining two words
+                and action.text
+                and ctx.last_action.word
+                and WORD_RX.match(action.text[0])
+                and WORD_RX.match(ctx.last_action.word[-1])):
+            action.text = action.stitch_char + action.text
```
benoit-pierre: This should be handled in `_apply_mode_stitch_char`.

plover/formatting.py
```diff
@@ -963,6 +990,21 @@ def _apply_mode_case(text, case, appended):
         return _capitalize_all_words(text)
     raise ValueError('invalid case mode: %s' % case)
 
+def _apply_mode_stitch_char(text, stitch_char):
```
benoit-pierre: I'd start with:
```python
if not stitch_char:
    return text
```
No need to do anything else if stitching is not enabled.

morinted:
I've amended with @benoit-pierre's action finalization code, as well as some additional black box tests to replace the mode tests.

I've also added a torture-torture test to ensure that orthography replace works even when the stitch mode has been changed in between the attach (which really should never happen but… 🤷‍♂️)

plover/formatting.py
```diff
@@ -776,6 +776,19 @@ def _apply_meta_attach(meta, ctx):
         new_word = add_suffix(last_word, meta)
         common_len = len(commonprefix([last_word, new_word]))
         replaced = last_word[common_len:]
+
+        if replaced.lower() != ctx.last_text(len(replaced)).lower():
```
benoit-pierre: I think the code should check if the previous word was stitched, rather than assume the previous word was stitched if `replaced.lower() != ctx.last_text(len(replaced)).lower()`.

plover/formatting.py
```diff
@@ -917,63 +931,103 @@ def _apply_meta_mode(meta, ctx):
...
+
+    text += _apply_stitching(action.text, action.stitch_char)
+    action.text = text
+
+def _apply_stitching(text, stitch_char):
+    words = WORD_RX.findall(text)
+    text = ''
+    for w in words:
+        sw = w.rstrip()
+        text += stitch_char.join(w.rstrip()) + w[len(sw):]
```
benoit-pierre:
```diff
- text += stitch_char.join(w.rstrip()) + w[len(sw):]
+ text += stitch_char.join(sw) + w[len(sw):]
```

plover/formatting.py
morinted: Is it possible to check if the previous word was stitched if the stitched action is several behind?

benoit-pierre: I can think of a lot of corner cases where getting the correct output would be problematic. 

Honestly, I don't think this feature is worth it: there's no practical use case, except for something like `S-P-E-L-L-I-N-G`, which can already be implemented with a  custom meta, so without the need for further complexity to the formatting code, and no expectations that it would work if followed by a suffix (which again make sense, I think).

morinted: So do you think the plugin is sufficient and this should be closed?

benoit-pierre: Yes.

PR #971, opened 2018-06-07
circleci: migrate configuration to 2.0

benoit-pierre:
Rational:
> Projects currently running on CircleCI 1.0 will no longer be supported after August 31st, 2018.

Note: I switched to a patched version of dmgbuild has I was getting a lot of build failures due to timeouts when `hdiutil detach` is called...


morinted:
Can you link the built artifact so I can test it on my machine?

benoit-pierre:
Can't you access the PR's build artifact?

morinted:
They might not be public by default

benoit-pierre:
The link can be accessed anonymously, but the artifacts tab is not shown on the build page if you're not logged-in: https://circleci.com/gh/benoit-pierre/plover/489#artifacts/containers/0

PR #981, opened 2018-07-08
gui_qt: Add open config folder action

### LINKED ISSUE #697: Open the config folder from the GUI
## Summary

As a user I want to be able to open the config folder from the GUI so that I don't need to search my file system for Plover's config file, log file, and plugins folder.

Considerations:
- If the log file has been saved to another folder, it might still be hard to find
- There's no clear place in the UI to put this.

nsmarkop:
Fixes #697.

I added an entry for "Open config folder" to the File menu that opens up the configuration folder.

image

The icon used is from [Icons8 Flat Color Icons](https://github.com/icons8/flat-color-icons).

Perhaps the oddest part here is the use of webbrowser.open to open the folder. os.startfile is Windows exclusive, so I searched around for cross platform ways to open folders and files and ended up on that standard library function accomplishing the desired behavior despite the name. Alternatively something like this could be implemented, but I figured if webbrowser.open works it's simpler to rely on it:

```python
import subprocess, os
if sys.platform.startswith('darwin'):
    subprocess.call(('open', filepath))
elif os.name == 'nt':
    os.startfile(filepath)
elif os.name == 'posix':
    subprocess.call(('xdg-open', filepath))
```

I've only verified behavior on Windows, so Mac / Linux testing would be nice.

### DIFF
--- a/plover/gui_qt/main_window.py
+++ b/plover/gui_qt/main_window.py
@@ -1,7 +1,9 @@
 
 from functools import partial
 import json
-import webbrowser
+import os
+import sys
+import subprocess
 
 from PyQt5.QtCore import QCoreApplication, Qt
 from PyQt5.QtGui import QCursor, QIcon, QKeySequence
@@ -242,7 +244,13 @@ class MainWindow(QMainWindow, Ui_MainWindow, WindowState):
         self._configure()
 
     def on_open_config_folder(self):
-        webbrowser.open(CONFIG_DIR)
+        # webbrowser.open does not work on Mac so we need OS-specific logic
+        if sys.platform.startswith('win'):
+            os.startfile(CONFIG_DIR)
+        elif sys.platform.startswith('linux'):
+            subprocess.run(['xdg-open', CONFIG_DIR])
+        elif sys.platform.startswith('darwin'):
+            subprocess.run(['open', CONFIG_DIR])
 
     def on_reconnect(self):
         self._engine.reset_machine()
--- a/plover/gui_qt/main_window.py
+++ b/plover/gui_qt/main_window.py
@@ -1,6 +1,7 @@
 
 from functools import partial
 import json
+import webbrowser
 
 from PyQt5.QtCore import QCoreApplication, Qt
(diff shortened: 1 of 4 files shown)


benoit-pierre:
@morinted: can you check it works on macOS?

morinted:
Yes -- will check in 36 hours when I have my Mac again.

morinted:
On initial go, this doesn't do anything on macOS. I'll try the PyQT update and see if that makes a difference as I'm on the current macOS version (10.14)

morinted:
After updating to the requirements in #1023, this still doesn't work :( might have to go with the OS-switch after all.

nsmarkop:
Darn. Since it doesn't seem to work for you on Mac, I updated the logic to be platform-specific instead. I only tested on Windows so I'll need verification again for Linux / Mac on this approach.

morinted:
Works for me on Mac

plover/gui_qt/main_window.py
```diff
@@ -238,6 +243,15 @@ def on_disable_output(self):
     def on_configure(self):
         self._configure()
 
+    def on_open_config_folder(self):
+        # webbrowser.open does not work on Mac so we need OS-specific logic
```
benoit-pierre: I think you can remove the mention, as testing it again on Linux with my configuration, I see that it always open a directory listing in my browser, which is clearly not the best behaviour.

plover/gui_qt/main_window.py
```diff
@@ -238,6 +243,15 @@ def on_disable_output(self):
...
+        if sys.platform.startswith('win'):
+            os.startfile(CONFIG_DIR)
+        elif sys.platform.startswith('linux'):
+            subprocess.run(['xdg-open', CONFIG_DIR])
```
benoit-pierre: For Python 3.4 compatibility:
```suggestion
            subprocess.call(['xdg-open', CONFIG_DIR])
```

benoit-pierre:
Works fine on Linux.

benoit-pierre:
Looks like the Circle CI build is not run on the result of the merge, but directly the PR branch, unlike AppVeyor / Travis...

PR #989, opened 2018-08-04
Add set config engine command

### LINKED ISSUE #983: Stroke command for opacity
## Summary

Currently the only way to toggle update window opacity between 0 and 100 is via the GUI. Having a {PLOVER:} command would allow this to be bound to a stroke and make it easier to switch on the fly. This is helpful for captioners who switch frequently between remote (where they'd prefer to see their dictionary update window) and onsite (where they want to hide it so their clients can't see it). If the opacity is left at 100 when starting an onsite job it can be obtrusive and unprofessional to change it in front of the client.

fourshade:
Fixes #983.
 
I've added an engine command: SET_CONFIG, which allows you to set a Plover config setting using a stroke. Mirabai asked for a command that would change the add-translation window opacity, and this does that among other things. Adding entries like this into your dictionary will allow you to set values in the global configuration:

``"O*EP": "{PLOVER:SET_CONFIG:translation_frame_opacity,100}",``
``"TR*PB": "{PLOVER:SET_CONFIG:translation_frame_opacity,0}"``

Those change the add translation window opacity, but it should work for any config setting with a value that is type-castable from a string. With a little further work I could make it usable for the more complex config settings, possibly allowing you to do things like switch out dictionaries using this command.

### DIFF
--- a/news.d/feature/989.core.rst
+++ b/news.d/feature/989.core.rst
@@ -1,0 +1,6 @@
+A new ``SET_CONFIG`` command can be used to change the configuration with a stroke, e.g.::
+
+        "O*EP": "{PLOVER:SET_CONFIG:'translation_frame_opacity':100}",
+        "TR*PB": "{PLOVER:SET_CONFIG:'translation_frame_opacity':0}",
+
+to change the opacity of the "Add Translation" dialog on the fly.
--- a/plover/command/set_config.py
+++ b/plover/command/set_config.py
@@ -1,0 +1,32 @@
+import ast
+
+
+def set_config(engine, cmdline):
+    """
+    Set one or more Plover config options upon executing a stroke pattern.
+
+    Syntax:
+    {PLOVER:SET_CONFIG:option:value}
+    {PLOVER:SET_CONFIG:option1:value1,option2:value2,...}
+
+    Example usage:
+    "O*EP":   "{PLOVER:SET_CONFIG:'translation_frame_opacity':100}",
+    "STA*RT": "{PLOVER:SET_CONFIG:'start_attached':True,'start_capitalized':True}",
+
+    Be careful with nested quotes. Plover's JSON dictionaries use double quotes
+    by default, so use single quotes for config option names and other strings.
+    """
+    # Each config setting can be processed as a key:value pair in a dict.
+    # The engine.config property setter will update all settings at once.
+    engine.config = _cmdline_to_dict(cmdline)
+
+
+def _cmdline_to_dict(cmdline):
+    """ Add braces and parse the entire command line as a Python dict literal. """
+    try:
+        opt_dict = ast.literal_eval('{'+cmdline+'}')
+        assert isinstance(opt_dict, dict)
+        return opt_dict
+    except (AssertionError, SyntaxError, ValueError) as e:
+        raise ValueError('Bad command string "%s" for PLOVER:SET_CONFIG.\n' % cmdline
+                         + 'See for reference:\n\n' + set_config.__doc__) from e
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,6 +41,7 @@ install_requires =
 	wcwidth
 packages =
 	plover
+	plover.command
 	plover.dictionary
 	plover.gui_none
 	plover.gui_qt
@@ -57,6 +58,8 @@ gui_qt =
 [options.entry_points]
 console_scripts =
 	plover = plover.main:main
+plover.command =
+	set_config = plover.command.set_config:set_config
 plover.dictionary =
 	json = plover.dictionary.json_dict:JsonDictionary
 	rtf  = plover.dictionary.rtfcre_dict:RtfDictionary
--- a/test/test_command.py
+++ b/test/test_command.py
@@ -1,7 +1,6 @@
 """ Unit tests for built-in engine command plugins. """
 
 import ast
-import io
 
 import pytest
 
@@ -9,9 +8,12 @@ from plover.command.set_config import set_config
 from plover.config import Config, DictionaryConfig
 from test.test_config import DEFAULTS, DEFAULT_KEYMAP
 
+from . import parametrize
 
-class _CFGFakeEngine:
-    """ The plugin sets the engine.config property, which unpacks a dict as keywords into config.update. """
+
+class FakeEngine:
+    """ The plugin sets the engine.config property, which unpacks a dict as
+    keywords into config.update. """
     def __init__(self):
         self._cfg = Config()
     @property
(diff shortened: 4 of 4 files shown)


plover/config.py
```diff
@@ -358,9 +362,13 @@ def _set(self, section, option, value):
...
-        opt = self._OPTIONS[name]
-        if opt.full_key is not None:
-            key = opt.full_key(self, key)
+        try:
+            opt = self._OPTIONS[name]
+            if opt.full_key is not None:
+                key = opt.full_key(self, key)
+        except KeyError:
+            log.error('invalid config option: %r', name, exc_info=True)
+            return key, null_option()
```
benoit-pierre: What are those changes (and the one above) for?

plover/engine.py
```diff
@@ -341,9 +341,18 @@ def _consume_engine_command(self, command):
...
+            if command_args[0] == 'SET_CONFIG':
+                key, value = tuple(command_args[1].split(','))
+                oldvalue = self[key]
+                try:
+                    self[key] = (type(oldvalue))(value)
+                except TypeError:
+                    log.error("Could not cast config value string to required type.")
+            else:
+                command_fn = registry.get_plugin('command', command_args[0]).obj
+                command_fn(self, command_args[1] if len(command_args) == 2 else '')
```
benoit-pierre: The new command should behave as a plugin, even if it is builtin one. So this whole hunk should disappear.

benoit-pierre:
I'm not against adding this to Plover, although I see no reason this can't be implemented in a separate plugin.

plover/config.py
fourshade: That code logs an error if the provided config option can't be found. Since the user is now able to attempt to set config values themselves through a command, I figured logging an error and returning a do-nothing option would be better than a KeyError that propagates out and crashes the program. The "null option" is just a dummy option with the required methods that does nothing at all; it allows the surrounding code to attempt to get and set operations without returning errors.

fourshade:
I actually did try implementing this as a plugin first, but there were a couple reasons I thought it would be better as a built-in feature.

1. The command itself is an incredibly simple addition. I ended up with a plugin that was six lines of code. About 90% of the plugin's total content was distribution metadata. Felt like kind of a waste.
2. Config values are central to the operation of Plover; they affect nearly every part of the system. The effects of config values are incredibly varied. Plugins are best for specific use-case scenarios; this command is really a much more general addition with all kinds of uses.
3. The current config code isn't very user-friendly. Right now, the _lookup of available config options by key in __getitem__ does not catch exceptions; without modifying the core program to catch the KeyError as mentioned above, the program would simply crash if there was a misspelled option key in the user's dictionary. A plugin couldn't have changed this behavior.

plover/engine.py
fourshade: I considered doing this like the built-in macros. I guess I was hesitant because there aren't any built-in PLOVER: commands that currently operate like that and doing so would entail making a new folder, file, and entry point. I'll make the change and push.

fourshade:
Okay, it's been converted to a built-in plugin in the same style as the built-in macro commands. engine.py is no longer changed.

morinted:
Just an aside, a short plugin is an asset in my mind.

fourshade:
I agree a short plugin can indeed be useful. I wish more of the simple command/meta plugins were built-in by default, since they take very little space and don't clutter the UI in any way.

Anyway, I've moved everything out of the main code into the plugin. I still feel like config.py could use a good cleanup (exception handling, refactoring all those ConfigOption factory functions into classes) but maybe that's best for another PR.

morinted:
Those could be made into default plugins which is also very convenient

plover/command/set_config.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Set a Plover config option to a given value upon executing a stroke pattern.
+
+    Example usage (enter these lines into dictionary):
+    "O*EP": "{PLOVER:SET_CONFIG:translation_frame_opacity,100}",
+    "TR*PB": "{PLOVER:SET_CONFIG:translation_frame_opacity,0}",
+    """
+    # Break down the command string into a key and a value. Throw an error if it isn't in the form of a 2-tuple.
+    try:
+        key, value = tuple(cmdline.split(','))
```
benoit-pierre: What if the value itself contains a comma? Might be better more consistent to split on `:` and only once:
```python
         key, value = tuple(cmdline.split(':', 1))
```

plover/command/set_config.py
```diff
@@ -0,0 +1,29 @@
...
+    except ValueError:
+        log.error("SET_CONFIG: Bad command string. Command must be in the form {PLOVER:SET_CONFIG:option,value}")
+        return
+    # We need to look up the old value of the config setting in order to determine its type.
+    try:
+        oldvalue = engine[key]
+    except KeyError:
+        log.error("SET_CONFIG: %s is not a valid Plover config option." % key)
```
benoit-pierre: It's better to not catch the exception: the engine already does that, and this will ensure the rest of the translation (if any) is not sent (as it might depend on the command being successful). You could catch and re-raise if you do want to ensure a more meaningful error message.

plover/command/set_config.py
```diff
@@ -0,0 +1,29 @@
...
+        return
+    # Type-cast the value string to match whatever type the old value was, and set the option to the new value.
+    try:
+        engine[key] = (type(oldvalue))(value)
+    except (TypeError, ValueError):
+        log.error("SET_CONFIG: Could not cast config value string \"%s\" to %s." % (value, type(oldvalue)))
+        return
```
benoit-pierre: Same here for the error handling.

benoit-pierre: I don't like the way de-serializing from a string is handled, I think a more robust approach would be to use `ast.literal_eval`:
```python
def set_config(engine, cmdline):
    """
    Set a Plover config option to a given value upon executing a stroke pattern.

    Example usage (enter these lines into dictionary):
    "O*EP": "{PLOVER:SET_CONFIG:translation_frame_opacity:100}",
    "TR*PB": "{PLOVER:SET_CONFIG:translation_frame_opacity:0}",
    """
    key, value = tuple(cmdline.split(':', 1))
    engine[key] = ast.literal_eval(value)
```
This would make it possible to use the command with complex settings, like machine options: `{PLOVER:SET_CONFIG:machine_specific_options:\{'arpeggiate': True\}}`.

benoit-pierre:
This really needs some tests.

plover/command/set_config.py
fourshade: Aha, that's perfect! I was looking for a way to handle the more complex config options! Thanks!

plover/command/set_config.py
fourshade: The value could just as easily contain a colon, but splitting just once does the trick no matter what the delimiter is (as long as the key can't contain it). Will do.

(I just thought a comma looked better in the dictionary. "SET_CONFIG:key,value" is easier to parse visually than "SET_CONFIG:key:value" for me. This one just comes down to preference, I guess.)

plover/command/set_config.py
fourshade: I just thought of something; should there be an extra level of exception messages for plugin commands? So that the user knows that the error is descended from the plugin and not the Plover engine itself?

Something like this:

```
def _consume_engine_command(self, command):
    ...
    command_args = command.split(':', 1)
    command_fn = registry.get_plugin('command', command_args[0]).obj
    try:
        command_fn(self, command_args[1] if len(command_args) == 2 else '')
    except Exception as e:
        raise Exception('Command error: %s' % command_args[0]) from e
```

I wouldn't have to manually add "SET_CONFIG" to each error message on the plugin side then.

fourshade:
Okay, this looks more promising. By parsing the value with the AST module, it should be possible to set any option you can set in a config file.

fourshade:
Commits are re-ordered and squashed, tests are fixed up to handle expected exceptions better.

plover/command/set_config.py
```diff
@@ -0,0 +1,41 @@
+import ast
+
+COMMAND_DELIMITER = ','
```
morinted: I think I'd still prefer `:` as a command delim. I agree that `,` is pretty readable but we've been doing params in most plugins with colon. For the sake of consistency and setting an example from within Plover's codebase

plover/command/set_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def plugin_error(e, msg):
+    """ Add an additional message to an exception before raising to the engine. """
+    e.args = (e.args[0] + "\nSET_CONFIG: " + msg,)
```
morinted: if you PEP8 this file I think it'll catch a few things. Can't remember if in python, single-quotes are preferred

plover/command/set_config.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Example usage (enter these lines into dictionary):
+    "O*EP": "{PLOVER:SET_CONFIG:translation_frame_opacity,100}",
+    "TR*PB": "{PLOVER:SET_CONFIG:translation_frame_opacity,0}",
+    """
+    # Break down the command string into a key and a value. Throw an error if it isn't in the form of a 2-tuple.
+    try:
+        key_str, value_str = tuple(cmdline.split(COMMAND_DELIMITER, 1))
+    except ValueError as e:
+        plugin_error(e,"Bad command string. Command must be in the form {PLOVER:SET_CONFIG:option,value}")
```
morinted: spacing between args is also in pep8 I think

morinted: I feel like it's unclear that `plugin_error` raises an exception. can we rename to something like `raise_command_error`?

plover/command/set_config.py
```diff
@@ -0,0 +1,41 @@
...
+    """ Add an additional message to an exception before raising to the engine. """
+    e.args = (e.args[0] + "\nSET_CONFIG: " + msg,)
+    raise e
+
+
+def set_config(engine, cmdline):
+    """
+    Set a Plover config option to a given value upon executing a stroke pattern.
+
+    Example usage (enter these lines into dictionary):
```
morinted: > (enter these lines into dictionary)

I don't think this aside is necessary

plover/command/set_config.py
morinted: (could also use COMMAND_DELIMITER constant in the error here)

plover/command/set_config.py
```diff
@@ -0,0 +1,41 @@
...
+        value_obj = ast.literal_eval(value_str)
+    except SyntaxError as e:
+        plugin_error(e, "Error parsing value string \"%s\"" % value_str)
+    # If it is (or can be coerced to) the type of the old value, set the option to the new value.
+    try:
+        engine[key_str] = value_obj
+    except (NameError, TypeError) as e:
+        plugin_error(e,"Config value type is incompatible: \"%s\" != %s." % (value_str, type(old_value)))
+    except ValueError as e:
+        plugin_error(e,"Value \"%s\" is not allowed for config option %s." % (value_str, key_str))
```
morinted: I found the control-flow with so many try-excepts hard to reason about, but overall I think it's okay if you don't wanna make any changes. Just for fun, I tried splitting out each except into its own function and then replacing some comments with function names:

```python
def get_config_key_value_strings():
    try:
        config_key, value_str = tuple(cmdline.split(COMMAND_DELIMITER, 1))
        return (config_key, value_str)
    except ValueError as e:
        raise_command_error(e, 'Bad command string. Command must be in the form {PLOVER:SET_CONFIG:option,value}')

def get_current_config_value(config_key):
    try:
        return engine[config_key]
    except KeyError as e:
        raise_command_error(e, '%s is not a valid Plover config option.' % config_key)
def convert_config_value_string_to_literal(value_str):
    try:
        return ast.literal_eval(value_str)
    except SyntaxError as e:
        raise_command_error(e, 'Error parsing value string "%s"' % value_str)

def set_config(config_key, config_value, current_value)
    # If it is (or can be coerced to) the type of the old value, set the option to the new value.
    try:
        engine[config_key] = config_value
    except (NameError, TypeError) as e:
        raise_command_error(e, 'Config value type is incompatible: "%s" != %s.' % (config_value, type(current_value)))
    except ValueError as e:
        raise_command_error(e, 'Value "%s" is not allowed for config option %s.' % (config_value, config_key))

config_key, value_str = get_config_key_value_strings()
current_config_value = get_current_config_value(config_key)
new_config_value = convert_config_value_string_to_literal(value_str)
set_config(config_key, new_config_value, current_config_value)
```

(note: didn't run this at all 😛 )

plover/command/set_config.py
fourshade: What about a colon to separate option:value, and commas to separate entire config settings? That allows for the syntax:

```{PLOVER:SET_CONFIG:option:value}```
```{PLOVER:SET_CONFIG:option1:value1,option2:value2:...}```

Which makes it easy to grab the command-line part, slap on a couple braces, and parse the whole thing as a dict literal. You could set any number of config options with one stroke that way. That feels too good to pass up...

**Comment by fourshade** [comment]
Okay, it's fixed up a bunch now. Split everything into methods and had them raise a custom exception class. Making the syntax the same as Python dicts means there's no need to keep track of special delimiter constants. Passes all PEP 8 checks (except for possibly line length, but 80 characters feels like the dark ages with tons of continuations on long strings and comments. PyCharm defaults to 120.)

**Comment by fourshade** [comment]
That's a good idea; doing the update in one shot also simplifies the plugin a great deal. All command parsing exceptions are folded into one try-except now and the actual config update is left to propagate exceptions as normal.

File: plover/command/set_config.py [hunk 11]
```diff
@@ -0,0 +1,32 @@
...
+
+def _cmdline_to_dict(cmdline):
+    """ Add braces and parse the entire command line as a Python dict literal. """
+    try:
+        opt_dict = ast.literal_eval('{'+cmdline+'}')
+        assert isinstance(opt_dict, dict)
+        return opt_dict
+    except (AssertionError, SyntaxError, ValueError) as e:
+        raise ValueError(f'Bad command string "{cmdline}" for PLOVER:SET_CONFIG.\n'
+                         f'See for reference:\n\n{set_config.__doc__}') from e
```
benoit-pierre: Please remove the use of f-strings: we support Python `>=3.4` and those are only available starting with 3.6.

benoit-pierre: We just removed support for 3.4, but we still want support for 3.5. And that does not preclude us using a more recent version of Python for the distributions (which we do). Anyway, thanks for your patience with this, I've updated the branch with a couple of tweaks (parametrized the tests + news entry). I'll go ahead and merge once the CIs have done their job.

news.d/feature/989.core.rst
```diff
@@ -0,0 +1,6 @@
+A new ``SET_CONFIG`` command can be used to change the configuration with a stroke, e.g.::
+
+        "O*EP": "{PLOVER:SET_CONFIG:translation_frame_opacity,100}",
+        "TR*PB": "{PLOVER:SET_CONFIG:translation_frame_opacity,0}",
```
morinted, line 4: This needs to be updated to show the correct format

```suggestion
        "O*EP": "{PLOVER:SET_CONFIG:'translation_frame_opacity':100}",
        "TR*PB": "{PLOVER:SET_CONFIG:'translation_frame_opacity':0}",
```

benoit-pierre: Good catch, amended.

PR #998, opened 2018-08-11
Replace case-reverse lookup system

fourshade:
I've been breaking down my changes to the search system into pieces, and this one is probably the most drastic (it removes quite a bit of code), so I feel like it should be discussed first. Basically, when finding word suggestions, the system uses a special case-reverse dictionary to find case variants of a given word. This is a dict of Counter dicts which involves extra maintenance work from the steno dictionary class and takes up quite a bit of memory. When I looked at the Plover dictionary code for the first time, the case-reverse system confused me a bit and felt like kind of an ugly hack.

Currently, in order to find special entries such as prefixes and suffixes, the suggestions finder simply creates a list of possibilities to try and looks up each one in turn. Hash map lookups are cheap, so I thought "why not do this for case variants as well?". After all, for normal English text, case variants consist almost exclusively of three choices: lowercase, Capitalized, or UPPERCASE. Adding these to the possibilities list simplifies the lookup system a great deal and offloads a bunch of work from the dictionary class.

The bulk of the changes to the code are:

1. Changing the Suggestion finder to manually check for the most common case variants. The code is structured a little differently; hopefully I added enough comments. Basically, I took a list of base variants (regular, lowercase, uppercase, etc.) on the given search string, removed duplicates, then ran each one through the list of "mod" functions used to check for things like prefixes and suffixes. All of the possibilities are then looked up and the ones that exist are returned. I didn't really need to convert the mods list into lambdas since all of the current ones are just format strings (it happened while testing the approach of making case variants into mods themselves), but I feel like it makes the intentions clearer, and it allows variants other than simple string substitution in the future.

2. Removing all case-reverse code. The suggestion finder was the only module that used this, so it can be safely removed. This means less upkeep for the dictionaries, less test code, and one less path through the engine. On my system, this modification alone caused a 15% decrease in Plover's memory usage.

The two main downsides I can see to this approach are:
- it won't detect EXOtic CAse VARiants, or multi-word case variants such as Title Case and CamelCase.
- if there are any plugins that currently use engine.casereverse_lookup, they will break.

For the first point, an exact match will still work for looking up complicated case variants, and a future regex search (probably my next PR, I've cleaned up my existing code a bit) could do a full case-insensitive search. For the second, there could be some sort backwards compatibility code there instead, but at the moment I can't think of any current plugin that might need to use this engine function directly.

### DIFF
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -15,7 +15,7 @@ from plover.registry import registry
 from plover.resource import ASSET_SCHEME, resource_filename
 from plover.steno import Stroke
 from plover.steno_dictionary import StenoDictionary, StenoDictionaryCollection
-from plover.suggestions import Suggestions
+from plover.suggestions import find_suggestions
 from plover.translation import Translator
 
 
@@ -468,7 +468,7 @@ class StenoEngine:
 
     @with_lock
     def get_suggestions(self, translation):
-        return Suggestions(self._dictionaries).find(translation)
+        return find_suggestions(self._dictionaries,translation)
 
     @property
     @with_lock
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -459,11 +459,6 @@ class StenoEngine:
         return [] if matches is None else matches
 
     @with_lock
-    def casereverse_lookup(self, translation):
-        matches = self._dictionaries.casereverse_lookup(translation)
-        return set() if matches is None else matches
-
-    @with_lock
     def add_dictionary_filter(self, dictionary_filter):
         self._dictionaries.add_filter(dictionary_filter)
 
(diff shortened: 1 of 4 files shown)


fourshade:
If one of the build tests fails, is there a way to manually try it again? It looks like one of them only failed because it couldn't update its own components from the internet.

benoit-pierre:
Closing and reopening the PR will re-trigger the builds.

fourshade:
Come to think of it, Suggestions doesn't need to be a class. It doesn't need to store any sort of state; right now the engine is just creating a new one each time, calling its only method, then destroying it. Refactored into a function.

fourshade:
Whoops, putting the base translations into a set to remove duplicates screwed up the order. We want the exact translation (if it exists) to appear first. Since there's no standard OrderedSet, some people have used keys of an OrderedDict this way.

The list comprehension profiled faster than the generator too...odd, but it's an easy change.

fourshade:
My new search implementation is superior to this one in almost every way except memory usage, so I'll go ahead and close this PR.

PR #999, opened 2018-08-13
Fix tooltip punctuation consistency (issue #963) and make tool tooltips localizable

### LINKED ISSUE #963: Tooltip format is inconsistent
## Summary

Tooltips in Plover don't seem to agree on if they should end with a period or not.

For example, the buttons to move dictionaries up and down end with periods:

image

But the ones for editing / deleting / adding dictionaries don't:

image

Most of the places I've checked end with periods.

Tool plugins are a place outside of core Plover code that also have inconsistencies since they have tooltips constructed from the class documentation which may or may not have a period depending on the developer. I'm not sure if it's a good thing to try and check then force one for those specific use cases, though.

## Reproducing

Open Plover and mouse over some stuff like the buttons mentioned in the problem summary.

## Plover Version

Plover 4.0.0.dev8

## System

Windows 10

SeaLiteral:
Fix #963 by removing periods at the end of some tooltips, supporting localizable tooltips for the Tool class (which used to only fetch tooltips from class docstrings) and adding such tooltips to the built in tools.

### DIFF
--- a/plover/gui_qt/add_translation_dialog.py
+++ b/plover/gui_qt/add_translation_dialog.py
@@ -8,9 +8,9 @@ _ = get_gettext()
 
 class AddTranslationDialog(Tool, Ui_AddTranslationDialog):
 
-    ''' Add a new translation to the dictionary. '''
 
     TITLE = _('Add Translation')
+    TOOLTIP = _('Add a new translation to the dictionary.')
     ICON = ':/translation_add.svg'
     ROLE = 'add_translation'
     SHORTCUT = 'Ctrl+N'
--- a/plover/gui_qt/config_window.py
+++ b/plover/gui_qt/config_window.py
@@ -285,14 +285,14 @@ class ConfigWindow(QDialog, Ui_ConfigWindow, WindowState):
                 ConfigOption(_('Add translation dialog opacity:'), 'translation_frame_opacity',
                              partial(IntOption, maximum=100, minimum=0),
                              _('Set the translation dialog opacity:\n'
-                               '- 0 makes the dialog invisible\n'
-                               '- 100 is fully opaque')),
+                               '- 0 makes the dialog invisible.\n'
+                               '- 100 is fully opaque.')),
                 ConfigOption(_('Dictionaries display order:'), 'classic_dictionaries_display_order',
                              partial(BooleanAsDualChoiceOption,
                                      _('top-down'), _('bottom-up')),
                              _('Set the display order for dictionaries:\n'
-                               '- top-down: match the search order; highest priority first\n'
-                               '- bottom-up: reverse search order; lowest priority first\n')),
+                               '- top-down: Match the search order; highest priority first.\n'
+                               '- bottom-up: Reverse search order; lowest priority first.\n')),
             )),
             (_('Logging'), (
                 ConfigOption(_('Log file:'), 'log_file_name',
--- a/plover/gui_qt/dictionaries_widget.ui
+++ b/plover/gui_qt/dictionaries_widget.ui
@@ -81,7 +81,7 @@
     <string>&amp;Edit dictionaries</string>
    </property>
    <property name="toolTip">
-    <string>Edit selected dictionaries</string>
+    <string>Edit selected dictionaries.</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+E</string>
@@ -96,7 +96,7 @@
     <string>&amp;Remove dictionaries</string>
    </property>
    <property name="toolTip">
-    <string>Remove selected dictionaries</string>
+    <string>Remove selected dictionaries.</string>
    </property>
    <property name="shortcut">
     <string>Del</string>
@@ -126,7 +126,7 @@
     <string>&amp;Add dictionaries</string>
    </property>
    <property name="toolTip">
-    <string>Add dictionaries</string>
+    <string>Add dictionaries.</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+O</string>
@@ -141,7 +141,7 @@
     <string>Add &amp;translation</string>
    </property>
    <property name="toolTip">
-    <string>Add a new translation</string>
+    <string>Add a new translation.</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+N</string>
--- a/plover/gui_qt/dictionary_editor.ui
+++ b/plover/gui_qt/dictionary_editor.ui
@@ -155,7 +155,7 @@
     <string>&amp;New translation</string>
    </property>
    <property name="toolTip">
-    <string>Add a new translation</string>
+    <string>Add a new translation.</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+N</string>
--- a/plover/gui_qt/lookup_dialog.py
+++ b/plover/gui_qt/lookup_dialog.py
@@ -13,9 +13,9 @@ _ = get_gettext()
 
 class LookupDialog(Tool, Ui_LookupDialog):
 
-    ''' Search the dictionary for translations. '''
 
     TITLE = _('Lookup')
+    TOOLTIP = _('Search the dictionary for translations.')
     ICON = ':/lookup.svg'
     ROLE = 'lookup'
     SHORTCUT = 'Ctrl+L'
--- a/plover/gui_qt/main_window.py
+++ b/plover/gui_qt/main_window.py
@@ -103,8 +103,8 @@ class MainWindow(QMainWindow, Ui_MainWindow, WindowState):
             for parent in (tools_menu, self.toolbar, self.toolbar_menu):
                 action = parent.addAction(*action_parameters)
                 action.setObjectName(tool_plugin.name)
-                if tool.__doc__ is not None:
-                    action.setToolTip(tool.__doc__)
+                if tool.TOOLTIP is not None:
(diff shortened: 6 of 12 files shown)


nsmarkop:
Should the direction here be making everything have periods, or removing them? At least Microsoft's UX guide has tooltips (well, "infotips") using periods: https://docs.microsoft.com/en-us/windows/desktop/uxguide/ctrl-tooltips-and-infotips

SeaLiteral:
A lot of programmes (e.g. LibreOffice.org Writer) have one-word-and-one-shortcut tootips like `New (Ctrl+N)` or `Bold (Ctrl+B)`. But those are pretty short and some of those in Plover are longer. GIMP has longer tooltips like "Select by Cilir Tool: Select regions with similar colors **Shift+O**", which still don't have periods. I mostly use Linux, but I think Plover should try to be consistent with other open source applications rather than trying to do whatever CaseCat is doing. And that page you link to doesn't say that tooltips should or shouldn't have periods. It just happens to have one on the example in the screenshot. That being said, the multi-line tooltip in machine_options.py is pretty long and contains more than one sentence, so that one should probably keep the period.

```
self.arpeggiate.setToolTip(_(
            'Arpeggiate allows using non-NKRO keyboards.\n'
            '\n'
            'Each key can be pressed separately and the\n'
            'space bar is pressed to send the stroke'
        ))self.arpeggiate.setToolTip(_(
            'Arpeggiate allows using non-NKRO keyboards.\n'
            '\n'
            'Each key can be pressed separately and the\n'
            'space bar is pressed to send the stroke.'
        ))
```

nsmarkop:
I'm not sure why you randomly brought up CaseCat. You can find many examples of programs, open source or not, using either of the styles (or in the case of what I linked to, both -- one for "tooltips" and one for "infotips"). My intent was just to provide one example doing something different than proposed here.

This is likely something that would benefit from Ted or Benoit making a decision for what they want so a standard can be enforced for future development.

SeaLiteral:
On the page you linked, there are two screenshots: an "infotip" with a period and a tooltip without one. And yeah, mentioning another steno programme was random. And at least it should be possible to find the strings, at the very least by running Plover and pointing at everything in every window Plover can open and using grep to find them (or looking for them in the pot file, but the pot file doesn't mention which strings are tooltips, and it might have to be regenerated). That being said, I personally prefer letting it depend on their length, so the multi-sentence ones should have punctuation and single-word tooltips should, in my opinion, not. As for the medium-length ones, I found it easier to remove them than to add them, but let's see what the others say. And it seems one of the multi-line ones has periods and one doesn't, I guess I should fix that.

SeaLiteral:
Such a guideline would also be useful for plugin developers.

benoit-pierre:
I'm for make the tooltips proper phrase, meaning they should end with a `.`. Especially since the goal is to have long helpful tooltip that explain what a button/feature does rather than just a substitute for the lack of label in some cases.

plover/gui_qt/add_translation_dialog.py
```diff
@@ -11,6 +11,7 @@ class AddTranslationDialog(Tool, Ui_AddTranslationDialog):
     ''' Add a new translation to the dictionary. '''
 
     TITLE = _('Add Translation')
+    TOOLTIP = _('Add a new translation to the dictionary')
```
benoit-pierre: How about this instead:
```python
    __doc__ = _('Add a new translation to the dictionary.')
```
?

SeaLiteral:
Docstrings are supposed to be read-only in Python, so setting them to the return value of a function won't work. We could do `exec("class Test:\n    '''"+_('Tooltip text)+"'''")` and it would work, but, really? There are some more elegant ways to do this, but they do feel like they were meant to be known to the average programmer, so I think it might be better to use the TOOLTIP if it's set and the docstring if no TOOLTIP is specified, for backwards compatibility with Plugins written before this.

Speaking of Plugins, we might want to have a better tooltip than "Plugins Manager". Why even have a tooltip when the name of the tool already appears under the icon? Also, are plugins meant to be localizable?

benoit-pierre:
GUI plugins should be localizable using `plover.gui_qt.i18n.get_gettext(...)`, yes.

SeaLiteral:
How about strings that aren't tooltips? Some of them look like sentences. For instance there's the {stroke} maps to {translation}, which has the additional problem that it might look like the period is a part of the translation.
hadbeen
**Note**: There are several files I haven't added the periods to yet. I'm just writing this question before I finish editing those. I'll write back when there are no files missing periods as far as I know.

**Note 2**: I'm showing a "{stroke} maps to {translation}" screenshot that I originally took to show something else, it's not meant to show a smiley here, but this was faster to find than take another screenshot. I hope that's okay.

SeaLiteral:
**Answer to question**: Docstrings [used to be read-only](https://bugs.python.org/issue12773).

And now, back to the **main topic**: Using plover.pot to spot tooltips I think I got all of them now. You might want to look at this one from gui_qt/config_window.py. I edited the capitalization to fit:

```
Set the display order of dictionaries:
- top-down: Match the search order; highest priority first.
- bottom-up: Reverse search order; lowest priority first.
```
I'm editing a screenshot that contained it so that it doesn't nearly get as wide as the screen.
Screenshot with the tooltip moved to be below the option rather than outside the window
The tooltip matches capitalization of selectable options, but might not fit the punctuation inside the tooltip.

It's worth noting that right above that option is an opacity option that would have capitalization issues if it weren't because numbers are neatly exempt from that. If we change the punctuation in the former, we might need to do so in the latter for consistency.
```
Set the translation dialog opacity:
- 0 makes the dialog invisible.
- 100 is fully opaque.
```

**Tooltips**: Should be okay, but should the ones mentioning option names match the capitalization of the option names?
**Punctuation in other strings**: How do we deal with `{stroke} maps to {translation}` (we don't want the period to look like it's a part of the translation, which the background colour partially solves, but visually impaired users might be unable to see that) and `{translation} maps to stroke {strokes}` (where we don't need to worry about the period being mistaken for a key, but we still want those two messages to be consistent with each other).

SeaLiteral:
Also, I'm going to remove the code that gets the docstrings if there isn't a tooltip.

It might have been just in earlier versions of Python that docstring changing was hard to do, but normal docstrings are readable by things like IDEs whereas the ones set programatically are not. And then we might as well use Tool.TOOLTIP instead.

SeaLiteral:
And that's done now, including changing a comment. We might want to make changes to some plugins as well, but that's outside Plover.

SeaLiteral:
I think I've made the changes. I might post issues and pull requests to plugins where I notice they need to be changed after this (since their tooltips in the main window won't show once this PR gets merged, but other than that, the plugins should still work). But I think plugins shouldn't change their tooltips until Plover does.

morinted:
This looks good to me -- we could do a squash-and-merge -- anything else that needs updating?

plover/gui_qt/add_translation_dialog.py
SeaLiteral: Docstrings used to be read-only, but are changeable now. But their goal is to work as a type of comments that can be fetched automatically when you're trying to call a function but want to get reminded what the function does or how it's meant to be used. By encoding them in an unusually way, we're stopping IDEs from showing them, but fetching the docstring is just fetching the __doc__ attribute, so creating an attribute called TOOLTIP means anyone looking at the code can guess what that tooltip is rather than wondering why we're setting the docstring that way.

SeaLiteral:
I'm closing and reopening the PR to fix the problem that seems to have been caused by a new pip release.

SeaLiteral:
That doesn't seem to work. Also, Pierre wrote something while I was doing this. Closing again.

SeaLiteral:
So there's two main changes here:

- Make tooltip punctuation consistent between files
- Support localizable tooltips

I think 4066b60edcb8eb5a1dc94f1120eac9c36eff97a4 ("Make tooltip into sentence (and add period)") falls into the first category even if it's changing the wording in order to change the punctuation. And those things I changed and then changed back, I should just merge the change with the change back?

And then I suppose I should end up with two commits: one for punctuation changes and one for localizability.

SeaLiteral:
This is the list of commits as it shows now. I'm showing which of the two changes (punctuation and code) they relate to:

- Make tooltip punctuation consistent between files (punctuation) 65a5206
- Change where tooltips go to allow localization (TOOLTIP variable) 96615d4
- Add periods to tooltips. And make a tooltip into a sentence (both) e22cd62
- Add period to tooltip and remove docstring (both) 9cb45aa
- Add periods to tooltips (punctuation) ef46898
- Make tool tooltips localizable (TOOLTIP variable) 9da99aa
- re-add period (TOOLTIP variable!) 1222ca7

The last one re-adds a period that I'd removed while adding the TOOLTIP member variable. Hence why I group it with TOOLTIP variable list even though it's a punctuation change.

There's also the e22cd62 one that paraphrases a tooltip to make it a proper sentence with a full stop at the end in `paper_tape.py`.

Now, even when I gave a string a name and changed the punctuation in that string at the same time, it's pretty easy to see both changes without one making the other hard to understand. So I think we could merge the whole thing into a single commit.

SeaLiteral:
Since the tooltip text editing and the localizability fix don't make each other harder to understand in the diffs, I've merged it all into one commit. I also looked at the files to check I hadn't somehow lost any of the changes while merging stuff.

Issue #917, opened 2018-03-09
Dictionary Lookup not listing items  (  KR*GS but not KW*GS )


infodungeon:
## Summary

Example;  "  shows KR*GS but not KW*GS even though both are listed in the dictionary

## Reproducing

Look up "  only one entry presented.
Not sure if this is a general or specific issue.

## Plover Version

Plover 4.0.0.dev7+10.gc4d2892 

## System

Window 10

benoit-pierre:
Just to be clear:

* `"KR*GS": "{^\"}"` is shown
* but `"KR-GS": "{^~|\"}"` or `"KW-GS": "{~|\"^}"` are not

right?

seebs:
Yeah, I see only KR*GS, and I know that KW-GS and KR-GS also exist, because those are the ones I actually use.

zwagoth:
I am seeing this with entries like `really` which is in the dictionary as `R-L` which is getting assigned to `recall` which has `-RL`. I can see this in the 3.x branch of Plover as well, so it looks like a bigger issue with loading and normalizing strokes. Some entries are just outright deleted, such as `R-F` which maps to `rf` but it can't be found in the lookup tools by stroke or by translation.

I'll have a look and see if I can come up with anything more specific.

Issue #950, opened 2018-05-18
Should lookup be case-insensitive?


mulka:
## Summary

When you type 'return' (all lowercase) into the lookup tool, I feel like it should come up with {#Return}, but it doesn't.

## Reproducing

Open the lookup tool. Type 'return' in all lowercase. Notice how {#Return} doesn't show up. Now, type 'Return' with the capital letter. Notice how {#Return} does show up.

## Plover Version

Plover 4.0.0.dev8+8.ge062cda 

## System

Mac OSX El Capitan 10.11.6

DanLanglois:
a note:

plover/plover/suggestions.py

45        strokes_list = self.dictionary.reverse_lookup(modded_translation)

perhaps would work like this:

[ 45      strokes_list = self.dictionary.casereverse_lookup(modded_translation.lower()) ]

There is a plover/test/test_steno_dictionary.py here, but maybe doesn't need to be touched.. I need to test somehow but have to come back..

Issue #943, opened 2018-04-26
Make lookup window pinable pretty please :D


Greygal:
## Summary

Feature request: Please make it possible to "pin" the lookup window, thanks!

morinted:
Because using {PLOVER:LOOKUP} focuses the window, what's the utility in having it pinned?

Greygal:
Because when I then move my cursor or focus to another window/screen, I can't see what the lookup results were (short term memory fail, lol!)

Greygal:
It would also be grand if you could lookup commands in the lookup window.  For example, I wanted to know how to do a TAB key, but when I lookup TAB, I literally get, TAB, as in the word "tab," and not the command key for TAB (which is TA*B, I know now).  

Having the lookup tool display non-word entries would be very useful!

Greygal:
Would also be nice if the lookup window was less exacting and strict, and have it start returning partial matches as you are writing, instead of waiting for an exact match.

For example, typing "whe" and start seeing all possible words that start with "whe"
when
whet
whether
where
etcetera.  This not only makes the lookup more responsive and faster for the user, it also means the user doesn't have to write out the entire word to find what they are looking for (saving strokes, especially if you are fingerspelling!)

It also reduces possibility of not finding  the word you are looking for, because you spelled it wrong in the lookup.

PR #1001, opened 2018-08-18
Lookup/suggestions implementation overhaul

fourshade:
Fixes: #917, #950, and a big step toward #943

This is it: I've finally overhauled the inner workings of the reverse dict and search system to work much better. The main improvements are:

- Standard lookup/suggestions search is now fully case-insensitive and finds more entries with special characters than it did before.
- Search is about twice as fast on average.
- Plover uses 15 MB less memory overall.
- Initial dictionary loading times are up to 30% faster.
- No plugins or other external features should break; the only API change so far is that multi-dict reverse lookup returns sets instead of lists (which it probably should have anyway since it was never strictly ordered in the first place)
- Partial word matching (like autocomplete, i.e. "mas" returns "mash" "mask", and "mass") and search using regular expressions are now available using checkboxes in the lookup window.

Anyone wants to try it out, feel free. It passes all the tests at every commit stage, and current benchmarks are here: [search_benchmark.txt](https://github.com/openstenoproject/plover/files/2408448/search_benchmark.txt)

8/20: And with the latest change, #917 is fixed. The search feature will find more things that have meta commands.

8/28: Everything except possible config options (for controlling things like the maximum number of returned results) is done. I realize this is a rather large change from somebody who hasn't been a part of Github or in fact any collaborative editing platform for very long, but people have been asking for these types of search features for a while and I do believe this is a very worthwhile addition to Plover. I tried to stay away from anything related to the actual input->translation->output part of the engine, so there shouldn't be much chance of breaking anything crucial to Plover's core operation.

9/23: Config option is in. Also ran new benchmarks that should be more reliable (i.e. not as dependent on how lucky you get with the test data).

### DIFF
--- a/plover/config.py
+++ b/plover/config.py
@@ -112,12 +112,14 @@ def json_option(name, default, section, option, validate):
 def int_option(name, default, minimum, maximum, section, option=None):
     option = option or name
     def getter(config, key):
-        return config._config[section].getint(option)
+        return config._config[section][option]
     def setter(config, key, value):
         config._set(section, option, str(value))
     def validate(config, key, value):
-        if not isinstance(value, int):
-            raise InvalidConfigOption(value, default)
+        try:
+            value = int(value)
+        except ValueError as e:
+            raise InvalidConfigOption(value, default) from e
         if (minimum is not None and value < minimum) or \
            (maximum is not None and value > maximum):
             message = '%s not in [%s, %s]' % (value, minimum or '-∞', maximum or '∞')
--- a/plover/config.py
+++ b/plover/config.py
@@ -32,6 +32,11 @@ OUTPUT_CONFIG_SECTION = 'Output Configuration'
 DEFAULT_UNDO_LEVELS = 100
 MINIMUM_UNDO_LEVELS = 1
 
+DEFAULT_SEARCH_WORD_LIMIT = 5
+MINIMUM_SEARCH_WORD_LIMIT = 1
+DEFAULT_SEARCH_STROKE_LIMIT = 10
+MINIMUM_SEARCH_STROKE_LIMIT = 1
+
 DEFAULT_SYSTEM_NAME = 'English Stenotype'
 
 SYSTEM_CONFIG_SECTION = 'System: %s'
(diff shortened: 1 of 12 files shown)


benoit-pierre:
But you did change the API, the most important part: not the name of the function, or the arguments, but the behaviour! If `casereverse_lookup` is not doing a case insensitive reverse lookup anymore, then it's name should be changed (and the implementation should return a set if order is not guaranteed).

fourshade:
Strictly case-similar matches are a subset of those returned by this search, so I kept the name unchanged on the off-chance a plugin used it. But I see your point; I guess it would be simple enough to move the search to a new method and have casereverse() just call that and toss any extra entries for strict backward compatibility.

fourshade:
All done. We may not really need those methods unless current plugins use them.

Making multi-dict reverse lookup return sets is technically an API change too, I guess, but one that makes sense. The results can't contain duplicates, and the order doesn't matter unless we're visually displaying them (at which point they get sorted). Reversing the order of the dictionaries makes the tests fail if it returns lists, even if none of the keys overlap.

fourshade:
Holy crap, forget 10% faster, I'm getting numbers that have the search running almost THREE TIMES as fast now for random queries taken from the dictionary (i.e. real words; it's even faster for random garbage data, but that's not horribly relevant). What's more, it should work almost as fast when doing partial word searches, since it's just iterating over the internal search list and grabbing whatever's next in order at that point.

There's probably a couple more good fast paths to get the initial loading times down more, but I'd have to reach outside of the steno dictionary code and into the stroke parser for that. No more optimization for now :P

[benchmark.txt](https://github.com/openstenoproject/plover/files/2320372/benchmark.txt)

fourshade:
All squashed down and and reworded file-by-file. Should look a lot more like the other commits I see in the history.

fourshade:
Cleaned up a couple things, added the config option, and did a (hopefully) more reliable benchmark.

fourshade:
Gonna close this one for now and wait for the 4.0 release. Most of the search code is being used in a plugin I'm writing now, so it'll get a good bit of field testing in the meantime.

mulka:
Maybe I missed something, but why didn't this get merged? Waiting on the 4.0 release? Why?

marnen:
Yes, this would be awesome to have.

PR #1003, opened 2018-08-19
Prune longest (mostly useless) entries from main dict

fourshade:
Okay, here's a simple one. In order to support multi-stroke entries with possible boundary conflicts, Plover has to keep track of a list of the most recent strokes in order to see if your last stroke completed a phrase. For example, using this entry:

TKE/KHRAOEUPB/APBD/TPAUL/-FT/RO*EPL/A*PB/EPL/PAOEUR: Decline and Fall of the Roman Empire

You could enter:

TKE/KHRAOEUPB/APBD/TPAUL/-FT/RO*EPL/A*PB/EPL

and the system would translate all those strokes into normal words, but it still has to keep them in memory just in case you finish with PAOEUR, in which case it backspaces everything and replaces it all with the above phrase. Not only that, but it has to try looking up each possible sequence from the end in turn, up to the maximum length. That "maximum length" determines translation performance and is based on the longest sequence of strokes in any dictionary.

Right now that longest sequence is 11 strokes long. I removed the longest stroke sequences from the dictionary in order to bring this maximum down to 8. It turns out that all of these are either incredibly obscure medical/scientific terms, or capitalized titles of creative works that are useless unless you specifically know that they're there (otherwise you'd just manually capitalize each word). It could be brought down to 6 by removing about 100 more entries, but some of those are less obscure and I'd rather do this in steps if it can be helped.


SeaLiteral:
I think this is something that would be better put in a comment to issue #400 (dictionary suggestions). And I agree these entries and kind of useless. I just think the might be better to combine this with the other changes that people have suggested.

fourshade:
The purpose of this change is a bit different than the others. Rather than being for language/usability reasons, this change results in an immediate performance increase in the software. 10 near-useless entries removed in exchange for a non-trivial reduction in translation attempts...surely that's worth it?

benoit-pierre:
Unless this make a user noticeable impact on performance, I don't see the point. I do agree about pruning the dictionary of obscure medical entries, some translations for automatic casing, or workarounds Eclipse magic features, but this should be handled as part of #400.

fourshade:
It might not be user-noticeable to beginners, or people with fast machines, but people like Mirabai tax the hell out of their stenography systems on a daily basis. I know some people have contemplated putting Plover on an embedded system too (to make an all-in-one steno board), and every little bit helps there.

fourshade:
I'm gonna close this one and move on....there's a good optimization I'm thinking of that would make entries like these not affect performance nearly as badly anyway.

PR #1004, opened 2018-08-19
Add trailing newline when writing JSON dictionary

KeyboardFire:
The POSIX definition of a text file requires that it end with a newline
character.  The JSON files created through e.g. the ADD_TRANSLATION
command do not conform to this definition; this commit fixes that.


KeyboardFire:
Well, it depends on what you consider an "issue." I noticed because it added an extra change when I was committing some dictionary updates (I had previously been adding them with a text editor), and I think it's better to have it so that tools that expect it to be there will work properly.

PR #1005, opened 2018-08-21
steno_dictionary: speedup update implementation

benoit-pierre:
Twice as fast on my machine using a Magnum JSON dictionary as main.


fourshade:
I tried something like that at one point, and there's a rather serious problem with this: the iterable given to update could be (and is for json_dict) a generator, which gets consumed by the first method to use it. Then the for loop tries to consume it again, and there's nothing left, so the reverse dictionaries don't get updated.

Please take a look at my search overhaul if you can. It addresses quite a few of these sorts of issues.

benoit-pierre:
Indeed...

PR #1007, opened 2018-09-02
Fix read-only dict saving (#897)

### LINKED ISSUE #897: Error when attempting to edit a read-only dictionary
## Summary

When trying to edit a read only dictionary, Plover always crashes with the following trace:

``` cmd
O:\Software\Projects\plover>edit: editing failed
  File "O:\Software\Projects\plover\plover\gui_qt\dictionary_editor.py", line 384, in on_finished
    in self._model.modified)
  File "O:\Software\Projects\plover\plover\steno_dictionary.py", line 271, in save
    d.save()
  File "O:\Software\Projects\plover\plover\steno_dictionary.py", line 74, in save
    assert not self.readonly
AssertionError
```

Obviously if a dictionary is code / read only then someone shouldn't be wanting to edit it through the Dictionary Editor but it doesn't appear to be restricted.

## Reproducing

1. Add a read-only dictionary (like a .py dictionary via the plover-python-dictionary plugin)
2. Double click the dictionary in the main Plover window to open up the Dictionary Editor window
3. Click the "Add a new translation" button
4. Click the X on the Dictionary Editor window without having entered anything
5. Error out / crash

## Plover Version

## System

Windows 10

fourshade:
Fix #897.

I came across this method during my search code revisions and my IDE always marked it with warnings. Now that I see there's an issue with the program attempting to save dictionaries marked as read-only, it's a simple fix.

### DIFF
--- a/plover/steno_dictionary.py
+++ b/plover/steno_dictionary.py
@@ -276,10 +276,11 @@ class StenoDictionaryCollection:
 
         If <path_list> is None, all writable dictionaries are saved'''
         if path_list is None:
-            dict_list = [d for d in self if dictionary.save is not None]
+            dict_list = [d for d in self if not d.readonly]
         else:
             dict_list = [self[path] for path in path_list]
         for d in dict_list:
+            assert not d.readonly
             d.save()
 
     def get(self, path):


plover/steno_dictionary.py
```diff
@@ -264,11 +264,12 @@ def save(self, path_list=None):
 
         If <path_list> is None, all writable dictionaries are saved'''
         if path_list is None:
-            dict_list = [d for d in self if dictionary.save is not None]
+            dict_list = [d for d in self]
```
benoit-pierre:
```suggestion
            dict_list = [d for d in self if not d.readonly]
```

plover/steno_dictionary.py
```diff
@@ -264,11 +264,12 @@ def save(self, path_list=None):
...
         else:
             dict_list = [self[path] for path in path_list]
         for d in dict_list:
-            d.save()
+            if not d.readonly:
```
benoit-pierre:
```suggestion
            assert not d.readonly:
```

fourshade: So it throws an error instead of silently failing to write. Got it. Will rebase on the last merge too.

mulka: I'm not 100% sure I'm testing this code the right way, but won't an assert here crash the program? There's a second way that disc_list can be built and it doesn't look like it is checking for the readonly attribute.

fourshade: That's precisely the reason for the assert. Any code that calls this by manually specifying the paths is responsible for making sure those dictionaries aren't read-only. It's much easier to debug an AssertionError than a mysterious case of "I made changes and they didn't save."

km-pg: Ok, I just wanted to make sure there wasn't something I was missing. Like, may an assert just popped up an error window or something. An assert actually just stops Plover completely.

I'm thinking we should prevent this assert from triggering. With this change, if you try to edit a read-only dictionary, it will crash Plover.

mulka: Ok, I just wanted to make sure there wasn't something I was missing. Like, maybe an assert just popped up an error window or something. An assert actually just stops Plover completely.

I'm thinking we should prevent this assert from triggering. With this change, if you try to edit a read-only dictionary, it will crash Plover.

PR #1009, opened 2018-09-18
focus the input field when the lookup window is activated

### LINKED ISSUE #976: Using {PLOVER:LOOKUP} should put focus in text box
## Summary

I would like it whenever I used the {PLOVER:LOOKUP} command that the focus be put in the text box and maybe all the text highlighted so I can start typing right away and it would override whatever is already there.

## Reproducing

Have a stroke defined for {PLOVER:LOOKUP} if you don't already
Open the lookup tool.
Click in the larger of the two boxes as though you were going to copy the text there.
Without closing the lookup tool, click on a different application.
Use the {PLOVER:LOOKUP} stroke.
Notice that the text box where you would start typing is not in focus.

## Plover Version

4.0.0.dev8+8.ge062cda

## System

Mac 10.11.6

tux21b:
Automatically focus the input field and pre-select the previous input when the lookup window is activated.

Fixes #976

### DIFF
--- a/plover/gui_qt/lookup_dialog.py
+++ b/plover/gui_qt/lookup_dialog.py
@@ -43,3 +43,9 @@ class LookupDialog(Tool, Ui_LookupDialog):
         translation = unescape_translation(pattern.strip())
         suggestion_list = self._engine.get_suggestions(translation)
         self._update_suggestions(suggestion_list)
+
+    def changeEvent(self, event):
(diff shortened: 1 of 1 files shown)


mulka:
Just tested this. Seems to work well for me.

morinted:
Thanks for your contribution :)

PR #1010, opened 2018-09-18
fix ambiguous tool shortcuts

tux21b:
the tool shortcuts (e.g. Ctrl+L for the lookup window) are not working because the shortcuts are registered multiple times.


plover/gui_qt/main_window.py
```diff
@@ -105,7 +105,7 @@ def __init__(self, engine, use_qt_notifications):
                 action.setObjectName(tool_plugin.name)
                 if tool.__doc__ is not None:
                     action.setToolTip(tool.__doc__)
-                if tool.SHORTCUT is not None:
+                if tool.SHORTCUT is not None and parent == tools_menu:
```
morinted: Could you add a comment to the code to explain why this works? (both for me and for the next time someone ends up on this line of code)

morinted:
Thank you for pointing out the error here :)

PR #1011, opened 2018-09-19
fix dictionary entries that are not in steno order

tux21b:
The main dictionary shipped with Plover contains 4 entries that are not in Steno order.


morinted:
This is great. We should probably add a unit test that sanitizes the default dictionaries in our default system.

Another PR sorted the main.json, would you be willing to update this PR please?

benoit-pierre:
See #1038.

Issue #1013, opened 2018-09-23
KW-LS stroke produces error


amikumanto:
## Summary

With the current default dictionary, the stroke KW-LS produces a KeyError. I believe this it to be because it is mapped to `==` and so the first `=` is being interpreted as "the next thing should be a macro" and the second is being interpreted as the name of a macro. Since there is obviously no macro named `=` it causes an error. 

## Reproducing

1. Open Plover with the current default dictionary 
2. Write KW-LS

## Plover Version

Plover 4.0.0.dev8

## System

Windows 10

PR #1015, opened 2018-09-24
Fix: allow non-macros starting with =

### LINKED ISSUE #903: Better handle nonmacro translations that begin with =
## Summary

Macro support makes programming-style translations like `=`, `==`, `=>` hard to use, as they are treated as macros. A workaround is to add a do-nothing-meta: `{}=`

## Possible Solution

If we pattern-match the stroke to check for `^=\w+$` then most legitimate programming strokes should not be affected.

fourshade:
Fix #903, #1013.

Simple fix for the problem where programming symbols such as == are parsed as macros. Just try looking these entries up in the plugin registry as if they were macros, and when a key error pops up, catch it and continue translating it as a normal stroke. Should also help the case where someone incorrectly enters a macro in their dictionary; they'll see the raw stroke output as text instead of getting an arcane "on_stroke failed KeyError" (leading them to think the problem is with Plover and not their macro entry).

### DIFF
--- a/plover/translation.py
+++ b/plover/translation.py
@@ -300,18 +300,21 @@ class Translator:
         mapping = self.lookup([stroke])
         macro = _mapping_to_macro(mapping, stroke)
         if macro is not None:
-            self.translate_macro(macro)
-            return
+            # Look up the macro in the plugin registry and execute it.
+            # If it wasn't found, just translate the stroke as normal.
+            try:
+                macro_fn = registry.get_plugin('macro', macro.name).obj
+            except KeyError:
+                pass
+            else:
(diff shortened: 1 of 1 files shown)


fourshade:
Just occurred to me that registry lookup and execution of macros should be separated. The macros themselves could raise a KeyError that _shouldn't_ be caught and suppressed at this level, and inlining the method makes it clearer how the code proceeds when a lookup fails.

morinted:
@benoit-pierre are you okay with this change, since the macro implementation is yours?

morinted:
Thanks for the PR -- we've merged #1025 which addresses the issue as well

PR #1018, opened 2018-10-18
Fix #991: add keyup events for Unicode characters on Windows

### LINKED ISSUE #991: Custom system's Unicode key labels confuse Plover's verbatim stroke output
## Summary

[I wrote a plugin to implement my own system for Plover](https://github.com/contrum/plover-phenrsteno/). It uses Unicode's IPA symbols for its key labels:

<pre align=center>
ə p t w     f t ʃ l i
s b d r  *  n d g z ə

     a e   ɪ ʊ       
</pre>

Problem is that when I do `{PLOVER:ADD_TRANSLATION}`, the 'Strokes' text box consistently gets the IPA symbols wrong, whereas it works fine when programmed to use ASCII substitutes.

For example I stroke `bʊg` in vim, then do `{PLOVER:ADD_TRANSLATION}`, then I stroke `bɪgə` in the 'Strokes' text box but 'bʊgʊ' appears (I expect to see 'bɪgə' of course). So it seems to replace all Unicode characters with the one it last received outside of that dialogue box.

Coming from contrum/plover-phenrsteno#5

## Reproducing

1. Plug an NKRO keyboard into your computer
2. Install Plover 4.0.0.dev8+66.g685bd33 for Windows
3. Install [`plover-phenrsteno 0.1.19`](https://pypi.org/project/plover-phenrsteno/0.1.19/) from the Plugins manager
4. Restart Plover
5. Configure > System > Select 'Phenrsteno'
6. Enable output
7. Open your favourite text editor and stroke `bʊg` (`SML` on QWERTY)
8. In Plover click 'Add translation'
9. Stroke `bɪg` into the 'Strokes' text box (`SNL` on QWERTY)

You should expect to see `bɪg` appear in the text box, because that's what you stroked. However you actually see `bʊg`. You could also stroke, for example, `-ʃ` (`O` on QWERTY) but again see that LATIN SMALL LETTER UPSILON `ʊ` character instead.

## Plover Version

Plover 4.0.0.dev8+66.g685bd33

## System

Windows 10 Pro (build 17134.191)

fourshade:
Fix #991 and possibly others relating to Unicode.

I finally tracked down the main issue with Unicode output being corrupted in certain Windows GUIs (including Plover's). When sending Windows messages for keyboard input, the code branches based on whether the character has a standard keycode or needs the special Unicode flag. For standard keycodes, it makes sure to send messages for each key without the KEYUP flag, then send each one with it to release them. For Unicode, however, it only sent the keydown messages.

This one was hard to track down because it only affected certain windows. Depending how the receiving framework handled the messages, it might either work fine, or it might cause the first Unicode character to get "stuck" down. Since the "normal" keycode sent with Unicode characters is 0 (the real code is sent in a different field), it would look like the same key to some conditional checks, and so subsequent messages would appear to be an autorepeat of that character until the input handler was closed/restarted (usually by closing the window or application).

At least that's my theory. The fix is just to send KEYUP events after the KEYDOWN events just like the regular keys do.

### DIFF
--- a/plover/oslayer/winkeyboardcontrol.py
+++ b/plover/oslayer/winkeyboardcontrol.py
@@ -389,7 +389,7 @@ class KeyboardEmulation:
     # Keyboard input type to send key input
     @staticmethod
     def _keyboard_input(code, flags):
-        if flags == KEYEVENTF_UNICODE:
+        if flags & KEYEVENTF_UNICODE:
             # special handling of Unicode characters
             return KEYBDINPUT(0, code, flags, 0, None)
         return KEYBDINPUT(code, 0, flags, 0, None)
@@ -424,7 +424,9 @@ class KeyboardEmulation:
 
     def _key_unicode(self, char):
         pairs = to_surrogate_pair(char)
-        inputs = [self._keyboard(code, KEYEVENTF_UNICODE)
+        # Send press events for all codes, then release events for all codes.
+        inputs = [self._keyboard(code, KEYEVENTF_UNICODE | direction)
+                  for direction in (0, KEYEVENTF_KEYUP)
                   for code in pairs]
         self._send_input(*inputs)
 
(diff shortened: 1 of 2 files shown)


fourshade:
pip updated again, so the checksum changed. I'll change that if only so the checks can complete.

morinted:
Thanks for tracking down the source of this issue!

PR #1033, opened 2018-12-07
prevent editing of still loading or readonly dictionaries

### LINKED ISSUE #897: Error when attempting to edit a read-only dictionary

mulka:
Fixes #897 

Don't open the dictionary editor with dictionaries that are still loading or are readonly.

### DIFF
--- a/plover/gui_qt/dictionaries_widget.py
+++ b/plover/gui_qt/dictionaries_widget.py
@@ -316,8 +316,14 @@ class DictionariesWidget(QWidget, Ui_DictionariesWidget):
         self._update_dictionaries(dictionaries, record=False)
 
     def _edit(self, dictionaries):
-        editor = DictionaryEditor(self._engine, [d.path for d in dictionaries])
-        editor.exec_()
+        paths_to_edit = []
+        for d in dictionaries:
+            loaded_dictionary = self._loaded_dictionaries.get(d.path)
+            if loaded_dictionary and not loaded_dictionary.readonly:
+                paths_to_edit.append(d.path)
+        if paths_to_edit:
+            editor = DictionaryEditor(self._engine, paths_to_edit)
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Opening a read-only dictionary is a valid use-case: for inspecting the contents (e.g. using the filtering features). There's already #1007 for fixing #897.

mulka:
@benoit-pierre If you read the comments for #897, @nsmarkop says that "the end user should probably not even be able to open the editor for read-only dictionaries, but that can be solved separately if needed." and @morinted replied "even if we merged [#1007] I'd want to keep this open.". This is the issue I was hoping to address. I actually don't know how you would have a read only dictionary that's not a python dictionary that you would have to edit with a different type of editor any way. Maybe you could enlighten me? How should we address this issue?

PR #1036, opened 2019-01-20
Save default machine config options, including serial ports

### LINKED ISSUE #739: Scan serial ports automatically in settings
## Summary

Instead of requiring the user to scan for serial ports, we should do it for them on settings load or when clicking the drop-down.

morinted:
Fix #739

1. Always scan serial ports when rendering machine serial options
2. If the current value is None, set the first scanned port. That port was previously getting displayed but not saved because the "on_port_changed" method was being called before the config_window was connected. A singleShot is used to scan the ports again which will set the first port in settings.

### DIFF
--- a/plover/gui_qt/machine_options.py
+++ b/plover/gui_qt/machine_options.py
@@ -25,11 +25,14 @@ class SerialOption(QWidget, Ui_SerialWidget):
 
     def setValue(self, value):
         self._value = copy(value)
+        self.on_scan()
         port = value['port']
-        if port is None or port == 'None':
-            self.on_scan()
-        else:
-            self.port.setCurrentText(port)
+        if port is not None and port != 'None':
+            port_index = self.port.findText(port)
+            if port_index != -1:
+                self.port.setCurrentIndex(port_index)
+            else:
+                self.port.setCurrentText(port)
         self.baudrate.addItems(map(str, Serial.BAUDRATES))
         self.baudrate.setCurrentText(str(value['baudrate']))
         self.bytesize.addItems(map(str, Serial.BYTESIZES))
--- a/plover/gui_qt/config_window.py
+++ b/plover/gui_qt/config_window.py
@@ -435,8 +435,8 @@ class ConfigWindow(QDialog, Ui_ConfigWindow, WindowState):
     def _create_option_widget(self, option):
         widget = option.widget_class()
         widget.setToolTip(option.help_text)
-        widget.setValue(self._config[option.option_name])
         widget.valueChanged.connect(partial(self.on_option_changed, option))
+        widget.setValue(self._config[option.option_name])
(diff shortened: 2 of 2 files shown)


benoit-pierre:
How about this instead:
```diff
 plover/gui_qt/config_window.py   |  2 +-
 plover/gui_qt/machine_options.py | 11 +++++++----
 2 files changed, 8 insertions(+), 5 deletions(-)

diff --git i/plover/gui_qt/config_window.py w/plover/gui_qt/config_window.py
index 428cf736..98a99cf0 100644
--- i/plover/gui_qt/config_window.py
+++ w/plover/gui_qt/config_window.py
@@ -435,8 +435,8 @@ def _update_keymap(self, system_name=None, machine_type=None):
     def _create_option_widget(self, option):
         widget = option.widget_class()
         widget.setToolTip(option.help_text)
-        widget.setValue(self._config[option.option_name])
         widget.valueChanged.connect(partial(self.on_option_changed, option))
+        widget.setValue(self._config[option.option_name])
         return widget
 
     def keyPressEvent(self, event):
diff --git i/plover/gui_qt/machine_options.py w/plover/gui_qt/machine_options.py
index 7ecb5691..61a7a60e 100644
--- i/plover/gui_qt/machine_options.py
+++ w/plover/gui_qt/machine_options.py
@@ -25,11 +25,14 @@ def __init__(self):
 
     def setValue(self, value):
         self._value = copy(value)
+        self.on_scan()
         port = value['port']
-        if port is None or port == 'None':
-            self.on_scan()
-        else:
-            self.port.setCurrentText(port)
+        if port is not None and port != 'None':
+            port_index = self.port.findText(port)
+            if port_index != -1:
+                self.port.setCurrentIndex(port_index)
+            else:
+                self.port.setCurrentText(port)
         self.baudrate.addItems(map(str, Serial.BAUDRATES))
         self.baudrate.setCurrentText(str(value['baudrate']))
         self.bytesize.addItems(map(str, Serial.BYTESIZES))
```
This way other type of options could dynamically set a default too.

morinted:
@benoit-pierre thanks for the suggestion -- much simpler and it makes more sense overall. I updated the PR and tested on my machine.

plover/gui_qt/machine_options.py
```diff
@@ -25,11 +25,15 @@ def __init__(self):
 
     def setValue(self, value):
         self._value = copy(value)
+        self.on_scan()
         port = value['port']
-        if port is None or port == 'None':
-            self.on_scan()
-        else:
-            self.port.setCurrentText(port)
+        first_port = self.port.currentText
```
benoit-pierre: This line can be removed.

morinted: Thanks, fixed

PR #1047, opened 2019-02-15
Exit edit mode before modifying the model

### LINKED ISSUE #1040: Trying to add a definition while editing one in the dictionary editor makes plover crash
## Summary

Plover crashes when clicking the "add" button in the dictionary editor while editing an existing entry.

## Reproducing

1. Select a dictionary
2. Click the pencil to edit
3. Click the green plus twice
4. Plover crashes

## Plover Version

4.0.0.dev8+66.g685bd33

## System

Running Lubuntu 18.04

user202729:
Fixes #1040.

It appears that `QTableView` has problem when the model is modified while it's in edit state.

### DIFF
--- a/plover/gui_qt/dictionary_editor.py
+++ b/plover/gui_qt/dictionary_editor.py
@@ -361,6 +361,7 @@ class DictionaryEditor(QDialog, Ui_DictionaryEditor, WindowState):
             row = self._selection[0]
         else:
             row = 0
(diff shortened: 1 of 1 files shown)


morinted:
Thanks for the fix

benoit-pierre:
I never managed to reproduce the error. At the very least it needs a comment as to why this is necessary.

PR #1048, opened 2019-02-15
Refactor adding tool actions

user202729:
It's cleaner to only put common code in the loop instead of putting everything in the loop and use if statements to distinguish them.

Also fix ambiguous shortcut.


morinted:
Thanks for the PR! -- #1062 addressed the same bit of code.

PR #1054, opened 2019-03-06
automatically set machine state to error upon disconnect of writer

tschulte:
Most of the implementations of the run method of the ThreadedStenotypeBase
do not catch exceptions. This causes the thread to silently die (the
exception is only logged to the console).

Now an exception in the run method is reported as an error to the system
and additionally the SerialStenotypeBase class does close the com port on
error.

Otherwise the com port resource might still be taken by plover upon
reconnect of the writer, and a different com port might be taken by the OS
and plover might not be able to reconnect to that machine.

This is a first part to solving issue #596. In addition there must be some
mechanism to retry connecting when the state is error.

### DIFF
--- a/plover/machine/base.py
+++ b/plover/machine/base.py
@@ -77,7 +77,7 @@ class StenotypeBase:
 
     def add_state_callback(self, callback):
         self.state_subscribers.append(callback)
-        
+
     def remove_state_callback(self, callback):
         self.state_subscribers.remove(callback)
 
@@ -119,7 +119,7 @@ class StenotypeBase:
 
     def _ready(self):
         self._set_state(STATE_RUNNING)
-            
+
     def _error(self):
         self._set_state(STATE_ERROR)
 
@@ -137,17 +137,29 @@ class StenotypeBase:
         """Get the default options for this machine."""
         return {}
 
+class _ThreadedStenotypeThread(threading.Thread):
+    def __init__(self, machine):
+        threading.Thread.__init__(self)
+        self.machine = machine
 
-class ThreadedStenotypeBase(StenotypeBase, threading.Thread):
+    def run(self):
+        try:
+            self.machine.run()
+        except Exception:
+            if not self.machine.finished.isSet():
+                log.error(self.machine.name + ' got disconnected.')
+                self.machine._error()
+
+class ThreadedStenotypeBase(StenotypeBase):
     """Base class for thread based machines.
-    
+
     Subclasses should override run.
     """
     def __init__(self):
-        threading.Thread.__init__(self)
-        self.name += '-machine'
+        self.name = self.__class__.__name__ + '-machine'
         StenotypeBase.__init__(self)
         self.finished = threading.Event()
+        self._machine_thread = _ThreadedStenotypeThread(self)
 
     def run(self):
         """This method should be overridden by a subclass."""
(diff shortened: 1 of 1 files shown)


tschulte:
I now created a (plugin https://github.com/tschulte/plover_auto_reconnect_machine) that does the reconnect. But that plugin relies on this pull request to be applied.

plover/machine/base.py
```diff
@@ -137,17 +137,29 @@ def get_option_info(cls):
...
+    def __init__(self, machine):
+        threading.Thread.__init__(self)
+        self.machine = machine
+
+    def run(self):
+        try:
+            self.machine.run()
+        except Exception:
+            if not self.machine.finished.isSet():
+                log.error(self.machine.name + ' got disconnected.')
```
morinted, line 150: Can we simplify the message to just "GeminiPr disconnected"?

```suggestion
                log.error('%s disconnected', self.machine.name)
```

Too bad there's no easy way to get the plugin name from what I can see.

plover/machine/base.py
```diff
@@ -137,17 +137,29 @@ def get_option_info(cls):
...
+class ThreadedStenotypeBase(StenotypeBase):
     """Base class for thread based machines.
-    
+
     Subclasses should override run.
     """
     def __init__(self):
-        threading.Thread.__init__(self)
-        self.name += '-machine'
+        self.name = self.__class__.__name__ + '-machine'
```
morinted, line 159:
```suggestion
        self.name = self.__class__.__name__
```

pianohacker:
@tschulte Are you still interested in getting this in? Would really like this functionality. Happy to do the last suggested edits if not.

tschulte:
@pianohacker yes, in principle I am interested in getting this in, although I currently don't have this issue any more, since my setup is different at the moment. Sorry that this pull request dropped of my mind.

As I recall, a problem might be, that these classes seem to be part of the API, or at least there are plugins that extend these classes. Therefore changing these might cause them to break. At least for example the treal machine is implemented as a plugin, and it does also implement the reconnect logic. But I think that would have only been an issue with my first attempt where I copied that reconnect logic from treal to ThreadedStenotypeBase. In the current sollution that should not be a problem. Even the treal machine automatically handling the reconnect itself should not be a problem.

I reset --hard my branch to your rebased branch and pushed it. Therefore this pull request now contains your changes.

user202729:
Remark: with a recent Plover/system update, this has a weird behavior that I haven't seen before:

If I do this

* Connect a machine
* Sleep the computer
* Wake it up

=> The machine stopped responding; however the "Machine disconnected" message is not shown until the USB is unplugged.

If I additionally do this

* Unplug the USB
* Plug it again **before** the "machine is disconnected" message is shown

Then the (linked) plugin will repeatedly try (and fail) to connect to the old port; while the machine is moved to a new port.

If I unplug the USB in the process, this message will be printed

```
> Exception in thread Thread-4:                                                                                  
Traceback (most recent call last):utput: Enabled | System: English Stenotype |                                   
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user202729/plover_auto_reconnect_machine/plover_auto_reconnect_machine.py", line 43, in run
    if isinstance(self._engine._machine, SerialStenotypeBase) and not self._port_exists(self._engine._machine.se:
  File "/home/user202729/plover_auto_reconnect_machine/plover_auto_reconnect_machine.py", line 65, in _port_exiss
    for port in list_ports.comports():
  File "/usr/lib/python3.9/site-packages/serial/tools/list_ports_linux.py", line 102, in comports
    for info in [SysFS(d) for d in devices]
  File "/usr/lib/python3.9/site-packages/serial/tools/list_ports_linux.py", line 102, in <listcomp>
    for info in [SysFS(d) for d in devices]
  File "/usr/lib/python3.9/site-packages/serial/tools/list_ports_linux.py", line 52, in __init__
    self.vid = int(self.read_line(self.usb_device_path, 'idVendor'), 16)
TypeError: int() can't convert non-string with explicit base
```

and the plugin stopped working completely.

PR #1066, opened 2019-04-26
steno_dictionary: fix `StenoDictionaryCollection.casereverse_lookup`

benoit-pierre:
Don't stop at the first result, potentially ignoring lower priority dictionaries. This also tweak the `StenoDictionary` code to be slightly faster on load.

Timings:
* `main.json` (4.02 MiB, 147424 entries): `0.781s` -> `0.625s`
* `german_palantype.json` (58.29 MiB, 1466824 entries): `13.492s` -> `10.511s`
* magnum.json (18.81 MiB, 616946 entries): `5.053s` -> `4.177s`
* `magnum.rtf` (45.8 MiB, 616946 entries): `22.953s` -> `22.731s` (yes RTF is still slow as hell, much better with the new parser: `14.338s`)

### DIFF
--- a/plover/steno_dictionary.py
+++ b/plover/steno_dictionary.py
@@ -35,7 +35,7 @@ class StenoDictionary:
         self._longest_listener_callbacks = set()
         self.reverse = collections.defaultdict(list)
         # Case-insensitive reverse dict
-        self.casereverse = collections.defaultdict(collections.Counter)
+        self.casereverse = collections.defaultdict(list)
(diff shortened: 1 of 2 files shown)


morinted:
Is it possible returning a set will break some external usage in plugins?

I guess only if they were mutating the returned value. Should be okay right?

benoit-pierre:
I think it should be okay, yes. Better to change the API now than after 4.0.0 in any case!

Issue #870, opened 2017-12-04
Add friendly names for built-in metas


morinted:
## Summary
We should add friendly names for built-in metas like was done for macros.

nsmarkop:
Just for my clarification, is this to get all the sorts of things for capitalization, currency, etc. in https://github.com/openstenoproject/plover/wiki/Dictionary-Format#text-formatting extracted out to a place like plover.meta by implementing them all as meta plugins with nice names? With aliases to the old syntax for use in plover.formatting.

PR #1069, opened 2019-04-27
Make using builtin commands / metas more user-friendly.

### LINKED ISSUE #688: Commands should be made case-insensitive
Command plugins were made case-insensitive, but it only applies for the part after the PLOVER. e.g. `{PLOVER:COMMAND}` & `{PLOVER:command}` work but `{Plover:command}` doesn't.

Similarly, our built-in commands (focus, toggle, quit) are all case sensitive and shouldn't be.

benoit-pierre:
Add user-friendly names for builtin metas (e.g. `{:retro_currency:$c}`), and ignore case when processing builtins commands / metas (so `{PloveR:toggle}` or `{:mode:Camel}` will work as expected).

Fix #688 and #870.


plover/engine.py
```diff
@@ -318,10 +318,11 @@ def _on_machine_state_changed(self, machine_state):
...
-        if command == 'RESUME':
+        command = command.lower()
+        if command == 'resume':
             self._set_output(True)
             return True
-        elif command == 'TOGGLE':
+        elif command == 'toggle':
             self._toggle_output()
             return True
         elif command == 'QUIT':
```
hwrod: `QUIT` should be lowercased here

benoit-pierre: Good catch!

plover/formatting.py
```diff
@@ -60,6 +38,83 @@
...
+    (r'<'    , 'case'      , Case.UPPER_FIRST_WORD.value),
+    (r'\*-\|', 'retro_case', Case.CAP_FIRST_WORD.value  ),
+    (r'\*>'  , 'retro_case', Case.LOWER_FIRST_CHAR.value),
+    (r'\*<'  , 'retro_case', Case.UPPER_FIRST_WORD.value),
+    # Mode.
+    (r'MODE:(.*)', 'mode', 0),
+    # Currency.
+    (r'\*\((.*)\)', 'retro_currency', 0),
+    # Glue.
+    ((r'{0}(.*)', META_GLUE_FLAG), 'glue' , 0),
```
morinted: just syntax: space before comma after 'glue'

morinted:
We may also consider `affix` over attach

benoit-pierre:
> What is the behavior of `{:attach}`? Would it be like `{:attach:^ing}`, with `{:attach:word}` having no effect?

```python
    def test_meta_attach_4(self):
        r'''
        'TEFT': 'test',
        'AT': '{:attach:attach}',

        TEFT/AT  ' test attach'
        '''
```

benoit-pierre:
> We may also consider `affix` over attach

Sure, do if you think it's clearer, do you want me to change it?

benoit-pierre:
Regarding the behavior of `{:attach:attach}`, we could change it so it's equivalent to `{:attach:^attach^}`.

morinted:
> 
> 
> Regarding the behavior of `{:attach:attach}`, we could change it so it's equivalent to `{:attach:^attach^}`.

I like both of these suggestions if you want to add the changes to this PR

benoit-pierre:
@morinted: I changed the default for `{:attach:attach}`. Just to be clear, you want also the meta to be renamed to `affix`?

morinted:
Thanks for the changes. I think "affix" is more technically correct, but "attach" will probably be easier to understand for users. Let's merge as-is.

PR #1077, opened 2019-06-15
Update PyQt5 dependencies

sidd:
**1. The issues or a description of the problem this PR fixes**

Fixes dark mode visuals on macOS. The screenshot below shows the behavior on `master`, and the screenshot below on the right shows the behavior with this patch:

image

**2. Describe what you did, and if this PR has any open questions or requires more testing.**

Just bumped some dep versions. Seemed harmless enough, but please let me know if this introduces any regressions. The same tests are failing on this branch as compared to `master`.


benoit-pierre:
I plan to update PyQt5, but it's a little more involved than just updating `requirements_distribution.txt` as:
* newer versions don't include `QtWebEngine` (it's now shipped as a separate wheel)
* and they are not compatible with Trusty, so updating breaks the AppImage (we need to update our docker image to Xenial)

sidd:
Good to know, thanks for the explanation. I'm not super familiar with PyQt. Feel free to close/ignore this PR since it seems pretty low priority.

PR #1082, opened 2019-08-03
Load non-us keyboard related keysyms from xkb

Elktro:
Enables the usage of ISO level 3 shift and thus makes possible to use the AltGr-modifier key found, for example on the Finnish keyboard. The usage is {#ISO_Level3_Shift(...)}.


morinted:
LGTM -- any problems possible from your side @benoit-pierre ?

PR #1084, opened 2019-08-07
Show dictionary on add translation

Elktro:
When adding new or checking existing stroke, the add translation dialog did not show in which dictionary mapped entries belong to.

On the add translation dialog, the mapping information on the stokes written to the "Strokes" text box will now show the dictionary in which the currently used mapping is defined as well as other mappings found in other dictionaries. This also fixes bug in which filtering from one dictionary mapping prevented other mappings to be used from other dictionaries.


morinted:
I'm excited about this PR! Some screenshots

Before:

Screen Shot 2020-04-03 at 11 04 01 AM


After:

Screen Shot 2020-04-03 at 10 33 14 AM

--------------------

Struggling with the "It is also mapped to:" sentence, wondering if we should consider something like…

- "Other dictionaries:"
- "Masked entries:"

--------------------

Do we want to consider a different format, too? Something with less English:

> `PWHERP`:
> • blerp *custom.json* (Active)
> • blerpoo *personal.json*
> • blerp *main.json*

With the dictionary text color maybe a gray so that the translations stand out.

morinted:
My previous screenshots didn't include the bold font present in the PR

Screen Shot 2020-04-05 at 7 33 16 PM

morinted:
I made my suggested changes, the result is
Screen Shot 2020-10-14 at 2 24 10 PM

plover/gui_qt/add_translation_widget.py
```diff
@@ -211,28 +232,52 @@ def on_dictionary_selected(self, index):
...
+    def _update_strokes(self):
         strokes = self._strokes()
         if strokes:
-            translation = self._engine.raw_lookup(strokes)
-            if translation is not None:
-                fmt = _('{strokes} maps to {translation}')
+            translations = self._engine.raw_lookup_from_all(strokes)
+            if translations:
+
+                info = self._format_label(_('{strokes} maps to '), (strokes,))
```
morinted, line 256:
```suggestion
            if translations:
                info = self._format_label(_('{strokes} maps to '), (strokes,))
```

morinted (approved):

PR #1092, opened 2019-12-10
Updated ortho rules for Magnum/Stanley's theory rules..

stanographer:
Added rules that account for orthographical errors that I've encountered while using Plover using the non-standard dictionary. Please review for efficiency suggestions or possible consolidations of rules as most of these, I added as I came across them.

Also added this line `from PyQt5.QtWebEngineWidgets import QWebEngineView` to`registry.py`  that was causing errors on launch.


user202729:
You could add `y+ify` to the `i+ify` ones. Such as `pretty+ify`.

benoit-pierre:
> Also added this line from PyQt5.QtWebEngineWidgets import QWebEngineView toregistry.py that was causing errors on launch.
What errors? I think I remember a similar error due to a hack in one plugin (spectra-lexer I think), but I believe this has been resolved. In any case this does not belong in this PR, and certainly not in a commit with no relation.

benoit-pierre:
Closing in favor of #1212.

PR #1096, opened 2019-12-31
Add missing orthography rule alias

user202729:
Stroking `RUS/-BLT` will result in "reducibility" instead of "reducability".

I know it's possible to get that word with `RUS/IBLT`, but this is for consistency with `"able": "ible"`.


morinted:
Thanks for this change—pulled it and it worked well and didn't conflict with something like traffic + -BL for trafficable

PR #1106, opened 2020-03-03
Add Ubuntu specific instructions to Linux development readme

smahood:
Here are the steps I had to follow to get a development environment setup on a recent Ubuntu 18.04 install. I haven't tried it on a completely fresh install, so it's possible that I missed some dependencies that I already installed. 

I'm also not totally sure if `python-dbus` is required, and I assumed that versions of python newer than 3.6 would work fine. If anything looks off, I can try it on a fresh install.

### DIFF
--- a/linux/README.md
+++ b/linux/README.md
@@ -20,3 +20,28 @@ To install the standard plugins, you can use:
 
 * `./launch.sh`: run from source
 * `./test.sh`: run tests
+
+
+## Setup on Ubuntu 18.04
+Check Python3 version - version 3.6 or higher required
+`python3 -v`
+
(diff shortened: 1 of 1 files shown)


morinted:
Thanks for this pull request! I agree that your documentation is more explicit about installing on Ubuntu 18.04, but I don't think it actually adds anything beyond what we have now. (The existing instructions provide all the information you need to install on any distribution, and it's left up to the user to fill in some of the blanks—pretty normal on Linux.)

This is normal for most projects so for now I'm going to close this PR. Again, thanks for submitting it and please feel free to contribute more in the future.

PR #1121, opened 2020-04-06
Update appimage build

benoit-pierre:
* use the new Xenial image when using packpack
* switch to using linuxdeploy for AppDir creation
* trim some more unnecessary PyQt5 stuff, most notably the plugin manager does not use WebEngine anymore (which itself is provided as a separate wheel for recent PyQt5 releases)

### DIFF
--- a/linux/appimage/build.sh
+++ b/linux/appimage/build.sh
@@ -84,7 +84,7 @@ run cp 'plover/assets/plover.png' "$appdir/plover.png"
 
 # Trim the fat.
 run cp linux/appimage/blacklist.txt "$builddir/blacklist.txt"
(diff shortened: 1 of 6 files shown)


linux/appimage/build.sh
```diff
@@ -49,97 +85,70 @@ then
...
+  } | xargs -0 --no-run-if-empty strip
 }
 run strip_binaries
 
+# Finalize the AppDir.
+# Note:
+# - use a custom launcher that does not change the working directory.
+# - temporarily move PyQt5 out of the way so linuxdeploy does not try
+#   to bundle its system dependencies.
+run mv "$pydist/$pypurelib/PyQt5" "$builddir"
```
morinted, line 134: Does builddir need a trailing slash to ensure

Issue #589, opened 2016-09-11
Conversion of rtf dictionary to json fails in some cases due to linebreaks


percidae:
I've used @benoit-pierres script [convert_dictionary.py](https://gist.github.com/benoit-pierre/0ae7f388aa367ecbd0c8aa5e1de2b9f0) to convert a rtf dictionary to json format. The rtf file has lots of unnecessary linebreaks, also during stroke definitions. In some cases but not all cases this causes the conversion to fail.

Attached is an example [rtf-dictionary](https://github.com/openstenoproject/plover/files/466239/example.rtf.txt) with some working entries with linebreaks in between and some entries with some linebreaks which do not work. See the output in the attached [example.json](https://github.com/openstenoproject/plover/files/466240/example.json.txt).
(I had to add a *.txt to both files to be allowed to upload it to github).

morinted:
Could you pull out some samples from the converted files to show, e.g., working with linebreaks vs. broken with linebreaks?

percidae:
rtf:
```
{\*\cxs #1*-E7S/SKP/TKPWR-EUGS/S-FSZ}the Citizenship and Immigration Services{\*\cxsvatdictentrydate
\yr2015\mo12\da7}
{\*\cxs #1*-E7S/SKP/TKPWR-EUPLGS/S-FSZ}the Citizenship and Immigration Services{
\*\cxsvatdictentrydate\yr2015\mo12\da7}
{\*\cxs #1*-EU7/SKP/TKPWR-EUGS/S-FSZ}the Citizenship and Immigration Services{\*\cxsvatdictentrydate
\yr2015\mo12\da7}
```

The first one and the third one work, the second one does not.

json:
```
"1*E7S/SKP/TKPWREUGS/S-FSZ": "the Citizenship and Immigration Services",
"1*E7S/SKP/TKPWREUPLGS/S-FSZ": "the Citizenship and Immigration Services{\r\n\\*\\cxsvatdictentrydate\\yr2015\\mo12\\da7}",
"1*EU7/SKP/TKPWREUGS/S-FSZ": "the Citizenship and Immigration Services",
```

morinted:
Thanks for at the samples!

PR #1130, opened 2020-06-07
Fix RTF newline bugs, add support for digitalCAT's ^ as a space

psethwick:
## Summary of changes
RTF files can have newlines practically anywhere, which still applies when you add the CRE.
Amended the loading of RTF files to treat un-escaped newlines as if they did not exist.

Also added support for using ^ as a space. DigitalCAT exports have it all over the place, usually after punctuation.
Old behavior was: keep it as is, but in Plover ^ has a special meaning .. and it's about the exact opposite.

### Pull Request Checklist
- [ :heavy_check_mark:] Changes have tests
- [ :heavy_check_mark:] News fragment added in news.d. See [documentation](https://github.com/openstenoproject/plover/blob/master/doc/developer_guide.rst#making-a-pull-request) for details


user202729:
This should fix #589 .

The first huge regex should probably use `re.VERBOSE`, if the main developers agree to use that. It's already very hard to read.

Other than that, looks good (assumes that it works correctly. I don't actually know the RTF specification)

psethwick:
Yeah, it should fix #589 unless I've missed something.

I didn't want to add a dependency if I could help it, the project seems like it wants to keep those low (and I generally agree), although happy to take advisement from maintainers. Would hope that my regex is easier to read than what I've replaced!

The one thing I'm not sure of is whether this ought to do the thing with ^. It's not really in the RTF/CRE spec, and other programs may use it differently.

This still doesn't get us round-trip RTF <-> plover format, which is also a shame

benoit-pierre:
#1364 would take care of properly parsing RTF, handling newlines in the middle of steno/translation.

> Also added support for using ^ as a space. DigitalCAT exports have it all over the place, usually after punctuation.

Do you have an actual example RTF with some entries using `^`? From the official RTF/CRE specification specifically mention not adding new control characters to avoid potential conflicts with future evolutions to the RTF standard:

> III.D      Invalid Control Symbols Are Out
>
> Another assumption is that creating new control symbols (as opposed to control words) is not recommended because of potential future conflicts.  For instance, the hard space in RTF is defined as the control symbol backslash-tilde (\~).  To implement the CRE concept of the delete-space, therefore, it would make sense to use a control symbol -- say backslash-plus sign (\+).  Unfortunately, there are so few symbols available that it’s assumed that future versions of RTF will define a meaning for all of them, which would conflict with the CRE.  For that reason, we don’t use any undefined control symbols in the CRE, even when it would make sense to do so.

The RTF standard already provide support for hard-space: the `\~` control character.

The Stenovations manual uses `^`, but without an example of the resulting export to RTF, I don't see how you can conclud that `\^` is going to be used for that (logically, those `^` would converted to `\~`).

psethwick:
> #1364 would take care of properly parsing RTF, handling newlines in the middle of steno/translation.

Well I hardly think that's exactly fair, it's not like that was existing work when I opened this!

But if you wanna just close I guess whatever :/

I don't feel strongly about ^

benoit-pierre:
> ```
> {\*\cxs W*-EFPLGTD}we got.^{\*\cxsvatdictflags N}{\*\cxsvatdictentrydate\yr2016\mo8
> [...]
> {\*\cxs STKPWHR*-FPLT/P-P}Z.^{\*\cxsvatdictflags N}{\*\cxsvatdictentrydate\yr2014
> ❯ grep '\^' zqMagnum\ Steno\ Dictionary\ 2-25-2020.rtf | wc -l
> 829
> ```

So how would you insert a literal `^`?

psethwick:
> So how would you insert a literal ^?
There is a test for that, it works with \^

But.. I'm not sure you understand how much I agree with you

it's not in the spec, it's some gonzo stenovations format, sure.. but it is _very_ wonky

I wasn't sure about it when I wrote this PR _a year ago_ and, well, I'm not even using Magnum anymore

psethwick:
The real reason I opened this was the newline bugs, and if you have a newer solution for that then I don't see a reason to think about this PR any more

PR #1139, opened 2020-08-19
Argument for custom command plugins should not be forced lowercase

user202729:
## Summary of changes

The argument received by command plugins is no longer converted to lower case.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -347,7 +347,7 @@ class StenoEngine:
             self._trigger_hook('lookup')
         else:
             command_fn = registry.get_plugin('command', command_name).obj
-            command_fn(self, command_args[0] if command_args else "")
+            command_fn(self, command_args[0] if command_args else '')
         return False
 
     def _on_stroked(self, steno_keys):
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -321,6 +321,7 @@ class StenoEngine:
 
     def _consume_engine_command(self, command):
         # The first commands can be used whether plover has output enabled or not.
+        original_command = command
         command = command.lower()
         if command == 'resume':
             self._set_output(True)
@@ -347,7 +348,7 @@ class StenoEngine:
         else:
             command_args = command.split(':', 1)
             command_fn = registry.get_plugin('command', command_args[0]).obj
-            command_fn(self, command_args[1] if len(command_args) == 2 else '')
+            command_fn(self, original_command[len(command_args[0])+1:])
         return False
 
     def _on_stroked(self, steno_keys):
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -321,34 +321,33 @@ class StenoEngine:
 
     def _consume_engine_command(self, command):
         # The first commands can be used whether plover has output enabled or not.
-        original_command = command
-        command = command.lower()
-        if command == 'resume':
+        command_name, *command_args = command.split(':', 1)
+        command_name = command_name.lower()
+        if command_name == 'resume':
             self._set_output(True)
             return True
-        elif command == 'toggle':
+        elif command_name == 'toggle':
             self._toggle_output()
             return True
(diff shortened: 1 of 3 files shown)


user202729:
On the other hand, some plugin might find this undesirable (for example the built-in `set_config` plugin, currently it does not differentiate between lowercase and uppercase keys)

plover/engine.py
```diff
@@ -321,6 +321,7 @@ def _on_machine_state_changed(self, machine_state):
 
     def _consume_engine_command(self, command):
         # The first commands can be used whether plover has output enabled or not.
+        original_command = command
```
morinted, line 324:
```suggestion
        command_name, command_args* = command
        # Lowercase command name but not the arguments.
        command_name = command_name.lower()
        command_args = command_args[0] if len(command_args) else ''
```

plover/engine.py
```diff
@@ -321,6 +321,7 @@ def _on_machine_state_changed(self, machine_state):
...
         command = command.lower()
         if command == 'resume':
```
morinted, line 326: If the above suggestion is accepted, command_name would be used instead of "command"

plover/engine.py
```diff
@@ -347,7 +348,7 @@ def _consume_engine_command(self, command):
         else:
             command_args = command.split(':', 1)
             command_fn = registry.get_plugin('command', command_args[0]).obj
-            command_fn(self, command_args[1] if len(command_args) == 2 else '')
+            command_fn(self, original_command[len(command_args[0])+1:])
```
morinted, line 351: and finally here, command_args would be sent in.

morinted:
It would also be great to open a bug describing the issue that not lowercasing the command arguments causes.

plover/engine.py
user202729: This would need another string split (instead of string slice), but that's obviously fast enough.

user202729:
It depends on which plugin is used. For the built-in plugins, it's not a problem.

In my case I was using [my plugin to run shell command from Plover](https://github.com/user202729/plover_run_shell).

user202729:
Turns out that the `set_config` plugin was wrong all the time (`{plover:set_config:"space_placement":"After Output"}` doesn't work because the parameter is forced lowercase), and the unit test can't catch it. Should revert that commit.

plover/engine.py
```diff
@@ -321,33 +321,33 @@ def _on_machine_state_changed(self, machine_state):
...
             self._trigger_hook('add_translation')
-        elif command == 'lookup':
+        elif command_name == 'lookup':
             self._trigger_hook('lookup')
         else:
-            command_args = command.split(':', 1)
-            command_fn = registry.get_plugin('command', command_args[0]).obj
-            command_fn(self, command_args[1] if len(command_args) == 2 else '')
+            command_fn = registry.get_plugin('command', command_name).obj
+            command_fn(self, command_args[0] if command_args else "")
```
morinted, line 350: Just single quotes for our code-style, otherwise should be good to go.

```suggestion
            command_fn(self, command_args[0] if command_args else '')
```

fourshade:
Good find! I have a development command plugin that feeds its argument directly into Python's exec(); lowercasing that could screw up identifiers. Haven't been bitten yet, but it would have only been a matter of time.

morinted:
If you mark this PR ready for review, I will merge it 👍

user202729:
Should the commits be squashed?

There's still no tests or news.d.

morinted:
No need to squash the commits, we can do that during merge.

A news.d line would be great, if there's a way to detect the mistake with `{plover:set_config:"space_placement":"After Output"}` then that would make a unit test. If not, let's move forward with this bug fix.

morinted:
I've added a news fragment

morinted (approved):

PR #1152, opened 2020-10-16
Use C launcher instead of sh to fix Mac permissions issues

### LINKED ISSUE #1134: macOS: Requires granting Accessibility access to env (security issue)
## Summary

In macOS Catalina, there are stricter security requirements than in previous OS releases. Plover requires Accessibility access in the Security & Privacy settings to work. However, the actual keyboard emulation appears to be attributed to `env`, and so [`env` must be given Accessibility access for Plover to work](https://github.com/openstenoproject/plover/issues/1131). But this means that another process could use `env` to perform malicious actions.

Ideally, the keyboard emulation should be attributed to Plover itself. Failing that, it should be attributed to a Plover-specific helper process so that the user doesn't have to grant access to something as powerful as `env`.

## Reproducing

1. Install Plover on Catalina.
2. Try to type with it.
3. See prompt to give access to `env`, and no output.
4. Enable access for `env` in Security & Privacy.
4. Typing now works.

## Plover Version

weekly-v4.0.0.dev8+66.g685bd33

## System

macOS 10.15.5 (19F101

morinted:
## Summary of changes

- Use C launcher instead of python launcher
- Add permissions for automation

Fixes #1134

### Pull Request Checklist
- ~Changes have tests~


osx/make_app.sh
```diff
@@ -76,8 +76,9 @@ bootstrap_dist "$plover_wheel"
 # Make launcher
 plover_executable=MacOS/Plover
 launcher_file="$app_dir/Contents/$plover_executable"
-sed "s/pythonexecutable/$target_python/g" "$osx_dir/app_resources/plover_launcher.sh" >"$launcher_file"
-chmod +x "$launcher_file"
+sed "s/pythonexecutable/$target_python/g" "$osx_dir/app_resources/plover_launcher.c" > "$launcher_file.c"
+gcc "$launcher_file.c" -o "$launcher_file"
```
benoit-pierre, line 80:
```suggestion
gcc -Wall -O2 "$launcher_file.c" -o "$launcher_file"
```

osx/app_resources/plover_launcher.c
```diff
@@ -0,0 +1,41 @@
...
+    uint32_t size = sizeof(path);
+    _NSGetExecutablePath(path, &size);
+
+    // Remove filename from path.
+    char * last_slash;
+    last_slash = strrchr(path, '/');
+    path[last_slash - path] = '\0';
+
+    // Join file path and relative path to file.
+    strcat(path, python_path);
```
benoit-pierre, line 26: It's best to avoid `strcat`: it's a common source of buffer overruns (if the buffer is not big enough).

osx/app_resources/plover_launcher.c
```diff
@@ -0,0 +1,41 @@
...
+
+    char **python_args = malloc((argc + ARGS_LIST_SIZE) * sizeof(char*));
+
+    for (int i = 0; i < ARGS_LIST_SIZE; i++) {
+      python_args[i] = arg_list[i];
+    }
+    for (int i = 1; i < argc; i++) {
+      python_args[i + ARGS_LIST_SIZE - 1] = argv[i];
+    }
+    python_args[argc + ARGS_LIST_SIZE - 1] = NULL;
```
benoit-pierre, line 38: That's convoluted, why not use simple assignments + `memcpy`?

osx/app_resources/plover_launcher.c
```diff
@@ -0,0 +1,41 @@
...
+
+    execve(python_args[0], python_args, environ);
```
benoit-pierre, line 40:
```suggestion
    return execv(python_args[0], python_args);
```

osx/app_resources/plover_launcher.c
```diff
@@ -0,0 +1,41 @@
...
+#define ARGS_LIST_SIZE sizeof(arg_list) / sizeof(char*)
+
+extern char **environ;
+int main(int argc, char *argv[]) {
+
+    // Relative path from this file.
+    char python_path[] = "/../Frameworks/pythonexecutable";
+
+    // Get this file's path.
+    char path[1048];
```
benoit-pierre, line 16:
```suggestion
    char path[PATH_MAX];
```
N.B.: `limits.h` needs to be included.

osx/app_resources/plover_launcher.c
```diff
@@ -0,0 +1,41 @@
...
+    uint32_t size = sizeof(path);
+    _NSGetExecutablePath(path, &size);
+
+    // Remove filename from path.
+    char * last_slash;
+    last_slash = strrchr(path, '/');
+    path[last_slash - path] = '\0';
```
benoit-pierre, line 23: Why not just use `dirname`?

benoit-pierre (commented):
I don't know about `_NSGetExecutablePath`, I'm having a hard time finding official Apple documentation about the function, I'd just use `realpath` with `argv[0]`.

Anyway, here's my (untested) simplified version of the launcher code:

```C
#include <libgen.h>
#include <limits.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

int main(int argc, char *argv[])
{
    char python_path[PATH_MAX];
    char *new_argv[4 + argc];
    char *app_dir;

    // Get this app bundle directory.
    app_dir = realpath(argv[0], NULL);
    app_dir = dirname(dirname(app_dir));

    // Get path to the Python interpreter.
    snprintf(python_path, sizeof (python_path),
             "%s/Frameworks/pythonexecutable",
             app_dir);

    // Build new arguments list, forwarding original arguments.
    new_argv[0] = python_path;
    new_argv[1] = "-s";
    new_argv[2] = "-m";
    new_argv[3] = "plover.dist_main";
    memcpy(new_argv + 4, argv + 1, argc * sizeof (argv[0]));

    return execv(new_argv[0], new_argv);
}
```

morinted:
Tested out the rewrite in .app form and when running with arguments -- working for me.

> That's convoluted, why not use simple assignments + memcpy?

I'd say the amount of c I've written wouldn't fill a pamphlet. I learned a lot from your review -- thank you!

PR #1153, opened 2020-10-18
Fix output capitalization issue on X

### LINKED ISSUE #1029: Case is not respected for all Unicode characters
## Summary

When writing a word starting with some Unicode characters, the word will be aggressively lower-cased.

## Reproducing

1) add a dictionary entry for an upper-case non-ASCII Unicode character, e.g. `Æ`:
```json
{ "5E": "Æ" }
```
2) use Plover to enter the stroke in a lower-case context.  See the character appear in lower-case (e.g. `æ`).
3) use Plover to enter the stroke in an upper-case context.  See the character appear in lower-case again.

## Plover Version

Plover 4.0.0.dev8

## System

NixOS, installed from Nix derivation.  XMonad, no DE, &c.

### LINKED ISSUE #1057: The letter Å sometimes can't be capitalized
## Summary

The letter Å sometimes can't be output capitalized. It only happens sometimes, and it seems to be some weird state Plover might get into at startup. Several actions, such as scrolling the suggestions window up and down might get Plover out of that state, and then the bug won't happen, at least until Plover is restarted.

This only seems to happen with the letter Å, but I wonder if it's related to this: https://github.com/openstenoproject/plover/issues/1029

## Reproducing

Load a dictionary with entries like these:

` ` `
"A*UB": "{&å}",
"A*UP": "{&Å}"
` ` `

Then try to use the entry that writes Å. It should always write Å, but sometimes it comes out as å. If you use the stroke several times it'll keep outputting what it first wrote. So if it wrote å, it'll keep doing that if you try again. But if you then use the other stroke to write å, it will change something so the next time you try to write Å it'll come out correctly.

## Plover Version and System

Running from source on Ubuntu 16.04

user202729:
## Summary of changes

Uses `VoidSymbol` instead of `NoSymbol`, so the special case (explained in [this comment](https://github.com/jordansissel/xdotool/issues/97#issuecomment-711143601)) won't kick in.

`VoidSymbol` isn't defined in Xlib, so I have to hard code the constant.

Closes #1057, closes #1029 .

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1153.linux.rst
+++ b/news.d/bugfix/1153.linux.rst
@@ -1,0 +1,1 @@
+Fix output capitalization issue.
--- a/plover/oslayer/xkeyboardcontrol.py
+++ b/plover/oslayer/xkeyboardcontrol.py
@@ -39,7 +39,7 @@ from plover import log
 XK.load_keysym_group('xf86')
 # Load non-us keyboard related keysyms.
 XK.load_keysym_group('xkb')
-   
+
 # Create case insensitive mapping of keyname to keysym.
 KEY_TO_KEYSYM = {}
 for symbol in sorted(dir(XK)): # Sorted so XK_a is preferred over XK_A.
@@ -1137,6 +1137,8 @@ class KeyboardEmulation:
     CUSTOM_MAPPING_LENGTH = 3
     # Special keysym to mark custom keyboard mappings.
     PLOVER_MAPPING_KEYSYM = 0x01ffffff
+    # Free unused keysym.
+    UNUSED_KEYSYM = 0xffffff # XK_VoidSymbol
 
     def __init__(self):
         """Prepare to emulate keyboard events."""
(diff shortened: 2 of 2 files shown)


morinted:
- We'll need a news fragment for this fix
- I'd like @benoit-pierre to approve as he's the most knowledgeable about all the Linux code

user202729:
I wouldn't consider myself best for writing these snippets...

Something not related to this PR:

* Somehow using a list for a queue feels inefficient. (this might be insignificant, because Python is already somewhat-inefficient anyway and there's only about 1000 custom mappings)
* I don't really like using `.index()` to determine the `keysym_index` in the `_get_mapping()` function; storing the `keysym_index` in the `Mapping` struct is better -- because there may be multiple equal indices, and the current code just happens to work because the queue is initially in increasing-modifier order. See https://github.com/user202729/plover/commit/b5aa8ee5f2823b8c80b89beeb5a704426b5367be.

benoit-pierre:
Nice work, I've tweaked the code slightly (minor cosmetic changes).

user202729:
Just noticed that the news snippet is still in rst format...

benoit-pierre:
I noticed too, but I checked, and towncrier does not mind.

Issue #987, opened 2018-08-01
support for Melani system


benoit-pierre:
2 changes are needed for fulling supporting a Melani system plugin:

- [x] support for prefix strokes (only match at the beginning of a new word)
- [x] formatting look ahead support (for conditional output)

SeaLiteral:
What does "support for prefix strokes" mean? Does it mean tucked prefixes as described in #974 or does it mean applying orthography rules when adding prefixes? It seems orthography.py has an add_suffix function but no add_prefix function, but I don't know if Italian would need that (Spanish does have a rule that if a word starts with an r and you add a prefix that ends with a vowel, you have to double that r.

benoit-pierre:
Prefix strokes only match at the beginning of a word: this work with Melani because words are explicitly terminated (when using one of the terminating vowels). So all output from the orthographic Python dictionary is basically prefixes, unless ending with one of those terminating vowels.

Example blackbox test:
```python
def test_prefix_strokes(self):
    r'''
    "/S": "{prefix^}",
    "S": "{^suffix}",
    "O": "{O'^}{$}",

    S/S/O/S/S  " prefixsuffix O'prefixsuffix"
    '''
``` 
Note: `{$}` is new formatting syntax to explicitly mark the end of a word

benoit-pierre:
And some example blackbox tests for look ahead support):
```python
def test_conditionals_1(self):
    r'''
    "*": "=undo",
    "S-": "{=(?i)t/true/false}",
    "TP-": "FALSE",
    "T-": "TRUE",

    S-   ' false'
    TP-  ' false FALSE'
    *    ' false'
    T-   ' true TRUE'
    *    ' false'
    S-   ' false false'
    TP-  ' false false FALSE'
    *    ' false false'
    T-   ' true true TRUE'
    '''
 
def test_conditionals_2(self):
    r'''
    "1": "{=(?i)([8aeiouxy]|11|dei|gn|ps|s[bcdfglmnpqrtv]|z)/agli/ai}",
    "2": "oc{^}chi",
    "3": "dei",
    "4": "sti{^}vali",

    1  ' ai'
    2  ' agli occhi'
    1  ' agli occhi ai'
    3  ' agli occhi agli dei'
    1  ' agli occhi agli dei ai'
    4  ' agli occhi agli dei agli stivali'
    '''
```

SeaLiteral:
[...]

PR #351, opened 2016-02-12
Add CONTRIBUTING.md [fixes #346]

jeremy-w:
Also makes progress towards #261.

PR #1158, opened 2020-10-23
formatting: look ahead support

benoit-pierre:
#351  Summary of changes

Another change needed for full Melani support: conditional output, based on the text following a translation. The new syntax is `{=REGEXP/TRANSLATION_IF_FOLLOWING_TEXT_MATCH/TRANSLATION_IF_NOT}`. See blackbox tests for examples.

Needed for #987.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/plover/meta/conditional.py
+++ b/plover/meta/conditional.py
@@ -1,0 +1,20 @@
+import re
+
+from plover.formatting import _LookAheadAction
+
+
+IF_NEXT_META_RX = re.compile(r'((?:[^\\/]|\\\\|\\/)*)/?')
+IF_NEXT_ESCAPE_RX = re.compile(r'\\([\\/])')
+
+
+def meta_if_next_matches(ctx, meta):
+    pattern, result1, result2 = [
+        IF_NEXT_ESCAPE_RX.sub(r'\1', s)
+        for s in filter(None, IF_NEXT_META_RX.split(meta, 2))
+    ]
+    action_list = []
+    for alternative in result1, result2:
+        action = ctx.new_action()
+        action.text = alternative
+        action_list.append(action)
+    return _LookAheadAction(pattern, *action_list)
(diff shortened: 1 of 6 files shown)


plover/meta/conditional.py
```diff
@@ -0,0 +1,10 @@
+from plover.formatting import _LookAheadAction
+
+def meta_if_next_matches(ctx, meta):
+    pattern, result1, result2 = meta.split('/')
+    action_list = []
+    for alternative in result1, result2:
+        action = ctx.new_action()
+        action.text = alternative
```
user202729, line 8: Although not really needed for Melani system, this is some suggestions:

* perhaps the alternative should be interpreted as a Plover stroke? so people can do something like `{=pattern1/alternative1/{=pattern2/alternative2/fallback}}`
* or just `{=pattern1/alternative1/pattern2/alternative2/fallback}`. Currently it doesn't do anything useful when given more than 2 slashes anyway.
* currently there's no way to include a slash in the pattern or the replacement.

benoit-pierre: > * perhaps the alternative should be interpreted as a Plover stroke? so people can do something like `{=pattern1/alternative1/{=pattern2/alternative2/fallback}}`
> * or just `{=pattern1/alternative1/pattern2/alternative2/fallback}`. Currently it doesn't do anything useful when given more than 2 slashes anyway.

I really don't want to increase the complexity.

> * currently there's no way to include a slash in the pattern or the replacement.

I've amended the code to support that use case.

> * or should this be included in the orthography rules regex instead? The user might want to define multiple stroke to output the same prefix, and there's cumbersome.

Orthography rules are really limited, they're only used when prefixing a word (`{^post}`), and an initial prefix (`{pre^}`) is not taken into account.

morinted (approved):
Looks good to me

PR #1160, opened 2020-10-24
Allow higher priority dictionaries to delete entries from others

### LINKED ISSUE #726: Add definition deletions to user.json
Feature request:

I want to be able to add something to my user.json that has the same effect as removing that stroke from main.json, but without modifying main.json at all.

For single-stroke briefs, it's usually easy to just add the definition I want, but for multi-stroke briefs, when I want the strokes to be looked up separately, the only good way to do this currently is to delete from main.json

This leads to two problems:

First, for individuals, this makes it hard to upgrade to the latest main.json from the latest version of plover (because it'll re-add all the definitions you deleted.)

And Second, as a consequence, there is much less participation in improving main.json, because most experienced stenos have long ago forked main.json, and have no decent way of using the latest main.json. It's hard to be motivated to improve Free Software stuff that we don't use personally.

I think the default main.json is a critical part of the value provided by Plover, and its quality will determine whether many people stick with steno or not.

Also, those of who stay spend many duplicated hours of work fixing our local main.json files.

For these reasons, I think it is important that we can make all our dictionaly modifications without editing the main.json file.

A secondary benefit of this feature is that dictionaries that we share can be more powerful. For example a dictionary for using ``SKP[...]`` for "ex[...]" words would fix several word conflicts, and with this feature could remove the multi-stroke definitions that are no longer needed, thus reducing colateral damage to other sequences.

Example: I've trained myself to type "cannot" with ``KANLT`` so I expect to get "can not" from ``K/NOT``. options:

1. I can delete ``K/NOT`` from main.json, but that makes it hard for me to upgrade main.json later

2. I can add ``K/NOT`` → "can not" but sometimes doing this creates conflicts (or fails to get rid of them.) It's probably fine in this case, but if I had a better example, you'd see how I would want the first stroke to be interpreted as the final stroke of the previous definition where appropriate.

I want a third option that is:

3. I can add something to my user.jason (or any dictionary) that makes plover ignore that key in all lower-priority dictionaries.

I'm not real particular about how it's implemented, but here some ideas:

``"K/NOT", null,``

``"K/NOT", "{PLOVER:DELETED}",``

The advantage of the latter is that it can be pasted into the DUPT dialog box, offering a solution with no UI changes required.

If we want to get fancy, we could add a new dialog box for deleting definitions in this way. It would be very much like the DUPT dialog box, but without the translation stuff (but including the pull-down for which dictionary.)

P.S. One more benefit: with deletions in user.json, this makes it easier to batch-submit all your dictionary changes to issue 400 by just uploading your whole user.json, then starting a new dictionary (so next time you submit you don't resend all the stuff you sent before again.)

user202729:
## Summary of changes

If there's an entry `{plover:deleted}` in a dictionary, Plover will understand that the entry is deleted.

TODO: currently reverse-lookup still lists the deleted entry.

I've seen the POC in the linked issue, but it has a lot of merge conflicts, it's more complex (edits a lot of places in the code to special-case handle it), it doesn't work well with the add translation dialog (by default it does nothing if either field is empty) and it doesn't seem to work.

Closes #726 .

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/steno_dictionary.py
+++ b/plover/steno_dictionary.py
@@ -210,8 +210,10 @@ class StenoDictionaryCollection:
             if key_len > d.longest_key:
                 continue
             value = d.get(key)
-            if value:
+            if value is not None:
                 if not any(f(key, value) for f in filters):
+                    if value.lower() == "{plover:deleted}":
+                        return None
                     return value
 
     def _lookup_from_all(self, key, dicts=None, filters=()):


mkrnr:
Hey @user202729, I really like this addition! 
Do you have time to revive this one? Otherwise, I can work on it.

mkrnr (approved):

PR #1161, opened 2020-10-24
Sticky metakeys with :down and :up syntax

user202729:
## Summary of changes

`{#key:up}` and `{#key:down}` can be used to hold the key.

I use `:` instead of `_` because of the obvious conflict with `page_up` and `page_down`.

The obvious issue is that user might accidentally press a sticky key and don't know what was pressed or how to release it. However, if the user knows some Python, they can make a Python dictionary -- there are a few options:

* Shortcut on the fly (control+`=` for example): I don't think this is necessary. Just define separate entries for every key on the keyboard. There isn't that much keys, and a program can be used to automate the generation.
* Alt+tab: the user can make a Python dictionary to map stroke `X` to alt:down tab, and `X/Y` to alt:up for all stroke `Y` not equal to `X`. (so the user can repeat alt+tab)
* Control+drag mouse: similarly, map `X` to control:down, and `X/Y` to control:up for all stroke `Y`

These are complex use cases, and I don't think having a built-in for all of these is a good idea. Perhaps an additional dictionary format (JSON with regex/wildcard matching)?

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/key_combo.py
+++ b/plover/key_combo.py
@@ -135,7 +135,7 @@ CHAR_TO_KEYNAME = {
 }
 
 
-_SPLIT_RX = re.compile(r'(\s+|(?:\w+(?:\s*\()?)|.)')
+_SPLIT_RX = re.compile(r'(\s+|(?:\w+(?:\s*(?:\(|:up|:down))?)|.)')
 
 def parse_key_combo(combo_string, key_name_to_key_code=None):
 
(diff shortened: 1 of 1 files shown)


JoshuaGrams:
What about a `{#all:up}` option? That would solve the problem with not knowing which sticky keys are down.

user202729:
That would work too, but if the user only realize that some key is held after pressing some chord, then it's still problematic. Also it takes a nontrivial change to make Plover stateful.

As a workaround, it's possible to do `{#control_l:up shift_l:up alt_l:up <list all keys here>...}` -- release an already-up key is not a problem.

(by the way, `{#alt:down alt(c)}` will release the alt key.)

Another possibility is to show some message, like windows's stickykey.

morinted:
@benoit-pierre took a stab at this problem with the [locking-keys branch here.](https://github.com/benoit-pierre/plover/tree/locking_keys)

One of the ways to prevent disaster was to have a rule where only keys marked with `+` and `-` in keyboard shortcuts would be allowed.

Hence, if you wrote `{#+command}`, command would be held (you could command-click, for example) and then wrote the word `duck`, it would release command and then write duck.

If you wanted to, say, command tab and tab through, you'd send `{#+command}` and then a stroke that's `{##tab}` (the double-hash used to mean that it's okay to keep modifiers held for this stroke)

PR #1163, opened 2020-10-27
Lock the display on use in KeyboardCapture for X

### LINKED ISSUE #1126: Toggle-related crashes on Linux
## Summary

Plover freezes or becomes unresponsive after toggling it on/off a couple times.

## Reproducing

I recently discovered the PHROLG stroke to toggle Plover between the enabled and disabled states without using the mouse. Unfortunately after toggling back and forth a number of times(varies a bit) Plover will either stop translating any inputs regardless if it's enabled or not and become unresponsive(e.g. clicking "quit plover" does nothing), or it completely freezes up with a similar outcome(UI freezes up completely and system pops up a dialog offering to kill the program). In both cases it can only be stopped by manually killing the process.

Toggling by clicking the switches with the mouse does not seem to cause this effect.

## Plover Version

4.0.0.dev8+66.g685bd33 (AppImage)

## System

Linux Mint 19.3 (Mate desktop)

user202729:
## Summary of changes

Use a lock so that only one thread use the `_display` object of `KeyboardCapture` (in `plover/oslayer/xkeyboardcontrol.py`).

(there's no Python-xlib documentation about thread-safety (in fact there's hardly any documentation at all), so it's safer to assume that it isn't.)

Should close #1126 . (I toggle Plover with keyboard input machine repeatedly, and it tends to freeze without this change, and doesn't freeze so far with this change)

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/oslayer/xkeyboardcontrol.py
+++ b/plover/oslayer/xkeyboardcontrol.py
@@ -21,6 +21,7 @@ http://tronche.com/gui/x/xlib/input/keyboard-encoding.html
 
 """
 
+from functools import wraps
 import errno
 import os
 import string
@@ -143,6 +144,19 @@ KEYCODE_TO_KEY = {
 
 KEY_TO_KEYCODE = dict(zip(KEYCODE_TO_KEY.values(), KEYCODE_TO_KEY.keys()))
 
+def with_display_lock(func):
+    """
+    Use this function as a decorator on a method of the XEventLoop class (or
+    one of its subclasses) to acquire the _display_lock of the object before
+    calling the function and release it afterwards.
+    """
+    # To keep __doc__/__name__ attributes of the initial function.
+    @wraps(func)
+    def wrapped(self, *args, **kwargs):
+        with self._display_lock:
+            return func(self, *args, **kwargs)
+    return wrapped
+
 class XEventLoop(threading.Thread):
 
     def __init__(self, name='xev'):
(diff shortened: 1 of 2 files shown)


user202729:
I use a similar pattern to `engine.with_lock` to implement the decorator, but I think accessing the lock directly is better than defining `__enter__` and `__exit__`.

user202729:
It's also possible to use only one global `_display` object (currently there are two for `KeyboardCapture` and `KeyboardEmulation`), or make a separate one for the event loop thread in `XEventLoop` (so there would be 3 of them)

user202729:
News fragment is added, but I'm still not confident about letting the contributor to write it.

I have no idea how to test this with the current testing utils.

benoit-pierre:
Good job. I refactored the code to make it more readable and avoid the need to take the lock twice for each event. @user202729 : can you retest on your side?

plover/oslayer/xkeyboardcontrol.py
```diff
@@ -165,39 +165,38 @@ def __init__(self, name='xev'):
...
         self._pipe = os.pipe()
         self._display_lock = threading.Lock()
+        self._readfds = (self._pipe[0], self._display.fileno())
 
     def _on_event(self, event):
         pass
 
+    @with_display_lock
+    def _process_pending_events(self):
+        for __ in range(self._display.pending_events()):
```
user202729, line 175: The double underscore looks weird...?

benoit-pierre: Yes, usually `_` is used to indicate a throw away variable, but it's also the default global shortcut for the gettext builtin, so...

user202729:
The functionality does work (as in it doesn't freeze)

Issue #1030, opened 2018-12-07
Ubuntu 18.04 output delay


CharleyShattuck:
## Summary

Using keyboard protocol in Ubuntu 18.04 there is always a noticeable lag between releasing all keys and seeing output on the screen. The paper tape display is equally delayed. I had no such problem with Ubuntu 16.04, but as soon as I upgraded to 18.04 the delay was there. I'm not always fast you enough to see this, but if I write continuously and fast Plover doesn't output anything, not until I pause for a moment. Then everything I just wrote appears all at once.

## Reproducing

Nothing special, just use keyboard protocol in Ubuntu 18.04.

Connected by a USB micro cable.

## Plover Version

Plover 4.0.0.dev8+66.g685bd33

Plover 4.0.0.dev8+66.g685bd33

## System

Linux, specifically Ubuntu 18.04,appImage, not installed. I click on the appImage on the desktop to start Plover.

benoit-pierre:
After investigation, it's a gnome-shell bug: for some reason, the keymap layout changes (generated when using Xtest to simulate a keyboard event and when a key on the real keyboard is pressed) result in gnome-shell doing something to the X server that delays all other events for a little while (and X server high cpu usage). This can be reproduce by using xdotool to simulate input and typing on the keyboard at the same time. This does not happen with other window managers (e.g. switching back to Unity). It's also harder for me to reproduce on my real machine (as opposed to with a Virtualvox VM install of Ubuntu Bionic), whether because the machine is faster or the gnome stack is more recent, I don't know. In any case, I don't see anything actionable on Plover' side.

mayanksuman:
My configuration is a bit different.
OS: Debian Linux ((kernel version: 4.19.0-2-amd64)
WM: Gnome (3.30.2)

However, plover (commit 2943a97) is also showing lag in the input. I observed that upon fast typing `Xorg` process start to consume full cpu till all the words appear on the screen.

I tried to use plover on wayland too, but it is not working there (issue #1050). Hence, combined with this issue the plover is having problem with Gnome WM on both mode (`Xorg` or `Wayland`).

zapakh:
I'm running XFCE4 (xubuntu 20.04), no Gnome in sight, and having the same problem. It's not just output that slows down. Video and anything else that's running tends to stutter, too. Running `xev -event keyboard` shows a lot of events like

```
MappingNotify event, serial 285, synthetic NO, window 0x0,
request MappingKeyboard, first_keycode 8, count 248
[...]

PR #1164, opened 2020-10-27
Update keymap on keymap change

### LINKED ISSUE #967: ( ͡° ͜ʖ ͡°) sometimes written as ( °  °)
## Summary

I have a stroke for the so-called Lenny face, `( ͡° ͜ʖ ͡°)` defined as the stroke L-N or HR-PB. In some sessions, when I write that stroke, this face is printed out: `( °  °)`. The characters in the middle have been converted to space characters (I inspected them in Emacs to find this out).

## Reproducing

Here is a stroke log of the stroke that fails:
```
2018-06-05 22:02:12,415 Stroke(HR-PB : ['H-', 'R-', '-P', '-B'])
2018-06-05 22:02:12,426 Translation(('HR-PB',) : "( ͡° ͜ʖ ͡°)")
```
It's unclear to me how exactly to reproduce this. I open Plover one day, then do my work and possibly put the computer to sleep for the night. At some point I find that the ( ͡° ͜ʖ ͡°) face is broken.

I can always fix the stroke by restarting Plover.

## Plover Version

I'm running Plover from source, today I was testing https://github.com/openstenoproject/plover/pull/964
The sha is 2fbd465 .

## System

Fedora Linux 24
Linux QFINL014 4.11.12-100.fc24.x86_64 #1 SMP Fri Jul 21 17:35:20 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

I know this isn't much to go by, but if you have any ideas on what I should inspect or capture or tinker, I should be able to make time during my day.

user202729:
## Summary of changes

Listens for `XMappingEvent`s and call `_update_keymap` whenever the keymap changes.

Closes #967.

Relies on the other pull request for `self._display_lock`.

Problem: for some reason, it refreshes the key map every time I use a different keyboard (in case I have more than one keyboard plugged in the machine -- which is usually the case with the built-in keyboard and a steno keyboard)

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1163.linux.rst
+++ b/news.d/bugfix/1163.linux.rst
@@ -1,0 +1,1 @@
+Fix a race condition that may freeze Plover while toggling with keyboard input machine.
--- a/plover/oslayer/xkeyboardcontrol.py
+++ b/plover/oslayer/xkeyboardcontrol.py
@@ -21,6 +21,7 @@ http://tronche.com/gui/x/xlib/input/keyboard-encoding.html
 
 """
 
+from functools import wraps
 import errno
 import os
 import string
@@ -143,6 +144,19 @@ KEYCODE_TO_KEY = {
 
 KEY_TO_KEYCODE = dict(zip(KEYCODE_TO_KEY.values(), KEYCODE_TO_KEY.keys()))
 
+def with_display_lock(func):
+    """
+    Use this function as a decorator on a method of the XEventLoop class (or
+    one of its subclasses) to acquire the _display_lock of the object before
+    calling the function and release it afterwards.
+    """
+    # To keep __doc__/__name__ attributes of the initial function.
+    @wraps(func)
+    def wrapped(self, *args, **kwargs):
+        with self._display_lock:
+            return func(self, *args, **kwargs)
+    return wrapped
+
 class XEventLoop(threading.Thread):
 
     def __init__(self, name='xev'):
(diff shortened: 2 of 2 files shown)


benoit-pierre:
Sorry for the delay, will review later (maybe after 4.0.0 is released): when I investigated updating the layout on keymap change, I notice some issues due to the fact that the code does not handle XKB, which meant in my case I would get a lot of keymap changes because of the switch between the events sent by my physical keyboard, and the XTest keyboard (which use different layouts). I realize that my use case might be special (multiple physical keyboards with different layouts), so maybe that's not a valid reason for not updating on change. The alternative might be proper XKB support, but `python-xlib` does not support that (and adding support would be a lot of work).

user202729:
About xkb... I think that it is also the cause of https://github.com/openstenoproject/plover/issues/1189 so to fix that without xkb support this one is necessary. 

Xref: Switching keyboards causes issues in other programs not supporting xkb too (https://github.com/openstenoproject/plover/issues/1030 )

user202729:
Rebased to latest master.

Remarks:

* Now (after the rebase) the KeyboardCapture object contains 2 display objects, one inside the event loop and one for (sending events|update keymap). The latter one needs a separate lock (just in case), which results in the code being duplicated from XEventLoop to KeyboardEmulation.

  (it may be a better idea to make 3 display objects instead so one lock is avoided... but this crucially relies on XEventLoop's implementation to all `on_event` functions sequentially...)

* The event loop is started, but never canceled. (as before the rebase)

* Now I see Benoit's pointed out problem (not sure what changed, maybe something xkb something?) that the keymap is refreshed on every stroke. Not good.

* Crucially, it may happen that the keyboard mapping is different for different keyboard, and **Plover sets the mapping for the real keyboard but send events using the xtest keyboard** which leads to the problem being "unfixably broken" even after a Plover restart.

  In this case it would be necessary to exit Plover, `setxkbmap us`, reopen Plover.

----

in any case, some investigation is needed. pull request is not ready.

PR #1166, opened 2020-11-04
Replace `popcount_8` with precalculated table

mcognetta:
## Summary of changes

Replace `popcount_8` logic with a precomputed table of popcount values.

This reduces code complexity and has a slight performance benefit (though it is not clear how often the places that popcount is called are run), around 3x on two machines that I tested.

### Pull Request Checklist
- [x] Changes have tests
  - tests already exist


morinted:
Thanks for this, I think I'd prefer memoizing/caching the value over the precalculated table, as regardless of which method we use, it's still pretty arbitrary looking (the giant tuple to just as perplexing to me as the magic function).

It would probably be worth finding a better-annotated function for generating popcount or at least linking to a source in case our requirements here ever change.

user202729:
I also don't think that this is an improvement, since that part is not the bottleneck; however instead of the hard coded constant a list comprehension [bin(x).count('1') for...]  or something similarly simple would probably achieve the same performance (with a tiny startup overhead).

mcognetta:
Ah I forgot about this. Yeah, if it's not hot it's probably not much gain.

Thanks for the comments. I'll close this one.

PR #1170, opened 2020-11-10
switch distributions to Python 3.7

benoit-pierre:
## Summary of changes

Updating the Linux and Windows versions was easy, but I ran into a snag with the macOS version: after the upgrade to Python 3.7, `python -m macholib standalone` does not create a working standalone version anymore; the system prefixes are not correctly set (still pointing to the original source version of Python), tripping `plover_build_utils.check_requirements`.

I ended up switching to [this script](https://github.com/gregneagle/relocatable-python) for creating a relocatable Python directly from one of the official installer package.

Hopefully someone with access to a mac can check the final result does indeed work.

### Pull Request Checklist
- [x] News fragment added in news.d.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ matrix:
     # Linux / Python 3
     - os: linux
       language: python
-      python: "3.6"
+      python: "3.7"
       dist: xenial
       sudo: required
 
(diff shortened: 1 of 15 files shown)


benoit-pierre:
Closing this:
* the keyboard machine does not work on Windows (see #1198), and the user installed plugins are not loaded (because the user site is disabled)
* but more importantly, while Python 3.7 still receives security fixes, those new releases are source only, so I think we might as well switch to 3.8

PR #1174, opened 2020-11-27
[WIP] Add developer-facing documentation (user-facing docs to follow)

sammdot:
## Summary of changes

This pull request adds more developer-oriented documentation to the `doc/` directory, primarily consisting of plugin development docs (based on [the plugins wiki page](https://github.com/openstenoproject/plover/wiki/Plugins)) and an API reference (not currently documented). I believe the ideal place for documentation is adjacent to the code it's for, but I'm not sure how y'all feel about docstrings and type annotations directly in the code so I think this is the next best thing.

I've set the repository up for Read the Docs (https://readthedocs.org), which provides documentation hosting free of charge for a public repository like this one. The documentation on that site is updated for every commit, so there's no need to worry as much about keeping it updated.

My WIP branch is currently hosted at https://plover.readthedocs.io/ using my personal account, though of course I'd be happy to hand over ownership of the RTD project when this gets merged. Eventually I'd like for it to be hosted at something like https://docs.ploversteno.org/.

In the future, I'd like for this site to contain much of the user-facing documentation that is currently in the wiki as well -- installation instructions, hardware options and setup guides -- but my priority right now is the developer docs since that is currently not as complete.

## Pull request checklist

- [x] Changes have tests:
    - Docs are hosted at https://plover.readthedocs.io/
    - I'm not sure how to automate testing but here's a build log from RTD: https://readthedocs.org/api/v2/build/12432624.txt
- [ ] News fragment added in news.d:
    - What should I call it, exactly? None of the categories mentioned in the developer guide seem to fit


user202729:
(I'm still maintaining my fork of the documentation at https://plover2.readthedocs.io, although it's still largely incompatible with this version because of autodoc.)

user202729:
Parallelly, some typing stub files (for mypy type checking for example) would also be useful.

doc/plugin-dev/machines.md
```diff
@@ -0,0 +1,94 @@
...
+See {doc}`../api/machine` for more information.
+
+## Machine Options
+
+If your machine requires additional configuration options, add a machine
+options entry point:
+
+```ini
+[options.entry_points]
+plover.gui_qt.machine_options =
```
tadeokondrak, line 72: I was following this guide, and found that I needed to use this instead:
```suggestion
plover.gui.qt.machine_option =
```

PR #1179, opened 2020-12-18
pin pip's version

### LINKED ISSUE #1177: pip 20.3 release breaks source install
## Summary

Building Plover from source causes conflicts in pip's package resolution algorithm. See #1176 for an example of what the error looks like -- it seems pip takes the duplicate installation of local Plover and PyPI Plover as a conflict rather than automatically resolving to the local version (as was the behavior in older versions of pip).

[pip documentation](https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020) says this about the new package resolver:

> * Providing an editable requirement (`-e .`) does not cause pip to ignore version specifiers or constraints (see pypa/pip#8076), and if you have a conflict between a pinned requirement and a local directory then pip will indicate that it cannot find a version satisfying both (see pypa/pip#8307)

While Plover's requirements.txt specifies `pip==20.0.2`, it seems the build process forces an upgrade to 20.3, which is the first version that introduces the new resolution behavior.

## Reproducing

* Clone Plover repository at 4.0.0.dev8+235 (`d20167f`)

* Run the typical development build commands:
  ```
  python3 -m pip install -r requirements.txt
  python3 -m pip install --user -e . -r requirements_plugins.txt
  ```

* Build a macOS app bundle:
  ```
  python3 ./setup.py bdist_app
  ```

The build fails with the following error message:

```
ERROR: Cannot install -r requirements_plugins.txt (line 19), plover 4.0.0.dev8 (from ~/plover/dist/plover-4.0.0.dev8-py3-none-any.whl) and plover[gui-qt]==4.0.0.dev8 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested plover 4.0.0.dev8 (from ~/plover/dist/plover-4.0.0.dev8-py3-none-any.whl)
    plover-treal 1.0.1 depends on plover>=4.0.0.dev0
    plover[gui-qt] 4.0.0.dev8 depends on plover 4.0.0.dev8 (from https://files.pythonhosted.org/packages/96/09/31e7455012feb2022bb5404308c5c6026ab192840927a46471c95903c31e/plover-4.0.0.dev8-py3-none-any.whl#sha256=c9f42573acd3729b6750d3c7644f81086512ae6ed831dea1440d1f113daa4225 (from https://pypi.org/simple/plover/))

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

## Plover Version

4.0.0.dev8+235 (`d20167f`)

## System

macOS 11.0 Big Sur, Python 3.8.5

benoit-pierre:
Improve builds reproducibility and avoid breakage with pip 20.3.

Closes #1177.

### DIFF
--- a/linux/appimage/build.sh
+++ b/linux/appimage/build.sh
@@ -71,8 +71,8 @@ do
   run extract_appimage "$cachedir/$tool" "$builddir/$tool"
   run_eval "$tool='$builddir/$tool/AppRun'"
 done <<\EOF
-zsync2        https://github.com/AppImage/zsync2/releases/download/continuous/zsync2-156-10e85c0-x86_64.AppImage
-zsync2        https://github.com/AppImage/zsync2/releases/download/continuous/zsync2-156-10e85c0-x86_64.AppImage
+zsync2        https://github.com/AppImage/zsync2/releases/download/continuous/zsync2-15-86cfd3a-x86_64.AppImage
+zsync2        https://github.com/AppImage/zsync2/releases/download/continuous/zsync2-15-86cfd3a-x86_64.AppImage
 linuxdeploy   https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage
 appimagetool  https://github.com/probonopd/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage
 EOF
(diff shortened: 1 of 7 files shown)


benoit-pierre:
Updated with a workaround for pip's new dumb resolver and a fix for the AppImage creation. I'm going to go ahead and merge so we can get the CIs passed again.

Issue #645, opened 2016-11-15
Feature Request: command to show/hide suggestions window


JasonWoof:
Classification: Feature (New)

## Summary

Plover command to toggle visibility of suggestions window.

## Use description/case

1. I map a stroke to {PLOVER:SUGGESTIONS_WINDOW} or some such.
2. I stroke said stroke and the suggestions window opens.
3. I stroke it again later and the suggestions window closes.

morinted:
I think that each GUI tool should optionally be able to define a command name for bringing to front. Thoughts, @benoit-pierre ?

Though, I would make it so that {PLOVER:SUGGESTIONS_WINDOW} only raises the window → to close it you would raise/focus it, then press escape (this is how add translation and look up currently work)

JasonWoof:
"raising" doesn't do anything on my system. I use a tiled window manager. I don't see the point of a window being "open" if you can't see it. From a user perspective, I can see it or I can't. Hiding it behind other windows is no different to me than it being closed.

I've mostly stopped using the suggestion window because I got sick of moving it around with the mouse.

I just want to be able to assign a stroke to show suggestions, and a stroke to hide it (could be the same stroke again, or one for closing windows)

morinted:
Is it not true that "raising" the window would focus it so that you could close it with `{#Escape}`?

JasonWoof:
No, raising is ignored. The "raising" approach might have problems on other systems/setups when the suggestions window is open, but not in the current workspace. I've been leaving mine open in a different workspace.

PR #1184, opened 2021-02-14
Add command to show suggestions window

antistic:
## Summary of changes

Add `{PLOVER:SUGGESTIONS}` to show the suggestions window in the same way as `{PLOVER:LOOKUP}` does with the lookup window.

(partially addresses) #645.

To fully address the issue we could add a `{PLOVER:SUGGESTIONS_TOGGLE}` or `{PLOVER:SUGGESTIONS_CLOSE}` (which wouldn't be too hard), but then we would likely also want it for other windows e.g. `{PLOVER:LOOKUP_TOGGLE}`. It may be better to instead address why `{PLOVER:SUGGESTIONS}{#Escape}` doesn't work on all DE/WMs.

### Pull Request Checklist
- [ ] Changes have tests :arrow_right: happy to add these if given pointers as to what kind of test
- [ ] News fragment added in news.d. See [documentation](https://github.com/openstenoproject/plover/blob/master/doc/developer_guide.rst#making-a-pull-request) for details :arrow_right: Not sure what category this falls under so I haven't done this yet


user202729:
Likely because the suggestion window appears too quickly. If there's something like `{#sleep 2000}` (this is another issue...) that could fix the issue.

Or the issue could be that they're on different windows/workspaces.

There's also a related pull request, that adds a delay between every key presses, which would probably also fix the issue.

Currently there's no test for the GUI related stuff. (that having said, I'm not a official Plover maintainer)

morinted (approved):

morinted (approved):

user202729:
Remark: apparently this breaks some plugin (in particular, plover_engine_server), which relies on the list of all hooks.

On the other hand, this can be considered just a minor change, because all programs will break if they rely on the list of all functions provided by an API to work correctly.

PR #1188, opened 2021-03-03
Fix memory leak caused by defaultdict access in suggestions feature

user202729:
## Summary of changes

Fixes a memory leak in the suggestions feature.

The leak only become apparent to me when I use the new console GUI. (normally I never use the suggestions window)

Remark: when I use the console GUI, I occasionally get a hang too. I guess it's caused by either dict rehash or garbage collection.

### Pull Request Checklist
- [ ] Changes have tests -> Will have later. Idea: count the number of items in the dicts after an nonexistent reverse lookup.
- [ ] News fragment added in news.d. See [documentation](https://github.com/openstenoproject/plover/blob/master/doc/developer_guide.rst#making-a-pull-request) for details -> TODO

### DIFF
--- a/news.d/bugfix/1188.core.md
+++ b/news.d/bugfix/1188.core.md
@@ -1,0 +1,1 @@
+Fix unbounded memory use in the lookup functions used by the Suggestions window.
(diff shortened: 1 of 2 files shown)


morinted (approved):

benoit-pierre:
Nice catch, technically not a memory leak, but a case of unbounded memory use when using the suggestions. Can you add a news entry?

morinted:
I've added a news fragment

PR #1197, opened 2021-04-01
convert all documentation to Markdown

benoit-pierre:
The readme was  converted from Markdown back when reStructuredText was the only supported format for a Python package's description, and then I guess the rest followed. We don't use the readme for the package description anymore, and even if we did, Markdown as been supported for a while. So convert everything back to Markdown so we don't have to deal with multiple documentation formats.


doc/releases.md
```diff
@@ -0,0 +1,17 @@
...
+
+- official release: `{major}.{minor}.{patch}` (e.g. `2.2.1`, `3.1.0`, `4.0.0`...)
+- pre-release: `{major}.{minor}.{patch}.{dev|rc}{devel}` (e.g. `4.0.0.dev7`, `4.0.0rc1`, ...)
+
+
+Steps to cut a new release (from a clean checkout of `master`):
+
+1. Update version in `plover/__init__.py`, and stage the change `git add plover/__init__.py`.
+2. Install towncrier (`pip install towncrier`) and update `NEWS.md`: `towncrier`.
+3. Review the stagged changes, check all news fragments in `news.d` were properly handled
```
user202729, line 13: (just notice that there's a typo here)

```suggestion
3. Review the staged changes, check all news fragments in `news.d` were properly handled
```

benoit-pierre: Good catch, will amend.

morinted (approved):

README.md
```diff
@@ -0,0 +1,133 @@
+# Plover
+
+> Bringing stenography to everyone.
+
+[![Linux CI badge][]][Linux CI] [![macOS CI badge][]][macOS CI] [![Windows CI badge][]][Windows CI]
+
+|              |              |          |          |                  |                  |
```
user202729, line 7: I overlooked that there is an extraneous table row here, it only become apparent when rendered on GitHub.

It could be a bug in whatever tool used to convert the format (if it was not done manually)...

benoit-pierre: Good catch, fixed in #1210. (For reference I used `pandoc -f rst -t gfm --wrap=preserve --reference-links` for the initial conversion).

PR #1201, opened 2021-04-01
setup: minor metadata cleanup

benoit-pierre:
Fix warning when checking the source distribution with `twine check`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,6 +21,7 @@ classifiers =
         Topic :: Adaptive Technologies
         Topic :: Desktop Environment
 keywords = plover
+long_description_content_type = text/x-rst
 
 [options]
 include_package_data = True


user202729:
With https://github.com/openstenoproject/plover/pull/1197 should this be text/markdown...

But then, currently the long_description is only the short string in `plover/__init__.py`, which is also the thing shown on pypi main page (what's the purpose of it being put there rather than all in setup.cfg and leave that file empty? On a related note, the copyright has not been updated for ages)

benoit-pierre:
Well, the content type does not really matter, since as you noticed we only fill it from the (not that long) string in `plover/__init__.py`. We may want to go back to using the readme, in which case will change it to Markdown, yes.

You're right that a copyright update is overdue...

morinted:
Do you want to update the copyright year in this PR?

benoit-pierre:
I'm preparing a big PR that depends on a number of those small PRs, so I would prefer to do that in another PR.

benoit-pierre:
I'll go ahead an merge this.

PR #1203, opened 2021-04-01
Add support for multiple issue templates

morinted:
Our current issue template is a little overwhelming as it tries to fit the three main use cases that we have issues for:

- Bug reports
- Feature requests
- Support requests

This pull request splits the templates into the three types so that the user is asked in-context questions about their issue or feature request.


benoit-pierre:
For the "support request" case, maybe we could enable this https://github.com/pypa/setuptools/issues/2557 ?

benoit-pierre (approved):

PR #1211, opened 2021-04-02
setup: fix `launch` command on Windows

benoit-pierre:
Same issue as with `dist_main.main`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,12 @@ class Launch(Command):
             python_path = os.environ.get('PYTHONPATH', '').split(os.pathsep)
             python_path.insert(0, sys.path[0])
             os.environ['PYTHONPATH'] = os.pathsep.join(python_path)
-            os.execv(sys.executable, [sys.executable, '-m', 'plover.main'] + self.args)
+            cmd = [sys.executable, '-m', 'plover.main'] + self.args
+            if sys.platform.startswith('win32'):
+                # Workaround https://bugs.python.org/issue19066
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -121,7 +121,12 @@ def run(self):
             python_path = os.environ.get('PYTHONPATH', '').split(os.pathsep)
             python_path.insert(0, sys.path[0])
             os.environ['PYTHONPATH'] = os.pathsep.join(python_path)
-            os.execv(sys.executable, [sys.executable, '-m', 'plover.main'] + self.args)
+            cmd = [sys.executable, '-m', 'plover.main'] + self.args
+            if sys.platform.startswith('win32'):
+                # Workaround https://bugs.python.org/issue19066
```
user202729, line 126:
```suggestion
                # Workaround https://bugs.python.org/issue19124
```

benoit-pierre: The problem is spaces in the executable directory, not a background execution issue.

user202729: Ah, I was confused. (Plover's workaround for background execution was the issue I got when I tried to execute plover_console_ui on Windows)

user202729: Looks like the "correct" duplicate is https://bugs.python.org/issue436259 which is closed (fixed?)?

benoit-pierre: Yeah ¯\\_(ツ)_/¯... That's with Python 3.8, I've not tested back with 3.6.

morinted (approved):

PR #1213, opened 2021-04-02
target macOS 10.13 (High Sierra) or later

benoit-pierre:
Let's not pretend we support older versions, like PyQt5 does... and we might as well update to the latest version of PyQt5.


morinted (approved):
Just verified that the artifact from this build runs successfully on 10.13 👍

PR #1214, opened 2021-04-03
setup: add target platform to macOS DMG filename

benoit-pierre:
=> `plover-4.0.0.dev8+344.gfdf0c01-macosx-10.13-x86_64.dmg`

This should make the minimum supported macOS release and supported architecture clearer.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -219,8 +219,11 @@ class BinaryDistDmg(Command):
 
     def run(self):
         self.run_command('bdist_app')
+        # Encode targeted macOS plaftorm in the filename.
+        from wheel.bdist_wheel import get_platform
+        platform = get_platform('dist/Plover.app')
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Wait, why is it 10.9?

morinted:
Maybe it's pulling it from python's minimum supported version?

benoit-pierre:
Yeah, I thought it would take `MACOSX_DEPLOYMENT_TARGET` into account.

benoit-pierre:
OK, it works now. And if the app directory actually contains libraries that need a more recent version than `MACOSX_DEPLOYMENT_TARGET`, the detected target platform will reflect that. (I would have preferred an exception, but still better than nothing).

morinted (approved):

PR #1219, opened 2021-04-03
gui_qt/dictionary_editor: improve updating a translation

benoit-pierre:
Don't delete the key first, so extra metadata (e.g. additional columns in an Excel dictionary) is not deleted.


user202729:
(for writing the changelog) Which means that if a dictionary entry is copied from one excel dictionary to another, the metadata is still not preserved; however editing non-JSON file formats from Plover is not a good idea.

benoit-pierre:
I don't think there's an expectation that extra metadata is copied from one dictionary to another. But if it's possible to preserve such data on modification, then we should do it.

morinted (approved):

PR #1221, opened 2021-04-04
Fix duplicate Ctrl+N shortcut on main window

### LINKED ISSUE #1215: Tool keyboard shortcuts not working on Windows
**Describe the bug**

The shortcuts listed in the Tools submenu don't seem to work, even if Plover is in focus. Asking around on Discord, this seems to affect other Windows users, but not Mac or Linux users.

**To Reproduce**

1. Open Plover
2. Press Ctrl+N, Ctrl+J, Ctrl+L or Ctrl+T
3. Nothing happens

**Expected behavior**

It should open up the relevant tools.

**Operating system**

 - OS: Windows 10
 - Plover Version: 4.0.0.dev8+66.g685bd33

**Hardware**

Keyboard

user202729:
## Summary of changes

Initially, there are two buttons with shortcut control+N on the main window -- the add translation button on the toolbar, and the one in the dictionary editor (which adds the translation on the **selected** dictionary; if there's no selected dictionary it has the same behavior as the other button.

I think that's more useful, so I remove the shortcut for the button; otherwise to remove the other shortcut do this

```diff
--- a/plover/gui_qt/dictionaries_widget.ui
+++ b/plover/gui_qt/dictionaries_widget.ui
@@ -143,9 +143,6 @@
    <property name="toolTip">
     <string>Add a new translation</string>
    </property>
-   <property name="shortcut">
-    <string>Ctrl+N</string>
-   </property>
   </action>
   <action name="action_MoveDictionariesUp">
    <property name="icon">
```

Closes #1215 .

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/gui_qt/add_translation_dialog.py
+++ b/plover/gui_qt/add_translation_dialog.py
@@ -13,7 +13,6 @@ class AddTranslationDialog(Tool, Ui_AddTranslationDialog):
     TITLE = _('Add Translation')
     ICON = ':/translation_add.svg'
     ROLE = 'add_translation'
-    SHORTCUT = 'Ctrl+N'
 
     def __init__(self, engine, dictionary_path=None):
         super().__init__(engine)


user202729:
There are some related issues...

* end-to-end test?
* If spectra lexer is installed, control+L is still broken.

   Plover should check if there's any shortcut duplicate programmatically? (which might not be that easy, because not all shortcuts are set programmatically, and there are internationalization)

benoit-pierre:
> Initially, there are two buttons with shortcut control+N on the main window -- the add translation button on the toolbar, and the one in the dictionary **editor**...

You mean the dictionaries **widget**, as opposed to the real editor. And which toolbars? This is confusing... The dictionaries widget toolbar has an "add translation" button (for the selected dictionary, otherwise the favorite), the application has another "add translation" button (selecting the favorite dictionary). Since the translation dialog allows selecting the target dictionary anyway, maybe we should just drop the dictionaries widget button.  

> Plover should check if there's any shortcut duplicate programmatically? (which might not be that easy, because not all shortcuts are set programmatically, and there are internationalization)

Yeah... Or maybe we just drop those shortcuts? When I created the new interface, I added shortcuts, because that's what regular applications do, but Plover is not a standard application: most of the time when you're using it, you not interacting with the interface. And when you do need to interact with it, commands like `{PLOVER:LOOKUP}` are a better alternative.

user202729:
You see, there's no user guide for Plover documentation, so I just guess what the name is -- but yes it's these two buttons.

I think that the in the dictionary widget is a little more convenient because it adds to the currently selected dictionary, but it's up to you.

Also what about the shortcut to open other tools? (like the spectra lexer) is there a dictionary entry to open it?

PR #1226, opened 2021-04-04
Add tox support

benoit-pierre:
Add a configuration for using [tox](https://tox.readthedocs.io/en/latest/index.html). I've been using tox for Plover and other Python projects for some time. This is spruced-up version of my configuration.

Rational: tox will take care of all the details of creating and managing an isolated virtual environment, installing the necessary dependencies, and isolating testsuite runs.

Quick howto:
- install tox
- then use `tox {-e envlist} {-- arguments}`

Use `tox -a -v` to get a list of available environments:
```
default environments:
test             -> run tests

additional environments:
py3              -> run tests using py3
py36             -> run tests using py36
py37             -> run tests using py37
py38             -> run tests using py38
py39             -> run tests using py39
py310            -> run tests using py310
release_prepare  -> prepare/finalize a release
release_finalize -> prepare/finalize a release
launch           -> launch plover
packaging_checks -> packaging sanity checks
plugins_install  -> install plugins into the environment
setup            -> run a `setup.py` command
```

So running `tox` with no arguments will create a virtual environment, install of the requirements (except plugins), setup Plover inside that environment (`./setup.py develop`), and run the testsuite (`./setup.py test`).

Running it again will execute the last step only (unless a requirement in `reqs/` was changed, in which case the environment will be automatically re-created).

The same virtual environment is reused by the following tox environments:
- `tox -e text -- ARGS`: run the testsuite
- `tox -e setup -- COMMAND`: run `./setup.py COMMAND`
- `tox -e launch -- ARGS`: run `./setup.py launch ARGS`
- `tox -e packaging_checks`: run the same packaging checks as the CI (add `-- -n` to see a dry-run of the exact checks)
- `tox -e plugins_install`: install the distribution plugins (or the specified plugins when run with `tox -e plugins_install -- REQS`)
- `tox -e release_prepare -- NEW_VERSION`: execute all the steps necessary for preparing a new release (patch the version to `NEW_VERSION`  and update `NEWS.md`, staging all the changes for review)
- `tox -e release_finalize`: finalize the release (commit the staged changes, create an annotated tag, and print the git command necessary for pushing the release to GitHub)

Note:
- the actual environment lives in `.tox/dev`, and can be "activated" like other virtual environment
- `plugins_install` does not use the plugins manager for installing
- the plugins manager currently does not support installing/uninstalling from a virtual environment (update forthcoming)

The `pyX` environments are lightweight tests only environments (with only the bare minimum requirements necessary to run the tests); e.g. running `tox -e 'py3{6,7,8,9}` will execute the testsuite for each version of Python we support.

Todo: consolidate and update the developer documentation to advocate using tox.


morinted:
This looks very cool.

I think this PR description could be largely copy/pasted into doc/tox.md (or maybe better to add to the existing developer guide) and then linked in the final section of the repo readme, and maybe remove the platform-specific instructions?

benoit-pierre:
I added some documentation.

doc/developer_guide.md
```diff
@@ -1,3 +1,62 @@
+# Environment setup
+
+You need Python >= 3.6 installed, and you need [tox](https://pypi.org/project/tox/) >= 3.10.
```
morinted, line 3: May be good to indicate that it can be pip installed, something like this?

```suggestion
You need Python >= 3.6 installed, and you need [tox](https://pypi.org/project/tox/) >= 3.10 (`python3 -m pip install tox`).
```

doc/developer_guide.md
```diff
@@ -1,3 +1,62 @@
...
+Using tox takes care of all the details of creating and managing an isolated
+virtual environment, installing the necessary dependencies, and isolating
+testsuite runs.
+
+The command for using tox is: `tox {-e envlist} {-- arguments}`. Use `tox -a
+-v` to get a list of available environments.
+
+The same virtual environment is reused by the following tox environments:
+- `tox -e text -- ARGS`: run the testsuite. This is the default environment
+  when not precised.
```
morinted, line 14:
```suggestion
- `tox -e test -- ARGS`: run the testsuite. This is the default environment
  when not provided.
```

doc/developer_guide.md
```diff
@@ -1,3 +1,62 @@
...
+- `tox -e setup -- COMMAND`: run `./setup.py COMMAND`.
+- `tox -e launch -- ARGS`: run `./setup.py launch ARGS`.
+- `tox -e packaging_checks`: run the same packaging checks as the CI (add `--
+  -n` to see a dry-run of the exact checks).
+- `tox -e plugins_install`: install the distribution plugins (or the specified
+  plugins when run with `tox -e plugins_install -- REQS`). Note that this does
+  not use the plugins manager for installing.
+- `tox -e release_prepare -- NEW_VERSION`: execute all the steps necessary for
+  preparing a new release; patch the version to `NEW_VERSION`  and update
```
morinted, line 23:
```suggestion
  preparing a new release; patch the version to `NEW_VERSION` and update
```

doc/developer_guide.md
```diff
@@ -1,3 +1,62 @@
...
+  `NEWS.md`, staging all the changes for review.
+- `tox -e release_finalize`: finalize the release; commit the staged changes,
```
morinted, line 25:
```suggestion
- `tox -e release_finalize`: finalize the release: commit the staged changes,
```

doc/developer_guide.md
```diff
@@ -1,3 +1,62 @@
...
+  create an annotated tag, and print the git command necessary for pushing the
+  release to GitHub.
+
+The actual virtual environment lives in `.tox/dev`, and can be "activated" like
+any other virtual environment.
```
morinted, line 30: I think that we should provide a link here to explain what `"activated" like any other virtual environment` means, maybe like this?
```suggestion
The actual virtual environment lives in `.tox/dev`, and can be "activated" like
any other [virtual environment.](https://docs.python.org/3/tutorial/venv.html)
```

doc/developer_guide.md
```diff
@@ -1,3 +1,62 @@
...
+
+You need Python >= 3.6 installed, and you need [tox](https://pypi.org/project/tox/) >= 3.10.
+
+Using tox takes care of all the details of creating and managing an isolated
+virtual environment, installing the necessary dependencies, and isolating
+testsuite runs.
+
+The command for using tox is: `tox {-e envlist} {-- arguments}`. Use `tox -a
+-v` to get a list of available environments.
+
```
morinted, line 11: Maybe right here it would be helpful to just state something like

> To set up and activate the tox environment for Plover and install all dependencies, in your cloned repository simply run:
>
> `tox`

It's so easy that I almost didn't trust it.

morinted: Alternatively I would put this right at the top, after saying that python and tox are required.

That way, I could see someone who's just trying to run from source would figure it out…

- Install python
- pip install tox
- tox
- Mac and Linux: ./launch.sh
- Windows: ./launch.bat

Or maybe we just need a really brief "Running from Source" for someone who's just looking to get set up as quickly as possible.

morinted (changes requested):

doc/developer_guide.md
benoit-pierre: Yeah, but that's a given, no? It's a link to PyPI. And you may want to use your distribution package.

doc/developer_guide.md
benoit-pierre: Above as well?

doc/developer_guide.md
benoit-pierre: I'll link to https://virtualenv.pypa.io/en/latest/user_guide.html?highlight=activate#activators.

doc/developer_guide.md
benoit-pierre: Except that won't work, you need to activate the environment before using lanch.sh/launch.bat. Honestly, I almost remove those. And whether you're using the bat or the sh on Windows will depend on your environment: since we requiring bash to be present for creating the distribution, someone might use the git shell.

doc/developer_guide.md
morinted: Yes, full colon to prefix a list 👍

doc/developer_guide.md
morinted: Fair, so would the real bootstraps be

- Install python
- pip3 install tox
- tox -e launch

?

morinted: I'm also cool with removing the launch scripts 👍

benoit-pierre: OK, I'll remove the launch/test scripts.

doc/developer_guide.md
morinted: Only a concern to me as there's always a chance this is someone's first python project, and they aren't familiar with what PyPI is (and assume that it's a program you have to download.)

I think that if we do make a quick bulleted list like I did in other comments, then it wouldn't be necessary to include a command here, or we could just say "available from pip"

benoit-pierre: If you want. Personally, I'd prefer things not to be as easy: I don't want people to mindlessly copy/paste commands, it's a recipe for disaster.

morinted: I understand the principle. I guess the trade-off is just inevitably someone's gonna open a discussion asking "How do I run from source" and another one that's "how to activate tox environment", but maybe that's a feature, not a bug.

doc/developer_guide.md
```diff
@@ -1,3 +1,53 @@
...
+The same virtual environment is reused by the following tox environments:
+- `tox -e text -- ARGS`: run the testsuite. This is the default environment
```
morinted, line 13: This command has a typo
```suggestion
- `tox -e test -- ARGS`: run the testsuite. This is the default environment
```

doc/developer_guide.md
```diff
@@ -1,3 +1,53 @@
...
+  when not provided.
+- `tox -e setup -- COMMAND`: run `./setup.py COMMAND`.
+- `tox -e launch -- ARGS`: run `./setup.py launch ARGS`.
```
morinted, line 16: Maybe we can label this one
```suggestion
- `tox -e launch -- ARGS`: run `./setup.py launch ARGS`. Run Plover from source.
```

morinted (approved):

PR #1228, opened 2021-04-04
Tweak setup metadata

benoit-pierre:
* use README.md for the long description contents
* update the project URLs

Preview: https://test.pypi.org/project/plover/4.0.666.dev7/


morinted (approved):

benoit-pierre:
Updated to ensure `README.md` does not use relative links, as those will break when the description is rendered on PyPI. I'll go ahead and merge.

PR #1230, opened 2021-04-05
gui_qt/config: fix configuring extension plugins

### LINKED ISSUE #1180: Configuration is not saved if the only change is enabling one extension plugin
## Summary

See title.

## Reproducing

* Open Plover
* find some way to install any extension plugin
* go to Plover configuration to enable it (by clicking the check box, obviously)
* click "Ok" in the configuration window
* exit Plover (control+Q)
* reopen Plover. Observe that the plugin is still disabled.

## Plover Version

Latest git master at the present.

## System

Arch GNU/Linux (but this should be system-independent), run from source.

benoit-pierre:
## Summary of changes

Don't mutate the original configuration set, or changes won't be saved.

Closes #1180.

### Pull Request Checklist

### DIFF
--- a/news.d/bugfix/1230.ui.md
+++ b/news.d/bugfix/1230.ui.md
@@ -1,0 +1,1 @@
+Fix changes to the list of enabled extension plugins not being saved to the configuration file.
--- a/plover/gui_qt/config_window.py
+++ b/plover/gui_qt/config_window.py
@@ -207,6 +207,9 @@ class MultipleChoicesOption(QTableWidget):
         self.setRowCount(0)
         if value is None:
             value = set()
+        else:
+            # Don't mutate the original value.
+            value = set(value)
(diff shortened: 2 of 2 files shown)


user202729:
Good job tracking the bug down... 

But would it be a potential trap for new GUI plugin developers?

benoit-pierre:
Well, now that `MultipleChoicesOption` works as expected, that should not be a problem. (Unless of course said developer decides to redevelop it's own version of that particular config widget).

user202729:
I was talking about people who create new GUI plugins (like the console ui)

morinted (approved):

benoit-pierre:
Ah. Yes, they could end up with the same issue. We might be able to work around that by changing `Config.__getitem__` to return a copy.

PR #1240, opened 2021-04-07
drop support for Linux packages

benoit-pierre:
Let's not pretend we support anything else than using the AppImage (or the separately maintained AUR packages). This include dropping the packpack (docker) stuff. As a result, I in-lined the AppImage docker support (which is necessary if you want to build a valid AppImage from a Linux distribution more recent than Ubuntu Xenial), and we can archive the [`plover-docker-images` repo](https://github.com/openstenoproject/plover-docker-images).


redvers:
Another actively maintained option is the nix package.

The nix package manager runs on any and every Linux distribution from busybox to BloatLinux.

I know you don't support it, but someone in our community does :-)

```
nixpkgs.plover.dev python3.6-plover-4.0.0.dev8
nixpkgs.plover.stable python2.7-plover-3.1.1
```

The nix foundation's CI pipelines build and test these packages.

benoit-pierre:
We won't provide packages for those distributions (we never really did), we'll still try to provide support when people using those externally maintained packages open an issue here.

morinted (approved):

PR #1241, opened 2021-04-07
Improve CI

benoit-pierre:
The initial goal was to add support for continuous GitHub releases on the master branch. Due to GA limitations (the allowed context types for `if:` tests varies, e.g. no access to `secrets.xxx` for `jobs.<job_id>.if`), and because I wanted the mechanism to be optional (configurable through a repository secret), I ended up piggy backing on another CI development not yet pushed: support for skipping jobs whose successful outcome is already known.

Please check `doc/ci.md` for the gory details.

Continuous release example [available here](https://github.com/benoit-pierre/plover/releases/tag/continuous). And here is the release job [cowardly refusing to downgrade a continuous release](https://github.com/benoit-pierre/plover/runs/2291611012?check_suite_focus=true).

Here is an example of job skipping in action:
- a initial push to the base branch resulted in a [full run](https://github.com/benoit-pierre/plover/actions/runs/724407001)
- a [PR was created](https://github.com/benoit-pierre/plover/pull/4), with a Linux specific change, resulting in a [partial run](https://github.com/benoit-pierre/plover/actions/runs/724433961)
- the PR was then updated with a news entry, resulting in a [fast run](https://github.com/benoit-pierre/plover/actions/runs/725796066)

Now that we're going to have a continuous release of master, maybe we should lower the retention period of artifacts, be a little nicer to GitHub? The default is 90 days, I think 30 days is more than enough.


.github/workflows/ci/helpers.sh
```diff
@@ -115,33 +115,156 @@ setup_python_env()
...
+      run_eval "last_version=\"\$(echo -n \"\$last_release\" | sed -n '/^# \\w\\+ \\([^ ]\\+\\)\$/{s//\\1/;P;Q0};\$Q1')\""
+      if python <<EOF
+from pkg_resources import parse_version
+import sys
+version = parse_version('$RELEASE_VERSION')
+last_version = parse_version('$last_version')
+sys.exit(0 if version < last_version else 1)
+EOF
+      then
+        die 1 "cowardling refusing to downgrade \`$tag\` from \`$last_version\` to \`$RELEASE_VERSION\`"
```
morinted, line 188:
```suggestion
        die 1 "cowardly refusing to downgrade \`$tag\` from \`$last_version\` to \`$RELEASE_VERSION\`"
```
?

.github/workflows/ci/helpers.sh
```diff
@@ -115,33 +115,156 @@ setup_python_env()
...
 }
 
 publish_pypi_release()
 {
   run "$python" -m twine upload dist/Source/* dist/Wheel/*
 }
 
+analyze_set_release_info()
+{
+  # Do ignore the 'continuous' tag!
```
morinted, line 213:
```suggestion
  # Ignore the 'continuous' tag!
```

doc/ci.md
```diff
@@ -66,25 +114,47 @@ This job will run a number of packaging-related checks. See
...
-are automatically generated from the last release section in `NEWS.md` and
-the template in `.github/RELEASE_DRAFT_TEMPLATE.md`.
+### GitHub release
+
+On *tagged* release, a new release draft is created on GitHub.
+
+On *continuous* release, the `continuous` release and corresponding tag
+are created / updated, but only if the existing release version is not
+newer, in order to:
+* prevent a old workflow re-run from overwriting the latest continuous release
```
morinted, line 142:
```suggestion
* prevent an old workflow re-run from overwriting the latest continuous release
```

morinted (changes requested):
Looks good to me, only had minor grammar nits 👍

.github/workflows/ci/helpers.sh
benoit-pierre: D'oh!

.github/workflows/ci/helpers.sh
benoit-pierre: I think I can just removed that part, now that the tag is ignored at the workflow level.

morinted (approved):

PR #1244, opened 2021-04-10
Add dictionaries "save as..." support

benoit-pierre:
## Summary of changes

Alternative to #1149.

See commit messages for the detail of the changes.

Rational for 8dcda12d7257432a3edfb3937521b6bbfbdc5908, and allowing to create copies when multiple dictionaries are selected: with this PR and our support for read-only `assets:` resources, I think we should consider changing the system default dictionaries definition to:
```python
DEFAULT_DICTIONARIES = (
    'user.json',
    DICTIONARIES_ROOT + '/commands.json',
    DICTIONARIES_ROOT + '/main.json',
)
```
This would result in:
- only `user.json` is (re-)created when missing
- the other 2 dictionaries are loaded read-only directly from the assets directory

This would make updating easier for users that don't change those dictionaries. And if you really want to edit those, you can always can now easily save a copy (and addressing #726 would provide an even better alternative).

### Pull Request Checklist
- [ ] Changes have tests
- [x] News fragment added in news.d.

### DIFF
--- a/plover/gui_qt/dictionaries_widget.py
+++ b/plover/gui_qt/dictionaries_widget.py
@@ -58,28 +58,6 @@ def _new_dictionary(filename):
     except Exception as e:
         raise Exception('creating dictionary %s failed. %s' % (filename, e)) from e
 
-def _get_dictionary_save_name(parent_widget, title, default_name=None,
-                              default_extensions=(), initial_filename=None):
-    if default_name is not None:
-        # Default to a writable dictionary format.
-        writable_extensions = set(_dictionary_formats(include_readonly=False))
-        default_name += '.' + next((e for e in default_extensions
-                                    if e in writable_extensions),
-                                   'json')
-        default_name = os.path.join(CONFIG_DIR, default_name)
-    else:
-        default_name = CONFIG_DIR
-    new_filename = QFileDialog.getSaveFileName(
-        parent=parent_widget, caption=title, directory=default_name,
-        filter=_dictionary_filters(include_readonly=False),
-    )[0]
-    if not new_filename:
-        return None
-    new_filename = normalize_path(new_filename)
-    if new_filename == initial_filename:
-        return None
-    return new_filename
-
 
 class DictionariesWidget(QWidget, Ui_DictionariesWidget):
 
@@ -95,6 +73,8 @@ class DictionariesWidget(QWidget, Ui_DictionariesWidget):
         self._config_dictionaries = {}
         self._loaded_dictionaries = {}
         self._reverse_order = False
+        # The save/open/new dialogs will open on that directory.
+        self._file_dialogs_directory = CONFIG_DIR
         for action in (
             self.action_Undo,
             self.action_EditDictionaries,
@@ -381,6 +361,29 @@ class DictionariesWidget(QWidget, Ui_DictionariesWidget):
         assert selection
         self._edit([self._config_dictionaries[row] for row in selection])
 
+    def _get_dictionary_save_name(self, title, default_name=None,
+                                  default_extensions=(), initial_filename=None):
+        if default_name is not None:
+            # Default to a writable dictionary format.
+            writable_extensions = set(_dictionary_formats(include_readonly=False))
+            default_name += '.' + next((e for e in default_extensions
+                                        if e in writable_extensions),
+                                       'json')
+            default_name = os.path.join(self._file_dialogs_directory, default_name)
+        else:
+            default_name = self._file_dialogs_directory
+        new_filename = QFileDialog.getSaveFileName(
+            parent=self, caption=title, directory=default_name,
+            filter=_dictionary_filters(include_readonly=False),
+        )[0]
+        if not new_filename:
+            return None
+        new_filename = normalize_path(new_filename)
+        self._file_dialogs_directory = os.path.dirname(new_filename)
+        if new_filename == initial_filename:
+            return None
+        return new_filename
+
     def _copy_dictionaries(self, dictionaries_list):
         need_reload = False
         title_template = _('Save a copy of {name} as...')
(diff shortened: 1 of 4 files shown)


plover/gui_qt/dictionaries_widget.py
```diff
@@ -91,6 +125,16 @@ def __init__(self, *args, **kwargs):
         self.menu_AddDictionaries.addAction(_(
             _('New dictionary'),
         )).triggered.connect(self._create_new_dictionary)
+        # Save menu.
+        self.menu_SaveDictionaries = QMenu(self.action_SaveDictionaries.text())
+        self.menu_SaveDictionaries.setIcon(self.action_SaveDictionaries.icon())
+        self.menu_SaveDictionaries.addAction(_(
+            _('Create a copy of each dictionary'),
```
user202729, line 132: ... double gettext call?

plover/gui_qt/dictionaries_widget.py
```diff
@@ -288,14 +332,25 @@ def _set_selection(self, row_list):
...
             self.action_MoveDictionariesUp,
             self.action_MoveDictionariesDown,
         ):
-            action.setEnabled(enabled)
+            widget.setEnabled(has_selection)
+        has_live_selection = next((
+            True for row in selection
+            if self._config_dictionaries[row].path
+            in self._loaded_dictionaries
+        ), False)
```
user202729, line 347: Isn't this just `any(self._config_dictionaries[row].path in self._loaded_dictionaries for row in selection)`?

plover/gui_qt/dictionaries_widget.py
benoit-pierre: Yeah... copying from above... o.O

plover/gui_qt/dictionaries_widget.py
user202729, line 347:
```suggestion
        has_live_selection = any(
            self._config_dictionaries[row].path in self._loaded_dictionaries
            for row in selection
        )
```

Is this correct?

plover/gui_qt/dictionaries_widget.py
```diff
@@ -91,6 +125,16 @@ def __init__(self, *args, **kwargs):
...
+        # Save menu.
+        self.menu_SaveDictionaries = QMenu(self.action_SaveDictionaries.text())
+        self.menu_SaveDictionaries.setIcon(self.action_SaveDictionaries.icon())
+        self.menu_SaveDictionaries.addAction(_(
+            _('Create a copy of each dictionary'),
+        )).triggered.connect(self._save_dictionaries)
+        self.menu_SaveDictionaries.addAction(_(
+            _('Merge dictionaries into a new one'),
+        )).triggered.connect(functools.partial(self._save_dictionaries,
+                                               merge=True))
```
user202729, line 137:
```suggestion
        self.menu_SaveDictionaries.addAction(
            _('Create a copy of each dictionary'),
        ).triggered.connect(self._save_dictionaries)
        self.menu_SaveDictionaries.addAction(
            _('Merge dictionaries into a new one'),
        ).triggered.connect(functools.partial(self._save_dictionaries,
                                               merge=True))
```

plover/gui_qt/dictionaries_widget.py
user202729: (side note: I didn't notice that comment being posted being making the suggested change below. But I'm sure that I did delete the comment above...)

benoit-pierre:
Thanks for the review @user202729! :+1:

plover/gui_qt/dictionaries_widget.py
```diff
@@ -327,6 +381,59 @@ def on_edit_dictionaries(self):
...
+            title = title_template.format(name=dictionary.short_path)
+            name, ext = os.path.splitext(os.path.basename(dictionary.path))
+            default_name = default_name_template.format(name=name)
+            new_filename = _get_dictionary_save_name(self, title, default_name, [ext[1:]],
+                                                     initial_filename=dictionary.path)
+            if new_filename is None:
+                continue
+            with _new_dictionary(new_filename) as d:
+                d.update(self._loaded_dictionaries[dictionary.path])
+            need_reload = True
```
morinted, line 398: Just tried using this feature, and wondering if it makes sense when handling multiple dictionaries to remember the path that it was saved to.

I.e. instead of dictionary.path for initial_filename, do that for only the first, and then on subsequent dialogs open with where `new_filename` was pointing to.

I could definitely see wanting to make a copy of X, Y, Z dictionaries and hoping that they all end up in the same folder.

(I accidentally didn't realize that the path wasn't sticky and just tried to blindly "save" and it saved to the wrong folder)

benoit-pierre:
> Just tried using this feature, and wondering if it makes sense when handling multiple dictionaries to remember the path that it was saved to.
> I.e. instead of dictionary.path for initial_filename, do that for only the first, and then on subsequent dialogs open with where new_filename was pointing to.
> I could definitely see wanting to make a copy of X, Y, Z dictionaries and hoping that they all end up in the same folder.
> (I accidentally didn't realize that the path wasn't sticky and just tried to blindly "save" and it saved to the wrong folder)

OK, so right know the directory if always the config folder, but instead use that for the first one, and keep the same folder for the others? What if you do another operation?

benoit-pierre:
It could make a lot of sense to: default to the config folder for the first invocation, and re-use the selected file directory for other invocations (including adding dictionaries).

morinted:
I generally agree that caching the last-used folder would be a good idea, it would help with adding, saving, and merging dictionaries in more complex workflows.

E.g. merge two dictionaries, add it to Plover, remove the other two.

My suggestion was just how to make the save multiple copies workflow, that takes it to the next level :)

benoit-pierre:
OK, let's do that.

benoit-pierre:
OK, done.

morinted (approved):

PR #1253, opened 2021-04-13
force the use of the Qt GUI when running from a distribution

benoit-pierre:
This is so someone does not end up with an invisible Plover running in headless mode (because that's the fallback if the Qt GUI cannot be loaded and no other one is available), possibly catching keyboard input. Note that it's still possible to override this choice using the `--gui` option.


morinted (approved):
Excellent.

Issue #789, opened 2017-07-26
Scan machine port to /dev/serial/by-id/whatever if available (eg on Linux)


JasonWoof:
Classification: Feature (New)

This is a feature request for Linux, though there might be other systems that could benefit from this.

The issue: when I unplug my ErgoDox, then plug it in again, it gets a different path in /dev so I have to open Plover settings, go to the machine tab, scan and ok it every time.

Today I realized that there might be a path in /dev that is the same (always points to my ErgoDox) and there is, it's ``/dev/serial/by-id/usb-ErgoDox_EZ_ErgoDox_EZ-if03`` in my case.

Now that I've put that in my settings, I'm hopefully all set.

My feature request is to have Plover use this path automatically when it scans, instead of the far more ephemeral ``/dev/ttyACM0`` or whatnot.

P.S. The former path is a link to the latter, i.e.:
```
$ ls -l /dev/serial/by-id/usb-ErgoDox_EZ_ErgoDox_EZ-if03
lrwxrwxrwx 1 root root 13 Jul 26 00:01 /dev/serial/by-id/usb-ErgoDox_EZ_ErgoDox_EZ-if03 -> ../../ttyACM0
```

benoit-pierre:
I'd be interested to know what ports are listed after updating to pyserial 3.4 and applying the following patch:
```diff
 plover/gui_qt/machine_options.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git i/plover/gui_qt/machine_options.py w/plover/gui_qt/machine_options.py
index 63dbc3b9..ebe87daf 100644
--- i/plover/gui_qt/machine_options.py
+++ w/plover/gui_qt/machine_options.py
@@ -54,7 +54,7 @@ class SerialOption(QWidget, Ui_SerialWidget):
 
     def on_scan(self):
         self.port.clear()
-        self.port.addItems(sorted(x[0] for x in comports()))
+        self.port.addItems(sorted(x[0] for x in comports(include_links=True)))
 
     def on_port_changed(self, value):
         self._update('port', value)
```

JasonWoof:
Thanks for looking into this!

Here's my transcript:

	$ python3
	Python 3.5.4 (default, Sep  5 2017, 18:32:10)
	[GCC 7.2.0] on linux
	Type "help", "copyright", "credits" or "license" for more information.
	>>> import serial
	>>> serial.__version__
	'3.4'
	>>> from serial.tools.list_ports import comports
[...]

PR #1254, opened 2021-04-13
show port description in port dropdown menu

antistic:
## Summary of changes

Add port description to the dropdown of available ports. E.g. instead of `/dev/ttyACM0` it shows `/dev/ttyACM0 - EcoSteno`.

image

This was mentioned briefly in #789 but doesn't address the main issue (but might be useful anyway).

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/gui_qt/machine_options.py
+++ b/plover/gui_qt/machine_options.py
@@ -67,16 +67,18 @@ class SerialOption(QWidget, Ui_SerialWidget):
         self.valueChanged.emit(self._value)
 
     def on_scan(self):
-        self.port.clear()
+        self.ports.clear()
         for port in comports():
             portItem = QStandardItem(str(port))
             portItem.setData(port.device)
             self.ports.appendRow(portItem)
 
     def on_port_changed(self, value):
-        port = self.ports.item(self.port.currentIndex()).data()
-        self.port.setCurrentText(port)
-        self._update('port', port)
+        portItem = self.ports.item(self.port.currentIndex())
+        if portItem:
+            port = portItem.data()
+            self.port.setCurrentText(port)
+            self._update('port', port)
 
     def on_baudrate_changed(self, value):
         self._update('baudrate', int(value))
--- a/plover/gui_qt/machine_options.py
+++ b/plover/gui_qt/machine_options.py
@@ -63,10 +63,12 @@ class SerialOption(QWidget, Ui_SerialWidget):
 
     def on_scan(self):
         self.port.clear()
-        self.port.addItems(sorted(x[0] for x in comports()))
+        self.port.addItems(sorted(f"{port.device} ({port.description})" for port in comports()))
 
     def on_port_changed(self, value):
-        self._update('port', value)
+        port = value.split(" ")[0]
+        self.port.setCurrentText(port)
+        self._update('port', port)
 
     def on_baudrate_changed(self, value):
         self._update('baudrate', int(value))
--- a/plover/gui_qt/machine_options.py
+++ b/plover/gui_qt/machine_options.py
@@ -14,6 +14,7 @@ from plover.gui_qt.i18n import get_gettext
 
 _ = get_gettext()
 
+
 class SerialOption(QWidget, Ui_SerialWidget):
 
     valueChanged = pyqtSignal(QVariant)
--- a/plover/gui_qt/machine_options.py
+++ b/plover/gui_qt/machine_options.py
@@ -69,14 +69,14 @@ class SerialOption(QWidget, Ui_SerialWidget):
     def on_scan(self):
         self.ports.clear()
         for port in comports():
-            portItem = QStandardItem(str(port))
-            portItem.setData(port.device)
-            self.ports.appendRow(portItem)
+            port_item = QStandardItem(str(port))
+            port_item.setData(port.device)
+            self.ports.appendRow(port_item)
 
     def on_port_changed(self, value):
-        portItem = self.ports.item(self.port.currentIndex())
-        if portItem:
-            port = portItem.data()
+        port_item = self.ports.item(self.port.currentIndex())
+        if port_item:
+            port = port_item.data()
             self.port.setCurrentText(port)
             self._update('port', port)
 
--- a/plover/gui_qt/machine_options.py
+++ b/plover/gui_qt/machine_options.py
@@ -2,6 +2,7 @@ from copy import copy
 
 from PyQt5.QtCore import QVariant, pyqtSignal
 from PyQt5.QtWidgets import QWidget
+from PyQt5.QtGui import QStandardItemModel, QStandardItem
 
 from serial import Serial
 from serial.tools.list_ports import comports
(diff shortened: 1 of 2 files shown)


sammdot (commented):
* Can you give an example of what this would look like (either as an image or the text equivalent)? I'm trying `port.description` on my machine and I don't seem to get anything (either empty or `n/a`) for any of my serial ports.
* Is it safe to assume that the port names won't have spaces? Is there maybe a neater way of adding values to the individual items?

benoit-pierre:
On which platform(s) was this tested? (I think detailed info is not available on all platform), see: https://pythonhosted.org/pyserial/tools.html#serial.tools.list_ports.comports.

benoit-pierre:
>     * Is it safe to assume that the port names won't have spaces? Is there maybe a neater way of adding values to the individual items?

I agree.

antistic:
Tested on Linux (I don't have access to another platform atm)

plover/gui_qt/machine_options.py
```diff
@@ -13,14 +14,17 @@
 
 _ = get_gettext()
 
-
```
user202729, line 16: It's best to not make unnecessary style changes.

plover/gui_qt/machine_options.py
```diff
@@ -63,10 +67,13 @@ def _update(self, field, value):
...
         self.port.clear()
-        self.port.addItems(sorted(f"{port.device} ({port.description})" for port in comports()))
+        for port in comports():
+            portItem = QStandardItem(str(port))
+            portItem.setData(port.device)
+            self.ports.appendRow(portItem)
 
     def on_port_changed(self, value):
-        port = value.split(" ")[0]
+        port = self.ports.item(self.port.currentIndex()).data()
```
user202729, line 76: This doesn't work if I just press the "scan" button below... (`item()` returns None)

plover/gui_qt/machine_options.py
antistic: fixed in 7c262f88d161f70e3017bf86d5c91d6f593e2e38

plover/gui_qt/machine_options.py
antistic: This happens when you don't select something (e.g. on start up). I did a check on the item but there's a case for putting the check on `currentIndex() == -1` instead. 9819ed0eb54681b1eb9fd5432a867b79c5d2e69b

benoit-pierre (commented):
This approach is fundamentally at odds with the way the combo box is configured: it's specifically setup to be editable so (more knowledgeable) users can edit the field  (e.g. to force the use of a by-id device link on Linux, or use a fake serial port for testing).

Instead, we could use a multi-lines tooltip, something along those lines:
```diff
 plover/gui_qt/machine_options.py | 27 +++++++++++++++++++++++++--
 1 file changed, 25 insertions(+), 2 deletions(-)

diff --git i/plover/gui_qt/machine_options.py w/plover/gui_qt/machine_options.py
index 61a7a60e..62f92bb5 100644
--- i/plover/gui_qt/machine_options.py
+++ w/plover/gui_qt/machine_options.py
@@ -1,6 +1,6 @@
 from copy import copy
 
-from PyQt5.QtCore import QVariant, pyqtSignal
+from PyQt5.QtCore import Qt, QVariant, pyqtSignal
 from PyQt5.QtWidgets import QWidget
 
 from serial import Serial
@@ -14,6 +14,25 @@
 _ = get_gettext()
 
 
+def port_info_tooltip(port_info):
+    parts = []
+    description = getattr(port_info, 'description')
+    if description not in {None, 'n/a'}:
+        tooltip.append(description)
+    for attr, fmt in (
+        ('manufacturer', _('- manufacturer: {value}')),
+        ('product', _('- product: {value}')),
+    ):
+        value = getattr(port_info, attr)
+        if value:
+            tooltip.append(fmt.format(value=value))
+    if not parts:
+        return None
+    if len(parts) > 1:
+        parts.insert(1, '')
+    return '\n'.join(parts)
+
+
 class SerialOption(QWidget, Ui_SerialWidget):
 
     valueChanged = pyqtSignal(QVariant)
@@ -63,7 +82,11 @@ def _update(self, field, value):
 
     def on_scan(self):
         self.port.clear()
-        self.port.addItems(sorted(x[0] for x in comports()))
+        for n, port_info in enumerate(sorted(comports())):
+            self.port.addItem(port_info.device)
+            tooltip = port_info_tooltip(port_info)
+            if tooltip is not None:
+                self.port.setItemData(n, tooltip, Qt.ToolTipRole)
 
     def on_port_changed(self, value):
         self._update('port', value)
```
(Patch against master)

antistic:
> This approach is fundamentally at odds with the way the combo box is configured: it's specifically setup to be editable so (more knowledgeable) users can edit the field (e.g. to force the use of a by-id device link on Linux, or use a fake serial port for testing).

That wasn't intended, maybe I should have set this pull request to WIP since I'm new to Qt and still figuring it out :sweat_smile: . I think in theory it should be okay to use the model approach since [QComboBox does use a `QStandardItem ` under the hood](https://doc.qt.io/qt-6/qcombobox.html#details), but replacing it wasn't quite as straightforward as I hoped.

> Instead, we could use a multi-lines tooltip, something along those lines:

I think that's much simpler code-wise, and contains more useful information, but I know that I personally wouldn't even think to try looking at the tool tip to see this information. I could imagine it being a more useful if the tooltip appeared quicker/instantly?

benoit-pierre:
> > Instead, we could use a multi-lines tooltip, something along those lines:
> 
> I think that's much simpler code-wise, and contains more useful information, but I know that I personally wouldn't even think to try looking at the tool tip to see this information. I could imagine it being a more useful if the tooltip appeared quicker/instantly?

Agreed.

petercpark:
When will this be merged? I think this is great.

mrlinuxfish:
At the very least, adding path completion to the text box would go a long way to making it easier to change out keyboards and use `/dev/serial/by-id/` instead of the `/dev/TTYAMC0` (which changes every time the board is unplugged)

antistic:
I wanted to come back to this PR because it seems popular, but I'm not sure I understand what the current issues are.

I don't really like the tooltip idea — it seems really different UX-wise to other programs I've used, so I would prefer a different solution. But I don't think I can currently propose a better solution than the current implementation until I understand what the issues are.

My guesses:

- It doesn't give any extra information for Windows
- It doesn't give all the available information on other platforms (e.g. does not include manufacturer on Linux)
- The list item text and data are not the same, and this might be confusing
- It needs to be rebased on the current master

@Benoit (or other reviewers), can you let me know if I've missed anything, the priority of the problems, and which problems are deal breakers?

benoit-pierre:
> - It doesn't give any extra information for Windows
> - It doesn't give all the available information on other platforms (e.g. does not include manufacturer on Linux)
> - The list item text and data are not the same, and this might be confusing
> - It needs to be rebased on the current master

Yes, see #1510.

antistic:
:+1:

Closing in favour of #1510

PR #1258, opened 2021-04-15
Fix and improve i18n support

benoit-pierre:
## Summary of changes

The bad news is that our i18n support is broken, particularly for plugins that would want to use it. The good news is that since nobody is going to be using it, we can fix without fear of regressions!

See individual commits messages for a detail of the change, as well as the new documentation.

### Pull Request Checklist
- [ ] Changes have tests


benoit-pierre:
And I broke all the tests... Because `setup.py` will import plover, which will try to setup the`i18n` support and that will fail because the `.egg-info` directory is not present.

benoit-pierre:
I added a workaround:

- use `exec` in `setup.py`
- detect `exec` is being used in `plover`, and disable i18n support

Note: in this case ("import" from `setup.py`), we don't want i18n support anyway, since some of the strings are used for package metadata.

benoit-pierre:
Fixed the Linux build, as well as some `is_release` checks in the packaging job.

doc/i18n.md
```diff
@@ -0,0 +1,167 @@
...
+### Testing
+
+Save your translation, and start / restart Plover.
+
+
+### Tips & Tricks
+
+- You can use "File -> Open config folder" in Plover's menu to easily locate
+  its configuration folder. 
+
```
morinted, line 159: spaces in list should be removed here too

benoit-pierre:
Of course this broke towncrier too... Anyway I think this is finally ready now.

morinted (approved):

benoit-pierre:
And I broke the dictionaries file picker:
```diff
diff --git a/plover/gui_qt/dictionaries_widget.py b/plover/gui_qt/dictionaries_widget.py
index 74c04688..4ffaf346 100644
--- a/plover/gui_qt/dictionaries_widget.py
+++ b/plover/gui_qt/dictionaries_widget.py
@@ -40,7 +40,7 @@ def _dictionary_filters(include_readonly=True):
     formats = sorted(_dictionary_formats(include_readonly=include_readonly))
     filters = ['*.' + ext for ext in formats]
     # i18n: Widget: “DictionariesWidget”, file picker.
-    filters = [_('Dictionaries ({extensions}').format(extensions=' '.join(filters))]
+    filters = [_('Dictionaries ({extensions})').format(extensions=' '.join(filters))]
     filters.extend(
         # i18n: Widget: “DictionariesWidget”, file picker.
         _('{format} dictionaries ({extensions})').format(
```

All good now! I think...

benoit-pierre:
@Sillabix, @nvdaes: could you update the Italian / Spanish translations? This should be the last translations update before the final release. Here is the [latest release](https://github.com/openstenoproject/plover/releases/tag/continuous), and do check the new [i18n documentation](https://github.com/openstenoproject/plover/blob/master/doc/i18n.md#for-translators). Pretty much everything should be translatable now, except for plugins (plugins manager included).

nvdaes:
Í will update the spanish translation this weekend, thanks.

Sillabix:
Ok, same for me

PR #1271, opened 2021-04-18
osx: Fix notifications not showing as from Plover

### LINKED ISSUE #1237: Notifications are shown as from "Python" instead of "Plover"
**Describe the bug**

Since switching to using relocatable-python, Mac notifications are shown as from "Python" with the python launcher icon, instead of Plover.

**To Reproduce**

1. Download a DMG build from GitHub Actions, install it, and start Plover.
2. Generate a notification by trying to open a machine on an invalid serial port.
3. See that the notification comes from python, not Plover, and the icon is wrong.

**Expected behavior**

The warning has the Plover icon and the Plover application name.

**Screenshots**

image

**Operating system**

 - OS: macOS Big Sur
 - Plover Version starting from 5fc30340178c236ccd9b9ae5090f328ba3f222e2

**Additional details**

One dirty but simple workaround could be modifying the PList and .icns file in Plover.app/Contents/Frameworks/Python.framework/Versions/Current/Resources/Python.app/Contents/ to have Plover's icon and name instead of python's.

morinted:
- Extract the Python binary from the bundled Python.app
- Unsign it so that our custom PList is respected

Closes #1237 

- [ ] Changes have tests
- [x] Add news fragment

### DIFF
--- a/plover/oslayer/log_osx.py
+++ b/plover/oslayer/log_osx.py
@@ -19,8 +19,7 @@ class OSXNotificationHandler(logging.Handler):
         # Notification Center has no levels or timeouts.
         notification = NSUserNotification.alloc().init()
 
-        notification.setTitle_(__software_name__.capitalize())
-        notification.setSubtitle_(record.levelname.title())
+        notification.setTitle_(record.levelname.title())
         notification.setInformativeText_(self.format(record))
 
         ns = NSUserNotificationCenter.defaultUserNotificationCenter()
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,9 @@ nosetests.xml
 .idea
 .cache
 
+# VSCode
+.vscode
+
 #Mr Developer
 .mr.developer.cfg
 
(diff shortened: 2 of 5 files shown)


plover/oslayer/log_osx.py
```diff
@@ -19,7 +19,6 @@ def emit(self, record):
         # Notification Center has no levels or timeouts.
         notification = NSUserNotification.alloc().init()
 
-        notification.setTitle_(__software_name__.capitalize())
         notification.setSubtitle_(record.levelname.title())
```
sammdot, line 22: Can we make this the notification title instead?

morinted: I'll try it out

benoit-pierre: Should this be kept for older macOS versions?

morinted: Looks like we were the odd ones out -- on older versions of Mac, they didn't include the app name, just the icon:

image

and this PR on big sur:

image

morinted:
Force pushed to use title instead of subtitle, added a small news fragment.

benoit-pierre (approved):

sammdot (approved):

PR #1276, opened 2021-04-19
Add support for output plugins

benoit-pierre:
## Summary of changes

Add support for output plugins: add 2 new entrypoint types:
* `output`: for the output plugins themselves
* `gui.qt.output_option`: similarly to machine plugins, provide a way for a custom output plugin to provide a dedicated GUI widget for configuring the output

TODO: 
- [x] cleanup, there are some `u'...'` strings leftover from when Python 2 was supported
- [ ] move relevant docstrings from the platform specific implementations for `KeyboardCapture` / `KeyboardEmulation` to the new base classes in `plover.oslayer.keyboardcontrol_base`

Otherwise pretty close to ready for merging, I think.

### Pull Request Checklist
- [x] Changes have tests
- [ ] News fragment added in news.d. See [documentation](https://github.com/openstenoproject/plover/blob/master/doc/developer_guide.md#making-a-pull-request) for details

PR #1278, opened 2021-04-19
Add support for locking keys

benoit-pierre:
## Summary of changes

Closes #1161. Depends on #1276 (included).

How it works:
- a regular `{#combo}` key combo now can end with keys left in the pressed state
- a new `{##combo}` key combo syntax allow continuing a preceding combo
- otherwise, sending a new combo, sending backspaces, or sending a string will automatically disengage previously held keys
- when keys are being held, the keyboard machine is fully suppressed (the whole keyboard is suppressed, all keys, including events with modifiers)
- that's pretty much it!

I ended up moving some of the key combo handling one level higher, in the engine: the `KeyboardEmulation.send_key_combination` method no expect a pre-parsed list of event to send (pairs of `[pressed: boolean, key_code]`), parsing and validation is done in the engine.

Because the Windows implementation does not monitor keyboard layout changes though events, but instead would refresh the layout before sending a key combo / backspaces / string, the output instance can now be used as a context manager:
```python
with output:
    output.send_backspaces(...)
    output.send_string(...)
```
The layout is refreshed on `__enter__` on Windows. And on Linux, `display.sync()` is called at `__exit__`.

The (keyboard) machine suppression now has 3 possible modes:
* `SUPPRESSION_NONE`: no suppression (e.g. when output is disabled)
* `SUPPRESSION_PARTIAL`: only suppress some keys (standard mode when output is enabled)
* `SUPPRESSION_FULL`: full suppress the keyboard: only used while a combo is active (held keys)

TODO:
- more tests, including actually testing it works as expected on Linux/Windows/macOS
- finish TODO...

### Pull Request Checklist
- [x] Changes have tests, but more needed


user202729:
Note that this is a breaking change (for GUI plugins and others)

What's with the `u'...'` strings? Isn't Python 2 support dropped already?

It would be better to drop the `threading.Thread` inheritance, now that there's multiple inheritance...

> In fact it's forbidden to override/define anything other than `run` and `__init__` in a subclass of `Thread`, and in the future when/if output plugin is implemented, it's going to be a mess (with multiple inheritance, `KeyboardEmulationBase` and `XEventLoop`).

user202729:
Note: need a rebase.

petercpark:
I'm interested in this pull request. Can we make it happen?

PR #1284, opened 2021-04-21
Fix 1279

### LINKED ISSUE #1279: Raise and focus running instance of Plover when launching it while running

We get a lot of support requests when people try to start Plover and it says "Another instance of Plover is already running". Often it's just minimized to the tray. I've noticed that other programs on my computer handle this by just raising the minimized window rather than trying to open a second instance.

When a second instance of Plover with QT is launched and another instance with QT UI is already running, raise and focus the main Plover window instead.

**Additional context**

This will break any forks where people run multiple instances of Plover at once, but I'm not really worried about that. It might also lead people to believe that Plover is really closing every time that they hit the "x" button, but that's already happening so… maybe not worth solving.

benoit-pierre:
## Summary of changes

Remove the process lock, switch to a custom controller object for both ensuring that only one instance of Plover is running, and sending commands to that instance. Reuse the `focus` command to restore and focus the main window when attempting to launch a second time.

Closes #1279.

### Pull Request Checklist
- [ ] Changes have tests


user202729:
Note.

* Now Plover is remote-controllable (how much?), is it controllable from other machines in the same LAN?
* Yet another breaking change. Conflict somewhat with output plugin.
* Aside: I also made a plugin (plover-websocket-server-2) to remotely control Plover, but I guess it suffer from the above problem more; besides I'm probably the only one to use it.

benoit-pierre:
>     * Now Plover is remote-controllable (how much?), is it controllable from other machines in the same LAN?

Only from the same machine: Unix socket on Linux/macOS, and named pipe on Windows.

user202729:
Note: what happens if Plover is not properly exited and the file `/tmp/plover_socket` is not deleted?

On the other hand the old lock thing might have the same issue.

benoit-pierre:
> Note: what happens if Plover is not properly exited and the file `/tmp/plover_socket` is not deleted?

In case of hard crash (segfault, abort), it's possible that the socket won't be properly deleted.

> On the other hand the old lock thing might have the same issue.

Not on Windows, since it was using an OS primitive.

user202729:
It is possible to do something similar to vim -- include the PID information somewhere, and if the PID is dead assume that the socket is left after a crash (and just force delete it).

benoit-pierre:
> It is possible to do something similar to vim -- include the PID information somewhere, and if the PID is dead assume that the socket is left after a crash (and just force delete it).

Changed to something simpler: if we can't connect to send the `focus` command, clean-up any stray Unix socket and restart.

benoit-pierre:
OK, I think this is ready for review.

morinted (approved):
Looks good to me. On Mac, when Plover is in system tray, the focus puts Plover in focus (menu bar updates) but no main window appears. This is consistent with how quicktime works so I think it's okay.

PR #1293, opened 2021-04-24
Disable tabKeyNavigation for all tables

user202729:
## Summary of changes

Title.

**Note**: pull request only opened for running GitHub CI. It isn't clear whether this is the best option.

An alternative behavior would be

* Tab will jump to the next row, or the next component if the current row is the last.
* Shift+tab will jump to the previous row, or the previous component if the current row is the first.

However this requires more than toggling a simple flag.

Note: even with this, shift+tab cannot cycle through all components in the main interface -- might need more investigation.

### Pull Request Checklist
- [ ] Changes have tests *(GUI only, cannot test)*


nvdaes:
I've tested this locally running from source and I'd like this change to be included in Plover.
I'd also like that tables has a name, to be read when focus moves to them.
Thanks

benoit-pierre:
Closing in favor of #1308, which this PR commit.

PR #1304, opened 2021-04-29
Redirect indirect imports from GUI forms

fourshade:
## Summary of changes

I just tried to build the latest version of Plover and got a screenful of import errors that I tracked down to imports of the _ translation macro from the generated GUI forms. There's already a circular dependency between the GUI modules and their Qt forms that is handled by having the forms code do imports at the bottom, but the extra import was too much.

After redirecting the imports, I got a clean build. It's possible that my particular combination of build tools contributed to the errors, but even if so, direct imports from the macro's definition in the root module seem like a better idea than indirect imports through dynamically-generated Python code.

EDIT: The build issue turned out to be with the UI code generator, but those imports still don't need to be routed through the stuff it spits out.

### DIFF
--- a/plover/gui_qt/add_translation_widget.py
+++ b/plover/gui_qt/add_translation_widget.py
@@ -6,6 +6,7 @@ from os.path import split as os_path_split
 from PyQt5.QtCore import QEvent, QTimer
 from PyQt5.QtWidgets import QApplication, QWidget
 
+from plover import _
 from plover.misc import shorten_path
 from plover.steno import normalize_steno, sort_steno_strokes
 from plover.engine import StartingStrokeState
@@ -13,7 +14,7 @@ from plover.translation import escape_translation, unescape_translation
 from plover.formatting import RetroFormatter
 from plover.resource import resource_filename
 
-from plover.gui_qt.add_translation_widget_ui import _, Ui_AddTranslationWidget
+from plover.gui_qt.add_translation_widget_ui import Ui_AddTranslationWidget
 
 
 class AddTranslationWidget(QWidget, Ui_AddTranslationWidget):
--- a/plover/gui_qt/add_translation_dialog.py
+++ b/plover/gui_qt/add_translation_dialog.py
@@ -1,4 +1,6 @@
-from plover.gui_qt.add_translation_dialog_ui import _, Ui_AddTranslationDialog
+from plover import _
+
+from plover.gui_qt.add_translation_dialog_ui import Ui_AddTranslationDialog
 from plover.gui_qt.tool import Tool
 
 
--- a/plover/gui_qt/config_window.py
+++ b/plover/gui_qt/config_window.py
@@ -25,11 +25,12 @@ from PyQt5.QtWidgets import (
     QWidget,
 )
 
+from plover import _
 from plover.config import MINIMUM_UNDO_LEVELS
 from plover.misc import expand_path, shorten_path
 from plover.registry import registry
 
-from plover.gui_qt.config_window_ui import _, Ui_ConfigWindow
+from plover.gui_qt.config_window_ui import Ui_ConfigWindow
 from plover.gui_qt.config_file_widget_ui import Ui_FileWidget
 from plover.gui_qt.utils import WindowState
 
--- a/plover/gui_qt/dictionaries_widget.py
+++ b/plover/gui_qt/dictionaries_widget.py
@@ -18,15 +18,15 @@ from PyQt5.QtWidgets import (
     QWidget,
 )
 
+from plover import _
 from plover.config import DictionaryConfig
 from plover.dictionary.base import create_dictionary
 from plover.engine import ErroredDictionary
 from plover.misc import normalize_path
 from plover.oslayer.config import CONFIG_DIR
 from plover.registry import registry
-from plover import log
 
-from plover.gui_qt.dictionaries_widget_ui import _, Ui_DictionariesWidget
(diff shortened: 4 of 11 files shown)


user202729:
If you run it from source, it's more likely that you just need `python setup.py build_ui -f`.

fourshade:
The forms built successfully, that wasn't the problem. The problem was an inability to resolve the _ import at runtime. Here's what happened when the GUI plugins tried to import _ from the generated .py form.

```
2021-04-29 12:06:06,845 [MainThread] ERROR: error loading gui.qt.tool plugin: lookup (from plover.gui_qt.lookup_dialog)
  File ".\plover\plover\registry.py", line 56, in register_plugin_from_entrypoint
    obj = entrypoint.load()
  File "c:\python38\lib\site-packages\pkg_resources\__init__.py", line 2450, in load
    return self.resolve()
  File "c:\python38\lib\site-packages\pkg_resources\__init__.py", line 2456, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File ".\plover\plover\gui_qt\lookup_dialog.py", line 6, in <module>
    from plover.gui_qt.lookup_dialog_ui import _, Ui_LookupDialog
ImportError: cannot import name '_' from 'plover.gui_qt.lookup_dialog_ui' (.\plover\plover\gui_qt\lookup_dialog_ui.py)
```

user202729:
Just try running the commands. (and read the linked issue for a little more explanation)

benoit-pierre:
I don't see why this would fail... There's no circular import. What does `plover.gui_qt.lookup_dialog_ui` looks like?

fourshade:
When I build the files sometimes there's an import at the bottom of the generated forms that messes things up
```

# -*- coding: utf-8 -*-

_ = __import__(__package__.split(".", 1)[0])._

...

from PyQt5 import QtCore, QtGui, QtWidgets

class Ui_LookupDialog(object):

    def setupUi(self, LookupDialog):

....

from plover.gui_qt.suggestions_widget import SuggestionsWidget

```

I can't figure out what is causing it to generate that last line; without it the build does work.

But I stand by my assertion that indirect imports are bad Python mojo.

benoit-pierre:
> When I build the files sometimes there's an import at the bottom of the generated forms that messes things up
> [...]
> I can't figure out what is causing it to generate that last line; without it the build does work.

OK, so not a circular import issue... Maybe some specific designer-qt setting for setting the Python path for custom widgets?

> But I stand by my assertion that indirect imports are bad Python mojo.

Fair enough.

news.d/bugfix/1304.ui.md
```diff
@@ -0,0 +1 @@
+-Fix circular imports involving the translation macro and GUI forms.
```
benoit-pierre, line 1: It's not a user visible change, and the description is wrong, so I suggest you just remove the fragment.

plover/gui_qt/add_translation_widget.py
```diff
@@ -6,14 +6,16 @@
 from PyQt5.QtCore import QEvent, QTimer
 from PyQt5.QtWidgets import QApplication, QWidget
 
+from plover import _
+
```
benoit-pierre, line 10: Please remove the extra empty line.

benoit-pierre (changes requested):

benoit-pierre:
> When I build the files sometimes there's an import at the bottom of the generated forms that messes things up
> [...]
> I can't figure out what is causing it to generate that last line; without it the build does work.

Even with that last line, I don't have any issue.

benoit-pierre:
If you can still reproduce the problem, please provide an archive of the generated UI Python files. That's with a pristine checkout, right?

plover/gui_qt/add_translation_widget.py
fourshade: Oops, missed one! I got it.

(The newline formatting isn't terribly consistent in most of the code, so I'm surprised anyone cared...but this one would have bothered me too. Good catch!)

news.d/bugfix/1304.ui.md
fourshade: Right, this PR's focus is changed; still an improvement but much more minor.

(and since it's not a big deal any more I get to see what a force push does to the arrangement of comments on a PR)

fourshade:
I think I figured out what's causing it. If I force-regenerate all the files with -f, that line doesn't appear, but if I manually delete them all and regenerate them without the switch, the last line does appear.

[tries again] Except when I'm watching, then it behaves. >:(

My initial build was a pristine checkout, but who knows if the ui generator has a cache that got gunked up. Non-replicability SUCKS.

benoit-pierre (approved):

fourshade:
> 
> 
> > When I build the files sometimes there's an import at the bottom of the generated forms that messes things up
> > [...]
> > I can't figure out what is causing it to generate that last line; without it the build does work.
> 
> Even with that last line, I don't have any issue.

I think I picked the worst possible example there...that file in isolation won't cause a problem since the lookup dialog uses the suggestions widget. There was a circularity appearing between between the dictionary editor and its own custom widget, but now it won't manifest...pyuic is just screwing with me now.

Can't find the original tracebacks in plover.log either, but I do remember something about a bad unicode character that made Python's logging system blow up at the end. My Python runtime shat itself and I have no evidence. Must be one of those days.

benoit-pierre:
Hooray for Heisenbugs... ¯\\_(ツ)_/¯

PR #1310, opened 2021-05-05
Remove duplicates from wordlist

fourshade:
## Summary of changes

This one's been on my todo list for a while. It's not critical, but it opens the door for further improvements.

The list of English words used by the orthography code currently contains over 1000 words that are duplicated with different ranks.  A lot of these words are rather common ("box", "ice", "spin", etc.), and it would mess up the orthography engine if they used the wrong rank. This is currently handled by sorting the word list and keeping the lowest number, but sorting is an expensive operation on a list this large and I can't think of any other reason it's being done at all.

All I did was prune the duplicates to match what the current implementation uses. Both the old and new files evaluate equal under Python dictionary equality when loaded by the current commit ([904271](https://github.com/openstenoproject/plover/commit/9042719e4cf21f2255492bd9679a9d2f7a9ef334))

(Is a news file required for something like this? It's a large change in pure bytes, but by itself this change has no effect on the program's behavior)


fourshade:
(missed a few lines that had the same word *and* rank. slippery buggers. equality tests hold.)

```
And now I'm just bored enough to test where exactly each word is on the orthography precedence scale for a random sentence
10  10  35  10   20    10     10 10   10    10      10   10   10 10 10  35          20         10    10  1010     10 (new list)
10  10  35  10   20    10     10 10   10    10      10   10   10 10 10  35          20         10    10  1010     10 (old list) 
```

...is it a sign of brain damage to have fun with a Python console?

benoit-pierre:
No need for a news entry. Can you add a simple test to ensure the list does not contain duplicates (like `test_no_duplicates_categorized_files`)?

fourshade:
I was just thinking about some tests...this file is so large that detailed tests could take a long time, but I can knock this one out with a runtime assertion that's basically free. I already see an assertion in this file, so I'm assuming that AssertionErrors are allowable at this stage of loading.

PR #1313, opened 2021-05-08
engine: don't try to start missing extensions

### LINKED ISSUE #1305: It's hard to remove extension plugins uninstalled with pip
**Describe the bug**

Title.

**To Reproduce**

1. Install some extension plugin.
2. Enable it.
3. Exit Plover.
4. Uninstall the plugin with pip.
5. On every Plover startup, an error message will be shown saying that it's not possible to start the extension plugin...

**Expected behavior**

Perhaps Plover should remove the plugin from the configuration the first time it detect that it isn't available?

**Operating system**

 - OS: Probably unrelated.
 - Plover Version: somewhat new.

**Hardware**

Unrelated.

benoit-pierre:
## Summary of changes

Ignore missing extensions.

Closes #1305.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1313.core.md
+++ b/news.d/bugfix/1313.core.md
@@ -1,0 +1,1 @@
+Don't try to start missing extensions.
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -276,6 +276,10 @@ class StenoEngine:
             log.info('starting `%s` extension', extension_name)
             try:
                 extension = registry.get_plugin('extension', extension_name).obj(self)
+            except KeyError:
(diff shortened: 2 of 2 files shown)


benoit-pierre:
Yes, but since we don't know if the fact that the extension is missing is temporary (maybe the user started with `--no-user-plugins`, or just updated to a version using a newer Python release), I don't see a better alternative.

user202729:
... I didn't know about that feature. Perhaps the script `dist_main` should modify the help content too.

morinted (approved):

PR #1333, opened 2021-05-23
gui_qt: show a message when hiding to tray

### LINKED ISSUE #1280: Show a message to indicate that Plover is still running when minimized to tray

Similar to #1279, when users close Plover to tray, they sometimes don't realize that it's still running. Some programs solve this by showing a message that says "Application is still running."

When you close Plover to the system tray for the first time, it should show a message like "Plover is still running."

After showing the message once, we should probably save that the event has happened and never show it again.

**Additional context**

https://doc.qt.io/qt-5/qsystemtrayicon.html#showMessage

benoit-pierre:
## Summary of changes

Show a message when hiding to tray (the first time only, and not when starting minimized).

Closes #1280.

### Pull Request Checklist
- [ ] Changes have tests


plover/gui_qt/trayicon.py
```diff
@@ -52,8 +58,8 @@ def log(self, level, message):
             else:
                 icon = QSystemTrayIcon.Critical
                 timeout = 25
-            self._trayicon.showMessage(__software_name__.capitalize(),
-                                       message, icon, timeout * 1000)
+            self.show_message(__software_name__.capitalize(),
```
user202729, line 61: ... doesn't look like this will run properly with the software name.

plover/gui_qt/main_window.py
```diff
@@ -143,6 +143,7 @@ def __init__(self, engine, use_qt_notifications):
             self.on_configure()
         # Apply configuration settings.
         config = self._engine.config
+        self._hidden_to_tray_message_shown = config['start_minimized']
```
user202729, line 146: Given what this do, the variable name is somewhat misleading. I'm thinking "[hidden to tray] message show next time" but that's quite long.

news.d/feature/1333.ui.md
```diff
@@ -0,0 +1 @@
+Show a message when hiding to tray the first time.
```
user202729, line 1: Should this be "the first time for each Plover launch" or something similar?

If it's truly "the first time" Plover should probably store some config option "show message on minimize" with options (I think) "never", "once", "once every launch", "always"

benoit-pierre:
Amended, thanks for the review @user202729.

PR #1343, opened 2021-06-07
Update english_stenotype.py

JorWat25:
## Summary of changes

Added 'in'' as a suffix to the orthography rules. May not be formatted completely right.

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/system/english_stenotype.py
+++ b/plover/system/english_stenotype.py
@@ -60,7 +60,7 @@ ORTHOGRAPHY_RULES = [
 
     # == y ==
     # die+ing = dying
-    (r'^(.+)ie \^ ing$', r'\1ying'),
+    (r'^(.+)ie \^ (ing|in\')$', r'\1y\2'),
     # metallurgy + ist = metallurgist
     (r'^(.+[cdfghlmnpr])y \^ ist$', r'\1ist'),
     # beauty + ful = beautiful (y -> i)
@@ -77,7 +77,7 @@ ORTHOGRAPHY_RULES = [
     (r'^(.+)y \^ (ial|ially)$', r'\1\2'),
     # == +if ==
     # spaghetti +ification = spaghettification (*spaghettiification)
-    (r'^(.+)i \^ if(y|ying|ied|ies|ication|ications)$', r'\1if\2'),
+    (r'^(.+)i \^ if(y|ying|yin\'|ied|ies|ication|ications)$', r'\1if\2'),
 
     # == +ical ==
     # fantastic +ical = fantastical (*fantasticcal)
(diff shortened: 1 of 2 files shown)


JorWat25:
The later changes might not actually be necessary, I now realize. But the first one is, as otherwise `TKAOEU/*G` writes 'diein'' rather than 'dyin''.

plover/system/english_stenotype.py
```diff
@@ -106,20 +106,20 @@
...
-    (r'^(.+)ar \^ is(e|ed|es|ing|er|ers|ation|ations|m|ms)$', r'\1aris\2'),
+    (r'^(.+)ar \^ iz(e|ed|es|ing|in\'|er|ers|ation|ations|m|ms)$', r'\1ariz\2'),
+    (r'^(.+)ar \^ is(e|ed|es|ing|in\'|er|ers|ation|ations|m|ms)$', r'\1aris\2'),
 
     # category +ize/+ise = categorize/categorise (*categoryize/*categoryise)
     # custom +izable/+isable = customizable/customisable (*custommizable/*custommisable)
     # fantasy +ize = fantasize (*fantasyize)
-    (r'^(.*[lmnty]) \^ iz(e|es|ing|ed|er|ers|ation|ations|m|ms|able|ability|abilities)$', r'\1iz\2'),
-    (r'^(.*[lmnty]) \^ is(e|es|ing|ed|er|ers|ation|ations|m|ms|able|ability|abilities)$', r'\1is\2'),
+    (r'^(.*[lmnty]) \^ iz(e|es|ing|in\'|ed|er|ers|ation|ations|m|ms|able|ability|abilities)$', r'\1iz\2'),
```
marnanel, line 121: "-izm" and "-izms" aren't really used, are they?

JorWat25: That was in the original file.

benoit-pierre:
Can you add a couple of tests here: https://github.com/openstenoproject/plover/blob/master/test/test_orthography.py#L9.

JorWat25:
> 
> 
> Can you add a couple of tests here: https://github.com/openstenoproject/plover/blob/master/test/test_orthography.py#L9.

Have added some tests.

benoit-pierre:
This need a news fragment (in `news.d/feature/1343.core.md`).

And you picked the wrong word (*interpret*) for testing (since *interpretin'* is not in the wordlist).

JorWat25:
I'm sorry, but I don't know enough about how this code works to do this properly. Apologies for testing the wrong thing, I just took some of the words in the current test, and replaced 'ing' with 'in''.

benoit-pierre:
The only reason `interpret+ing -> interpreting` works is because `interpreting` is in the wordlist, so without `interpretin'` being in the wordlist the orthography rules will yield `interprettin'`. Call it an irregular, that would need a dedicated dictionary entry.

So the question is: is the use case frequent enough, and are the orthography rules going to be right more often than not for this PR to be really useful.

user202729:
If there were something like `{delete_last_char}` then `{^in'}` could be implemented as `{^ing}{delete_last_char}{^'}`.

(that having said, `*G` is defined as `{^in'}` in the dictionary in the current version)

JorWat25:
> 
> 
> If there were something like `{delete_last_char}` then `{^in'}` could be implemented as `{^ing}{delete_last_char}{^'}`.
> 
> (that having said, `*G` is defined as `{^in'}` in the dictionary in the current version)
> 
> Adding all the `in'` words into the word list sounds bad.

There's {#BackSpace}, but that does have the problem of not being undoable.

benoit-pierre:
How about this instead:
```diff
 plover/assets/main.json            |  2 +-
 plover/system/english_stenotype.py |  3 +++
 test/test_blackbox.py              | 18 +++++++++++++++++-
 3 files changed, 21 insertions(+), 2 deletions(-)

diff --git i/plover/assets/main.json w/plover/assets/main.json
index be14ac0d..8561e4e8 100644
--- i/plover/assets/main.json
+++ w/plover/assets/main.json
@@ -138064,7 +138064,7 @@
 "*BS": "action",
 "*L": "{^le}",
 "*LT": "{^let}",
-"*G": "{^in'}",
+"*G": "{^ing}{^'}",
 "*T": "{^th}",
 "*TS": "it's",
 "*S": "{^s}",
diff --git i/plover/system/english_stenotype.py w/plover/system/english_stenotype.py
index ac1cb918..5e26981b 100644
--- i/plover/system/english_stenotype.py
+++ w/plover/system/english_stenotype.py
@@ -138,6 +138,9 @@
     # == misc ==
     # defer + ed = deferred (consonant doubling)   XXX monitor(stress not on last syllable)
     (r'^(.*(?:[bcdfghjklmnprstvwxyz]|qu)[aeiou])([bcdfgklmnprtvz]) \^ ([aeiouy].*)$', r'\1\2\2\3'),
+
+    # ing + ' = in'
+    (r"(.+)ing \^ '$", r"\1in'",),
 ]
 
 ORTHOGRAPHY_RULES_ALIASES = {
diff --git i/test/test_blackbox.py w/test/test_blackbox.py
index e0644f4d..3e4d3a0c 100644
--- i/test/test_blackbox.py
+++ w/test/test_blackbox.py
@@ -1809,7 +1809,7 @@ def test_bug_1448_1(self, with_korean_system):
         # trying to represent the `ㅎㅁㄱㅈㄴ-ㄴㅅㅈㅁ`
         # stroke with the `English Stenotype` system.
         r'''
-        'TEFT': 'test'
+        'TEFT': 'test',
         'TEFT/-G': 'testing'
 
         :system 'Korean Modern C'
@@ -1834,3 +1834,19 @@ def test_bug_1448_2(self, with_korean_system):
         :system 'English Stenotype'
         -G  ' 12345 -G'
         '''
+
+    def test_ing_contraction_1(self):
+        r'''
+        "TEFT": "test",
+        "*G": "{^ing}{^'}",
+
+        TEFT/*G  " testin'"
+        '''
+
+    def test_ing_contraction_2(self):
+        r'''
+        "TERP": "interpret",
+        "*G": "{^ing}{^'}",
+
+        TERP/*G  " interpretin'"
+        '''
```

This way, we take advantage of the existing rules / wordlist so we get the correct output, even for irregulars like “interpretin'”.

TheTabletopGirl:
A new solution to this has been offered and seems like it would be a very simple way to add `in'` endings. Instead of it's own rule, the `in'` ending simply adds `ing` (using the existing rules), backspaces one character and types `'`, effectively just replacing `g` with `'`.

The main.json dictionary itself seems to need a lot of updating and fixing, but this should be fine to add in the meantime.

PR #1344, opened 2021-06-07
Change links at the bottom of the webpage

antistic:
## Summary of changes

- Removes the two links "Download Latest Stable" and "Weekly Builds"
  - These links don't convey that the v4 version is what most people on the discord recommend
  - They link to the releases page and it's tempting to scroll down to the files instead of reading the text and clicking through to the installation guide. This isn't great because on Mac it's necessary to read this and do a few more steps.
  - The troubleshooting guide is even more hidden (you have to click to the installation guide first)
- Replaces with the beginner's guide
  - This is the same link as on the [open steno homepage](https://www.openstenoproject.org/).
  - It might make sense to also add the link to the community page, but since the section is titled "download" and I didn't want to change too much I left it at the one link

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/index.html
+++ b/index.html
@@ -66,9 +66,7 @@
           <h2>Download</h2>
           <p>Plover sees regular updates, and accepts bug reports, feature requests, art, and usability ideas. Please see <a target="_blank" href="https://github.com/openstenoproject/plover" alt="Plover's GitHub">Plover's GitHub</a> if you would like to contribute. Please follow below to see the latest releases and installation instructions.</p>
           <div class="ploverBtns">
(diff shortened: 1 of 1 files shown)


benoit-pierre (approved):

didoesdigital:
This is an excellent change! This will help reduce friction for a lot of beginners. It would be lovely to see it merged, is there anything else that needs to be done? I'd be happy to help out.

morinted:
Let's add both!

Does this language make sense to people?

image

morinted:
Maybe the text below download could be moved to a "Contribute" section and we can focus the Download section a little. Would you mind if I took a stab at this tomorrow?

PR #1357, opened 2021-06-18
 Do not consider extended scan codes in Windows

### LINKED ISSUE #1289: Some extra keys are being treated as steno keys
**Describe the bug**

When Plover is active on my laptop, the mute and volume down keys become W and A keys respectively. According one website, they are keys <181> and <182>.

**To Reproduce**

1. Activate Plover
2. Press the mute or volume down keys
3. See W or A

**Expected behavior**

Either nothing should happen, or the keys should do their usual function.

**Operating system**

 - OS: Windows 10
 - Plover Version 4.0.0.dev9 

**Hardware**

The laptop is an Lenovo Ideapad.

I should note it is only the laptop keys that do this. An external keyboard, also with volume keys apparently with the same code numbers, works normally when Plover is activated.

user202729:
## Summary of changes

(see linked issue.)

Closes #1289 .

### Pull Request Checklist
- [ ] Changes have tests


mkrnr:
Thanks a lot for this! I'll do some testing on Windows next days and look into these flags for a code review.

user202729:
Could it be that some other change between the time the pull request is opened and now happens to already fix the bug?

mkrnr:
Either that or it only affects some laptops. If it would affect all Windows laptops, we probably would have gotten more complainants over the years. So I'd be cool with merging it anyways.

Issue #672, opened 2017-01-24
Please allow normal key names with the number bar in strokes


JasonWoof:
Classification: Feature (New)

## Summary

I just made a dictionary with all the number strokes I understand, and it was super-weird that it didn't work when I listed they key names in steno order like this:

    "#S": "1"

Please support that. It would be really confusing to move the number key mapping without this feature, and also without it, it's confusing to make miss-strokes for w ords with the number bar.

morinted:
I think this used to be supported. I think to normalize, we'd have two options:

- Convert `#STKP` into `12K3`
- Convert `12K3` into `#STKP`

I think the second option bodes well for the ideal of de-magicifying steno. The number bar gets so much special treatment and maybe it shouldn't.

JasonWoof:
I think of this as primarily a interface issue. I think that the ``DUPT`` dialog should be tolerant of ``#STKP`` just as the other input box in that form is tolerant of extra white space.

Don't get me wrong... I love the second version. it's just that I don't want this little interface clean-up to be held back because of that being a big decision or something. Also, I think both forms (``#S`` and ``1``) should be supported in the ``DUPT`` form regardless of what the canonical form is.

JasonWoof:
Oops, I mean I like the first version (``#STKP`` or ``#S``)

morinted:
Yeah, but the same code is used when loading and saving the dictionary. You
want to ensure that looking up #1, #S, and 1 all return the same entry.

JasonWoof:
Yes, of course it should be canonicalized before saving or searching. I'm suggesting that a filter be used that converts the value in the DUPT dialog box to the canonical form before it's used (saved in this case) and other forms (like lookup?) that allow you to enter a stroke should probably have the same filter.

morinted:
It should already work that way, if I'm understanding you correctly. It's
just that #S isn't being handled properly.

JasonWoof:
Cool, sound like we are on the same page, I wasn't sure before.

thumbquat:
Somewhat related, I was on the discord and someone said:

>are there any settings or plugins to show steno chords with the number bar as # plus the letters instead of showing numbers (in the Suggestions Window, Lookup, and Add Translation boxes and such)? Like #TA instead of 25

Made a quick branch here a few days ago:
[features/number-translations](https://github.com/thumbquat/plover/commit/44f5081c36bbd80d7408ed2d3ccd627887e3c428)

[...]

PR #1362, opened 2021-06-21
use `plover_stroke`

benoit-pierre:
## Summary of changes

The next version of `plover_stroke` (1.0.0) has been rewritten to use a CPython extension, making it possible to use it for the `steno` module implementation without negatively impacting the performance (particularly loading times).

Pros:
- one consistent implementation: same behavior (including normalization) for `steno.Stroke`, `normalize_steno`, and `plover_build_utils.testing.steno_to_stroke`
- better steno normalization (see tests), including detection of invalid strokes (used by the "add translation" dialog and dictionary editor to prevent adding invalid mappings)
- new `steno_to_sort_key` for sorting on steno order (used by the dictionary editor)
- slightly faster loading times

Cons:
- `plover_stroke` support a maximum of 63 keys
- using a CPython extension means no universal wheel for `plover_stroke` (but binary wheels for common platforms / Python versions are provided)
- sorting on steno order is slower than the standard string sort

TODO:
- [ ] use `steno_to_sort_key` for sorting on JSON dictionary save 
- [ ] use the framework added in #1332 do add a test for `gui_qt.steno_validator`

### Pull Request Checklist
- [x] Changes have tests


benoit-pierre:
I think this take care of #672.

Issue #1128, opened 2020-05-24
RTF dictionary bug with undoable new lines and returns?


ms009986:
## Summary

I just found out about undoable line breaks (\n, \r, \t, etc.). Up to now I have just been grumbling if I had to undo more than a whole line of steno.  So I was all excited and all set to replace all my {#Return}s in my RTF dictionaries with {\r}s.

So I started with one just make sure I knew what I was doing.  Apparently I don't.   I used Plover to define the stroke SPH-PBG/SPH-PBG (something something) as {\r}.  I then stroked something something in Word, but nothing happened. (I mean, what actually happened was the word "something" appeared and then disappeared as I stroked, but no _new line_ happened.)

So I tried using Plover to define the SPH-PBG/SPH-PBG stroke as just \r, then tried that in Word. And it worked.  If I went to my NEW ENTRIES dictionary in Plover, I saw that SPH-PBG/SPH-PBG was defined as \r.  And as I stenoed into Word, it did in fact add a new line.  

HOWEVER.  I looked at the entry in my RTF dictionary and it looked like this:

{\*\cxs SPH-PBG/SPH-PBG}

Nothing to the right of it.  I wasn't very clear how this empty entry was creating a new line in Word. Then I figured Plover must create a single working dictionary to refer to, as well as update all the loaded dictionaries when a change is made.  Or something like that, makes sense.

So I closed Plover and reopened it.  The RTF entry was still the same, and now Plover did nothing when I stroked it into Word.  And Plover showed the entry as defined as (empty field).

Long story short, I tried manually editing the RTF dictionary with the sorts of things I'm used to seeing in my RTF dictionaries (like putting a \ in front of the \r, putting {} around the entry, using \n instead of \r, etc. etc.), but nothing seemed to work.  In fact defining it in Plover as described in the dictionary explanation online as {\r} made the RTF entry this:

{\*\cxs SPH-PBG/SPH-PBG}{
}

(Note the literal new line.  :-)}

So, 1) This is apparently a bug with RTF dictionaries.

And 2), any ideas on what the syntax should be in an RTF to make it work?  I assumed it would come out in RTF as \{\\r\} or something instead of just coming out blank?

Thanks for your input!  I'm happy to clarify if I wasn't clear!  Love Plover!

## Plover Version

Possibly relevant information:
Plover 4.0.0.dev8+8.ge062cda
Windows 10
Stentura Protégé
Phoenix Theory
RTF dictionaries, not JSON

[...]

PR #1364, opened 2021-06-22
rework RTF/CRE code

### LINKED ISSUE #209: spacing bug when loading from RTF dictionary

benoit-pierre:
## Summary of changes

New parser:
- don't rely on regular expressions for parsing (only for tokenizing).
- correctly handle multi-lines mappings
- detect syntax errors (with recovery)
- use `\n\n` for new paragraphs (instead of non-undoable `{#Return}{#Returns}`)
- similarly, use `\t` and `\n` for `\tab` and `\line`
- faster than previous code (~1.8 faster at loading a 46.5 MB dictionary, ~2.4 when combined with #1362)

New save code:
- process each formatting atom separately
- correctly escape `{}\`
- use custom ignored groups for Plover macros and metas
- use groups to improve round-tripping affixes (so there's no ambiguity when parsing back, e.g. `{^in^}fix` -> `{\cxds in \cxds}fix` instead of `\cxds in\cxds fix`)

Closes #209, #589, #773, #1128.

### Pull Request Checklist
- [x] Changes have tests


test/test_rtfcre_dict.py
```diff
@@ -5,24 +5,67 @@
...
+    lambda: ('test\\nsomething', r'test\\nsomething'),
+    # Handling tabs.
+    lambda: ('test\tsomething', r'test\tab something'),
+    lambda: ('test\\tsomething', r'test\\tsomething'),
+    # Infix with force cap.
+    lambda: ('{^\n^}{-|}', r'{\cxds \line \cxds}\cxfc '),
+    lambda: ('{^\n\n^}{-|}', r'{\cxds \par \cxds}\cxfc '),
+    # Plover custom formatting:
+    # - meta: command
+    lambda: ('{PLOVER:TOGGLE}', r'{\*\cxplovermeta PLOVER:TOGGLE}'),
```
sammdot, line 43: `\cxplovercommand` maybe? and the `PLOVER:` is redundant here.

test/test_rtfcre_dict.py
```diff
@@ -5,24 +5,67 @@
...
+    # - meta: key combo
+    lambda: ('{#Return}', r'{\*\cxplovermeta #Return}'),
```
sammdot, line 45: `\cxploverkeycombo`

plover/dictionary/rtfcre_dict.py
```diff
@@ -13,326 +13,134 @@
...
-        def handler(s, pos):
-            for handler in handlers:
-                result = handler(s, pos)
-                if result:
-                    return result
+from .rtfcre_parse import parse_rtfcre
+
+
+HEADER = (r'{\rtf1\ansi{\*\cxrev100}'
+          r'\cxdict{\*\cxsystem Plover}'
```
sammdot, line 27: I think the `\cxsystem` here should include the version number

test/test_rtfcre_dict.py
benoit-pierre: No, the goal is to avoid the need for a specific control word for each feature supported by metas, so only one general `\cxplovermeta` control word, with the content.

test/test_rtfcre_dict.py
benoit-pierre: Ditto.

plover/dictionary/rtfcre_dict.py
```diff
@@ -13,326 +13,134 @@
...
+        (r'{\?}'                       , r'{{\cxp? }}'                 ),
+        (r'{\,}'                       , r'{{\cxp, }}'                 ),
+        (r'{:}'                        , r'{{\cxp: }}'                 ),
+        (r'{;}'                        , r'{{\cxp; }}'                 ),
+        (r'{\^\^?}'                    , r'{{\cxds}}'                  ),
+        (r'{\^([^^}]*)\^}'             , r'{{\cxds {0}\cxds}}'         ),
+        (r'{\^([^^}]*)}'               , r'{{\cxds {0}}}'              ),
+        (r'{([^^}]*)\^}'               , r'{{{0}\cxds}}'               ),
+        (r'{-\|}'                      , r'\cxfc '                     ),
+        (r'{>}'                        , r'\cxfls '                    ),
```
sammdot, line 87: `\cxfl`, I don't see a `\cxfls` in the spec.

benoit-pierre: Good catch.

plover/dictionary/rtfcre_dict.py
benoit-pierre: Done.

benoit-pierre:
I forgot to mention there's one xfailed test left:
```python
    # Why not 'eclipse command'?
    lambda: rtf_load_test('{eclipse command}', '{eclipse command}', xfail=True),
```
Which has the comment mention does not make sense: `eclipse command` it the logical output.

user202729:
Looks good. (although I don't really use RTF)

Notes:

* If you don't need to support Python <=3.3, use `fullmatch` instead of `match` and add `$` may be clearer.
* (possible improvement) Add some code to make RTF->Plover round-trip better for unsupported group (for example `{plover:rtf_unknown:\cxsvatdictentrydate\yr2006\mo5\dy10}`)
* It isn't yet possible to do rtf -> Plover -> rtf (for  `{^-^}` -> `\_`.), but if/when it's possible to do Plover -> rtf -> Plover then it may make sense to also test round-trip in the `test_format_translation`.

benoit-pierre:
>     * If you don't need to support Python <=3.3, use `fullmatch` instead of `match` and add `$` may be clearer.

Good point, done.

>     * (possible improvement) Add some code to make RTF->Plover round-trip better for unsupported group (for example `{plover:rtf_unknown:\cxsvatdictentrydate\yr2006\mo5\dy10}`)

Yeah, no, saving in Plover is always going to be a destructive operation: between the unsupported control groups/words, the fact that they can stack, the mostly ignored header part of the RTF (stylesheet...).

>     * It isn't yet possible to do rtf -> Plover -> rtf (for  `{^-^}` -> `\_`.), but if/when it's possible to do Plover -> rtf -> Plover then it may make sense to also test round-trip in the `test_format_translation`.

I'll see if I can merge those tests with the save ones (which include a round-trip check).

benoit-pierre:
> > * It isn't yet possible to do rtf -> Plover -> rtf (for  `{^-^}` -> `\_`.), but if/when it's possible to do Plover -> rtf -> Plover then it may make sense to also test round-trip in the `test_format_translation`.
> > ```
> 
> I'll see if I can merge those tests with the save ones (which include a round-trip check).

Perfect roundtrip is not always possible, e.g.:  `{*!}` -> `{\*\cxplovermacro retrospective_delete_space}` -> `=retrospective_delete_space`.

And there are corresponding load tests for checking the conversion in the other direction, so I'm gonna leave it as is, except I added support for `{^ ^}` -> `\~` and `{^-^}` -> `\_`.

PR #1365, opened 2021-06-22
RTF/CRE rework + small C extension to tokenize RTF

benoit-pierre:
## Summary of changes

This add a C extension to tokenize RTF (the parsing is still done in Python), this would bring the speed gains (with #1362, same dictionary) to ~5.3.

The problem: no more universal wheel for Plover! No universal wheel, mean installing from source unless a wheel for the correct platform / interpreter version is available, which means PyQt5 must be installed for building the UI. This would make installing Plover more difficult on things like a Raspberry Pi.

Possible fix: move the RTF/CRE support to a dedicated external plugin, like we did with Treal.


user202729:
There remains an option to provide a pure Python plugin, and a C extension plugin. But that's more maintenance work.

Otherwise, I assume that you would remove the Python implementation (before the import)?

benoit-pierre:
Or just make a small standalone `rtf_tokenize` package and depend on that (like for `plover_stroke` in #1362).

PR #1375, opened 2021-07-11
steno_dictionary: drop support for longest key callbacks

### LINKED ISSUE #1374: Dictionary is not deleted when changed and refresh?
**Describe the bug**

(might be a problem with my plugin. Not very

**To Reproduce**

1. Find some large dictionary file.
2. Touch the file or edit it.
3. Refresh Plover.
4. See that memory consumption is increased.
5. Repeat step 2-4 several times.

**Expected behavior**

The memory consumption should not increase, or at least until the garbage collector is run.

**Operating system**

 - OS: Linux
 - Plover Version: Latest.

**Hardware**

Doesn't matter.

benoit-pierre:
## Summary of changes

This simplifies the code and avoids circular references.

Closes #1374.

### Pull Request Checklist
- [x] Changes have tests


plover_build_utils/testing/steno_dictionary.py
```diff
@@ -207,28 +207,19 @@ def test_longest_key(self):
         Check `longest_key` support (including callbacks handling).
```
user202729, line 207:
```suggestion
        Check `longest_key` support.
```

Issue #1381, opened 2021-07-22
Toggling a dictionary while there's a nonexistent dictionary will make it impossible to exit the add translation dialog


user202729:
**Describe the bug**

(title)

(note: I knew this bug long ago, but previously I only get problem with the `plover_websocket_server` plugin, which uses private fields and thus not really supported)

**To Reproduce**

1. Add a dictionary to Plover.
3. Delete that dictionary, and reload Plover. Plover should show an error about the dictionary being nonexistent.
1. Add some stroke `A` to toggle any dictionary. Needs `plover-dict-commands` plugin.
2. Open the add translation dialog, focus the "Strokes:" field.
3. Stroke `A`. (the "no such file or directory" error is shown, and internally the filter is removed because the dictionary collection is reset)
3. Try to close the dialog.
4. See error. (the program tries to remove the filter again, which causes the error)

**Expected behavior**

No error, and the dialog closes.

**Operating system**

 - OS: Linux, but doesn't matter.
 - Plover Version: ~~latest master~~ v4.0.0.dev9-71-g0501aa3 .

PR #1388, opened 2021-07-25
Do not remove filters on dictionaries reload

user202729:
## Summary of changes

Closes https://github.com/openstenoproject/plover/issues/1381 .

*(is the approach okay? I don't think there's any other instance in Plover code that the filter feature is used)*

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1388.core.md
+++ b/news.d/bugfix/1388.core.md
@@ -1,0 +1,1 @@
+Do not discard existing filters on dictionaries reload.
--- a/plover/engine.py
+++ b/plover/engine.py
@@ -179,9 +179,8 @@ class StenoEngine:
         if not dictionaries_changed(dictionaries, self._dictionaries.dicts):
             # No change.
             return
-        self._dictionaries = StenoDictionaryCollection(dictionaries)
-        self._translator.set_dictionary(self._dictionaries)
-        self._trigger_hook('dictionaries_loaded', self._dictionaries)
+        self._dictionaries.set_dicts(dictionaries)
+        self._trigger_hook('dictionaries_loaded', StenoDictionaryCollection(dictionaries))
 
     def _update(self, config_update=None, full=False, reset_machine=False):
         original_config = self._config.as_dict()


user202729:
I'm not sure if the fail is a problem of the test or the code. Does the hook expect that the objects will never have the dicts modified?

benoit-pierre:
Yes, just:
```diff
 plover/engine.py | 5 ++---
 1 file changed, 2 insertions(+), 3 deletions(-)

diff --git i/plover/engine.py w/plover/engine.py
index 1b77dba8..21489a1d 100644
--- i/plover/engine.py
+++ w/plover/engine.py
@@ -175,9 +175,8 @@ def dictionaries_changed(l1, l2):
         if not dictionaries_changed(dictionaries, self._dictionaries.dicts):
             # No change.
             return
-        self._dictionaries = StenoDictionaryCollection(dictionaries)
-        self._translator.set_dictionary(self._dictionaries)
-        self._trigger_hook('dictionaries_loaded', self._dictionaries)
+        self._dictionaries.set_dicts(dictionaries)
+        self._trigger_hook('dictionaries_loaded', StenoDictionaryCollection(dictionaries))
 
     def _update(self, config_update=None, full=False, reset_machine=False):
         original_config = self._config.as_dict()
```
and leave the test alone.

user202729:
On the other hand, if it's done like that the hook argument will not have the filters set.

Perhaps the safest way would be...

```python
        old_filters = self._dictionaries.filters
        self._dictionaries = StenoDictionaryCollection(dictionaries)
        self._dictionaries.filters = old_filters
        self._translator.set_dictionary(self._dictionaries)
```

it's mainly annoying because `StenoDictionaryCollection` init doesn't take `filters` as input. Make it an optional argument is also an option.

------

Another option is to make `dicts` completely immutable, and remove `set_dicts` method entirely (while leaving `add_filter` / `remove_filter` intact)

benoit-pierre:
We don't want access to the dictionaries through the hook arguments, so ideally, we just notify a collection of steno-dict like object, with just the right information for monitoring the status of the dictionaries (names / paths, read-only state, wether each dictionary was successfully loaded or not, ...).

benoit-pierre (changes requested):
This just need a news entry.

user202729:
Done (not sure if the wording is clear enough however)

benoit-pierre (approved):

PR #1425, opened 2021-10-17
beyond PyQt5

benoit-pierre:
## Summary of changes

With the Qt5 LTS support moved to [commercial only](https://www.phoronix.com/scan.php?page=news_item&px=Qt-5.15-LTS-Commercial-Phase), the only maintained version seems to be the KDE version, which is shipped by some distributions.

The first Q6 LTS release is also [available](https://www.phoronix.com/scan.php?page=news_item&px=Qt-6.2-LTS-Released), and there are now alternatives to PyQt5 / PyQt6: the official [PySide2](https://pypi.org/project/PySide2/) / [PySide6](https://pypi.org/project/PySide6/#description) bindings.

The 4 sets of bindings are mostly compatible.

By switching to [QtPy](https://pypi.org/project/QtPy/), an abstraction layer for PyQt5/PyQt4/PySide2/PySide, and with a few compatibility fixes, we can target PyQt5, PySide2, PyQt6 and PySide6, making the transition to another set of bindings easier (and transparent for plugins).

Main changes:
- import from `qtpy`
- don't use `QVariant` when declaring signals (just use `object`).
- use scoped enums accesses: e.g. `QDialogButtonBox.StandardButton.Ok` instead of `QDialogButtonBox.Ok`. Necessary for Qt6 compatibility, increases our minimum supported version of PyQt5 to 5.11 (released in 2018).
- ~switch to the official `uic` and `rcc` helpers (provided by [qt5-applications](https://pypi.org/project/qt5-applications/)) for generating UI files, including resources, as PyQt6 drops support for `pyrcc`.~
- load resources directly from filesystem (no need for `pyrcc5` anymore).
- use the `QComboBox` `textActivated(QString)` signal instead of `activated(QString)`. Necessary for Qt6 compatibility, increases our minimum supported version of PyQt5 to 5.14 (released in 2019).
- a few other minor compatibility fixes for supporting PySide2 / PySide6 / PyQt6 / Qt6.

Known issues:
- ~the `qt5-applications` Windows' wheels are missing the `rcc` executable (reported [here](https://github.com/altendky/qt-applications/issues/21)).~
- ~the official `uic` Python generator may generate invalid strings for some Unicode characters (like for [this one in the `plover_wpm_meter` plugin](https://github.com/arxanas/plover_wpm_meter/blob/master/plover_wpm_meter/wpm_meter.ui#L113)).~
- ~the dictionaries widget checkboxes don't work correctly with PyQt6 / PySide6~

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,18 +35,21 @@ jobs:
       - name: Calculate skip cache keys
         id: set_cache
         run: |
-          job_cache_extra_deps=(); job_id=test_linux; job_name='Test (Linux)'; job_needs=(); job_os=Linux; job_platform=ubuntu-18.04; job_python=3.9; job_reqs=(reqs/dist.txt reqs/test.txt); job_skip_cache_name=skip_test_linux_py-3.9_ubuntu-18.04; job_skip_cache_path=.skip_cache_test_linux; job_skiplists=(job_test os_linux); job_test_args='-p no:pytest-qt --ignore=test/gui_qt'; job_type=test; job_variant=Linux; analyze_set_job_skip_cache_key
-          job_cache_extra_deps=(osx/deps.sh); job_id=test_macos; job_name='Test (macOS)'; job_needs=(); job_os=macOS; job_platform=macos-10.15; job_python=3.9; job_reqs=(reqs/dist.txt reqs/test.txt); job_skip_cache_name=skip_test_macos_py-3.9_macos-10.15; job_skip_cache_path=.skip_cache_test_macos; job_skiplists=(job_test os_macos); job_test_args='-p no:pytest-qt --ignore=test/gui_qt'; job_type=test; job_variant=macOS; analyze_set_job_skip_cache_key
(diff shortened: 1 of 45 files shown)


benoit-pierre:
OK, so I took another stab at this: by loading resources directly from the filesystem, we avoid the need for mucking around with `pyrcc5` and the like. This in turn mean we can keep using `pyuic5` for generating UI files (avoiding any issues with `qt5-applications`).

I noticed a new issue with PySide6, however: for some reason the dictionaries widget checkboxes don't work correctly (no check-mark for enabled dictionaries)...

benoit-pierre:
OK, the issue with the checkboxes is fixed, and I enabled / fixed the GUI tests for each supported variant (PyQt5, PyQt6, PySide2, and PySide6). Note that I had to update the platform to `ubuntu-20.04` for PyQt6 / PySide6, which means we'd mean we'd have to do the same for the Linux build (and loose support for some older distributions) if we do switch to Qt6.

PR #1455, opened 2021-12-15
fix QT GUI tests

benoit-pierre:
## Summary of changes

* fix tox lightweight tests environment so the QT GUI tests can be run
* add `mock` to the tests dependencies: the standard library version provided by Python <= 3.7 is too old


benoit-pierre:
Windows did not like that branch name...

Issue #1050, opened 2019-02-23
Not working with Gnome on Wayland in Linux


mayanksuman:
## Summary

Plover (latest github version) is not working on Gnome on Wayland.

## Reproducing

1. Use Gnome in wayland mode (default mode)
1. Navigate to plover git directory and launch it by `.\launch.sh`
1. Enable the output. However, it is not enabled.

The output of `./launch.sh` might be useful.
```
running launch
running build_py
running build_ui
running egg_info
writing plover.egg-info/PKG-INFO
writing dependency_links to plover.egg-info/dependency_links.txt
writing entry points to plover.egg-info/entry_points.txt
writing requirements to plover.egg-info/requires.txt
writing top-level names to plover.egg-info/top_level.txt
reading manifest template 'MANIFEST.in'
warning: no files found matching 'tox.ini'
writing manifest file 'plover.egg-info/SOURCES.txt'
generating plover/gui_qt/resources_rc.py
running compile_catalog
compiling catalog plover/gui_qt/messages/fr/LC_MESSAGES/plover.po to plover/gui_qt/messages/fr/LC_MESSAGES/plover.mo
copying plover/gui_qt/resources_rc.py -> build/lib/plover/gui_qt
running build_ext
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
xkbcommon: ERROR: Key "<CAPS>" added to modifier map for multiple modifiers; Using Control, ignoring Lock
```

## Plover Version

Latest from github (Plover 4.0.0.dev8; commit: 2943a976f1d3ec)

## System

OS: Debian Linux (kernel version: 4.19.0-2-amd64)
WM: Gnome (3.30.2)

benoit-pierre:
Wayland is not supported. Not sure if it provides everything needed for Plover to even support it.

hflw:
Has anyone ever looked into implementing plover as an input method engine? Idk if it's feasible at all but that's how most of the asian language input systems are implemented.

https://blogs.gnome.org/happyaron/2011/01/15/linux-input-method-brief-summary/

antoniusf:
Hey everyone,

I've recently been persuaded to start using sway, which is also a wayland-based window manager, so I've started to look into this issue a bit more. I thought it might be a good idea to share what I've learnt so far, so that other people can pitch in without having to start all the way from the beginning.

## Quick recap: what is Wayland?

[...]

Issue #650, opened 2016-11-30
XKB keyboard remapping affects output


nimble0:
Classification: Bug

Reproducibility: Always

## Summary

Emulated keystrokes are affected by XKB keyboard layout remapping.

## Steps to Reproduce

1. Set XKB keyboard layout to Colemak.
2. Activate Plover.
3. Press -T

## Expected Results

Output string "the".

## Actual Results

Output string "ghf".

## Version

Plover version 4.0.0.dev0

First appeared in commit 2419313b09c50d41efbb68bfdc8bead2c799771a.

Installed via: Git

## Notes

The cause seems to be in _send_keycode, specifically the change from target_window.send_event to xtest.fake_input.

Issue https://github.com/openstenoproject/plover/issues/298 is similiar but this bug appeared later.

Standard Qwerty layout behaves correctly.

## Configuration

OS: Linux Mint 17, KDE 4.13.3

benoit-pierre:
Plover does not handle keyboard layout changes on Linux (and never has). The layout is parsed one time at initialization. The XKB code provided by `python-xlib` is also a bit of a misnomer, because it does not use the XKB extension, so Plover also does not know anything about XKB groups (and group changes).

nimble0:
The layout isn't changed while Plover is running (I've changed the steps so that's clearer).

Looking at this more closely, there seems to be a global keyboard layout and some more local layout. Prior to that commit, the local layout didn't have any affect on Plover's output. After that commit though, it does affect the Plover's output but Plover is only aware of the global keyboard layout, so whenever the layouts are different, there's a problem.

The global keyboard layout can be changed with setxkbmap. The local one can be changed with KDE System Settings, but setxkbmap also changes it.

I suppose injecting the events into the window skips that local layout remapping.

benoit-pierre:
FWIU, there's a global server layout. You can also have per device layouts (including one for the fake XTest keyboard device). And then a layout can have XKB groups, which can be switched per application. Usually, DEs will configure a global layout with multiple XKB groups (one per language configured), and change group per application. Plover only has access to the equivalent of `xmodmap -pke`, and so is not group aware and behave as if the first group was always selected. It also does handle the fact that different devices can have different layouts, so if for example capslock/backspace are  swapped on one keyboard and not on the other, things will get weird when you switch keyboard.

nimble0:
That makes sense and is quite helpful to know. I do find it a bit strange there was no problem before.

hbaughman:
[...]

PR #1461, opened 2022-01-02
Add support for wlroots-based Wayland compositors like Sway

matteodelabre:
### Compatibility and testing

This PR adds keyboard emulation and capture support for some Wayland compositors using the `virtual_keyboard_unstable_v1` and `input_method_unstable_v2` protocols. As their names state, these protocols are not yet stable and may be changed in the future. Such changes would probably require changes in this implementation.

The required protocols are currently only implemented in wlroots-based compositors, such as Sway ([list of wlroots-based compositors](https://github.com/solarkraft/awesome-wlroots#compositors)). GNOME [has stated they don’t plan on implementing them](https://gitlab.gnome.org/GNOME/mutter/-/issues/1974) (but have alternatives in mind), so it’s not covered by this PR. When Plover is started on an unsupported Wayland compositor, it will display an error message stating this incompatibility.

I have tested this PR on Sway (sway version 1.6.1, wlroots version 0.14.1), with TX Bolt and Keyboard machines (and an otherwise default Plover configuration).

### Implementation details

This adds a new dependency on the [`pywayland`](https://github.com/flacjacket/pywayland) library, used to communicate with the compositor. Wayland protocols are defined as XML files and require a code generation step to be available from Python. I checked out the XML definitions, and added a build step to generate the required code (using a similar process as used to generate Qt code).

### References

- Related issue: #1050
  (thanks @antoniusf for some very useful pointers in that thread!)
- This implementation is not affected by #650

### Pull Request Checklist
- [ ] ~~Changes have tests~~ (is there a simple way to test these changes automatically?)


antoniusf:
Thank you so much for implementing this! I haven't looked at things in detail yet, but the implementation looks nice on first glance. I've only tested this briefly but so far everything worked nicely with both the keyboard machine and an external machine (dnaq's HID protocol), and that's with a non-US keyboard layout. (Also on Sway 1.6.1.)

It's unfortunate that this leaves out gnome, but it seems like we'll need a separate implementation to support them well anyways; for compositors that do support these two protocols, all things considered, this is probably the optimal solution. I'm very happy I can finally retire my temporary workaround and replace it with this.

---

A few notes: This might be due to my build environment, but I get a

```
ModuleNotFoundError: No module named 'plover.oslayer.wayland'
```

when trying to do a clean build. The workaround for me was to add these four lines to my setup.cfg packages list, and create the corresponding empty directories before building:

```
    plover.oslayer.wayland
    plover.oslayer.wayland.wayland
    plover.oslayer.wayland.input_method_unstable_v2
    plover.oslayer.wayland.virtual_keyboard_unstable_v1
```

I don't know enough about python packaging to say what the actual proper solution is here, but I wanted to bring it up in case you do.

matteodelabre:
Thank you for testing! I had only tested running with `tox -e launch`, so thanks for bringing that point up. I’ve added some changes to try to fix the build. With the latest changes:

- Using [the build command that’s used to generate the AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=plover-git#n52) works.
- Activating the dev environment with `source .tox/dev/bin/activate` and running `python setup.py bdist_wheel` works.
- Using `tox -e setup -- bdist_wheel` fails (it looks like it’s trying to run some commands before the Wayland module generation instructions are reached, and those commands expect those modules to already exist).

I assume the maintainers would have more insights than me regarding this.

antoniusf:
Thanks for the fix! The modified nix derivation I used for building now works as well.

mtoohey31:
First off, thanks so much for this PR, I've been wanting to get Plover working on Sway for forever! I was testing this out and I ran into some strange behaviour, it seems isolated to XWayland applications though. From what I can tell, keystrokes result in a lot of backspaces and an occasional number after those backspaces. My setup is a little non-standard though, so it may be a bug elsewhere, I'm running arch with:

```
sway-git r6855.dbaf2e4f-1
wlroots-eglstreams-git 0.16.0.r5319.f4578db2-1
xorg-xwayland 21.1.4-1
```

Has anyone else run into similar behaviour with XWayland applications?

matteodelabre:
Thank you for testing this! I’m also able to reproduce this issue on XWayland apps. I’ll take a look and see if I can fix this.

matteodelabre:
Related issue in a project that uses the same virtual keyboard protocol as this implementation: https://github.com/atx/wtype/issues/1

matteodelabre:
@mtoohey31 Should be fixed with this latest commit. I tested it on my end by running `GDK_BACKEND=x11 gedit`.

mtoohey31:
Can confirm, I just double checked with Brave Browser on my setup and it's fixed. Thanks!

TravisDavis-ops:
Thank you

BryanJacobs:
You can see the comment linked above, but I was asked to repost this here:

I tried out this PR as-is, but ran into serious problems in Slack (an Electron app) on both Wayland and X11 - backspaces not working, extraneous 'e' characters, etc. After doing some debugging, I discovered the issues only arise when sending a new keymap with every emitted string (as the PR does). At first I thought it was a timing issue, but no amount of slowing down the output rate seemed to help...

I made a few changes to cache a "default" keymap with the anglocentric usual characters (most of the output characters from the default Plover dictionary). I left fallback code in place to do what this PR was doing before.

With this change the Wayland support works flawlessly for me including in Electron apps. I think it also would make Plover behave like a more normal Wayland app - how many applications use two-plus new keymaps per second in the wild? If the lack of non-English characters is an issue, this patch could be modified to build a keymap that includes all the characters seen in any emitted string so far, or even all the characters in the configured dictionaries.

[bump-pywayland.txt](https://github.com/openstenoproject/plover/files/8074929/bump-pywayland.txt)
[cache-kb-mapping.txt](https://github.com/openstenoproject/plover/files/8074931/cache-kb-mapping.txt)

antoniusf:
thank you! :)

matteodelabre:
Hi @BryanJacobs, thanks for the patches!

> I tried out this PR as-is, but ran into serious problems in Slack (an Electron app) on both Wayland and X11 - backspaces not working, extraneous 'e' characters, etc.

I’m having trouble reproducing the exact behavior you’re describing. I’ve tested Slack and Discord and they work _mostly_ fine; sometimes the first word I type doesn’t use the right keymap, but no extraneous 'e's and the '*' key works correctly. Do you have any particular settings that I should enable on my end, or dictionaries/strokes that I should test, to reproduce your issue?

> I made a few changes to cache a "default" keymap with the anglocentric usual characters.. (…) With this change the Wayland support works flawlessly for me including in Electron apps. I think it also would make Plover behave like a more normal Wayland app - how many applications use two-plus new keymaps per second in the wild?

Interesting! Unless we manage to reproduce this bug in non-Electron apps (for example, does this happen in Chromium too?), it does seem like a bug that would be worth reporting to Electron folks. I do agree that it’s a bit unusual to regenerate the keymap dynamically at such a high rate. It’s allowed in the Wayland protocol though, and seems to work fine with most apps.

> If the lack of non-English characters is an issue, this patch could be modified to build a keymap that includes all the characters seen in any emitted string so far, or even all the characters in the configured dictionaries.

It would sure be great to keep this PR language-agnostic, if possible. Your suggestion to build an incremental keymap is interesting, we need to be careful though as there is a limit on the number of characters in an XKB keymap, and sending very large keymaps could hang the compositor.

BryanJacobs:
Versions:

System: Arch Linux
- Sway 1.7-2
- Wlroots 0.15.1-2
- Slack-wayland AUR 4.23.0
- Plover from your exact branch
- Machine: Gemini PR via serial (emulated from a QMK keyboard)

With these versions the problem can be immediately reproduced. I'm sure if you were to launch with the original strategy in here some others would have the same issues I did.

By the way, I was able to produce the issues (extra 'e' characters) in GDK_BACKEND="x11" gedit as well, so it's not specific to Electron 100%. An easy case that shows problems every time is stroking TEUPL/-G/* ("timing", then an undo). For me, I get "time", "timening", then "timeeeee" ('e' characters being emitted instead of backspaces). I'm also unable to make R-R send carriage returns properly in this mode. They do nothing. Suffix strokes that don't involve backspaces work fine; suffix strokes that do involve backspaces don't function properly, likely because in the base PR here they try to send two keymaps in extremely rapid succession (one for backspacing and one for typing).

Note that my patch uses a more recent version of the Pywayland binding than your original PR.

On another note, I believe the server allows keymap sizes up to 2GB and two billion characters, so it's rather unlikely that pre-sending a keymap that contains every output character in the user dictionary would be a practical problem. You wouldn't need to build an incremental keymap, you'd just need to hook into the dictionary update event (and startup), scan the dicts, and send the union set of every output character present. For most users that would be a one to two hundred characters maximum, and for a select few it would be a small number of thousands. This would also improve overall application performance by avoiding the need for a keymap event on every stroke - keeping in mind that fast stenographers produce over six strokes per second.

DarkKirb:
thanks for the patches! They seem to work in ail programs I write in, except for discord, where most strokes are changed to numbers (Plover becomes 21l3v54 for example)

benoit-pierre:
Forgot to mention known issues / limitations:
- because in Wayland key repeat is handled manually by each client, the keyboard capture cannot correctly handle those repeat events (e.g. to suppress a heavy-handed stroke mapped to `{PLOVER:RESUME}`).
- contrary to the X11 version, keyboard input is always grabbed (when the keyboard machine is used): so you better not `kill -STOP` Plover, or you'll completely loose keyboard input...
- similarly, I've tried to ensure the grab code is robust, and replay unsuppressed key events even in the event of an exception.

antoniusf:
i'm sorry for the late response, i haven't been able to use a computer for the past few weeks due to health reasons. i'm still not fully recovered and can't take a closer look at your changes, but i did at least want to write a small response because no one has said anything yet and i'm very grateful you took the time to look at this in detail and make a significant amount of improvements.

i can't speak on the issues you've mentioned, but they do seem sensible and i look forward to being able to take a closer look at the fixes you've implemented. as for the suggestion of making this an output plugin, i am very much in favor, precisely because of the multitude of approaches that is currently required to cover all wayland-based compositors, and especially considering all of the amazing work that @user202729 has already done the this area. i'd offer my help for this, but i doubt i could be of much use even if i was fully capable of using a computer right now.

antoniusf:
so, um, I did actually manage to test things out today, and I think there's a problem with the packaging?

```
2022-03-30 23:18:33,646 [MainThread] ERROR: error loading gui plugin: none (from plover.gui_none.main)
...
  File "/nix/store/3ln8ivqvjvk9zd2lfdwkmdrlh60wclaj-python3.9-plover-modified-4.0.0.dev10/lib/python3.9/site-packages/plover/oslayer/wayland/keyboardcontrol.py", line 21, in <module>
    from .input_method_unstable_v2 import ZwpInputMethodManagerV2
ModuleNotFoundError: No module named 'plover.oslayer.wayland.input_method_unstable_v2'
```

I'll see if I can find and fix the problem, but it might take a while.

benoit-pierre:
@antoniusf: my bad, should be fixed now.

antoniusf:
I have another question actually: I currently have problems with certain key combinations. In particular, trying to use page up and page down doesn't work and results in the following error (some lines cut from the traceback):

```
2022-04-03 00:14:21,809 [Dummy-2] ERROR: engine on_stroked failed
...
  File "/nix/store/43kdzpv6k3p9v48nqhi3984jn3n9l2cr-python3.9-plover-modified-4.0.0.dev10/lib/python3.9/site-packages/plover/oslayer/wayland/keyboardlayout.py", line 81, in parse_key_combo
    return parse_key_combo(combo_string, self._combo_from_keyname.__getitem__)
  File "/nix/store/43kdzpv6k3p9v48nqhi3984jn3n9l2cr-python3.9-plover-modified-4.0.0.dev10/lib/python3.9/site-packages/plover/key_combo.py", line 175, in parse_key_combo
    key_code = key_name_to_key_code(key_name)
KeyError: 'page_up'
```

I've spent way too long looking at all this and trying to figure something out, but I did notice that the key combinations are sent via the replay virtual keyboard, which uses the system keymap, and not the custom output keyboard, which uses the flexible keymap. Is that correct? Is there reason for this? My current theory is that the replay keyboard just doesn't have these keys for some reason...

benoit-pierre:
> I have another question actually: I currently have problems with certain key combinations. In particular, trying to use page up and page down doesn't work and results in the following error (some lines cut from the traceback):
> 
> ```
> 2022-04-03 00:14:21,809 [Dummy-2] ERROR: engine on_stroked failed
> Traceback (most recent call last):
> ...
>   File "/nix/store/43kdzpv6k3p9v48nqhi3984jn3n9l2cr-python3.9-plover-modified-4.0.0.dev10/lib/python3.9/site-packages/plover/oslayer/wayland/keyboardlayout.py", line 81, in parse_key_combo
>     return parse_key_combo(combo_string, self._combo_from_keyname.__getitem__)
>   File "/nix/store/43kdzpv6k3p9v48nqhi3984jn3n9l2cr-python3.9-plover-modified-4.0.0.dev10/lib/python3.9/site-packages/plover/key_combo.py", line 175, in parse_key_combo
>     key_code = key_name_to_key_code(key_name)
> KeyError: 'page_up'
> ```
> 
> I've spent way too long looking at all this and trying to figure something out, but I did notice that the key combinations are sent via the replay virtual keyboard, which uses the system keymap, and not the custom output keyboard, which uses the flexible keymap. Is that correct? Is there reason for this? My current theory is that the replay keyboard just doesn't have these keys for some reason...

We're using the replay keyboard because the goal of key combos is to simulate shortcuts, as you would type them on the keyboard.

Anyway, the reason `page_up` is not working is because `xkb.keysym_get_name` is returning `Prior`, which is another name for `Page_Up`:
```C
#define XK_Prior                         0xff55  /* Prior, previous */
#define XK_Page_Up                       0xff55
```

So will need to make sure we add aliases for those cases.

benoit-pierre:
List of aliases:
```
0xd0: ETH, Eth
0xffc8: F11, L1
0xffc9: F12, L2
0xffca: F13, L3
0xffcb: F14, L4
0xffcc: F15, L5
0xffcd: F16, L6
0xffce: F17, L7
0xffcf: F18, L8
0xffd0: F19, L9
0xffd1: F20, L10
0xffd2: F21, R1
0xffd3: F22, R2
0xffd4: F23, R3
0xffd5: F24, R4
0xffd6: F25, R5
0xffd7: F26, R6
0xffd8: F27, R7
0xffd9: F28, R8
0xffda: F29, R9
0xffdb: F30, R10
0xffdc: F31, R11
0xffdd: F32, R12
0xffde: F33, R13
0xffdf: F34, R14
0xffe0: F35, R15
0xff23: Henkan, Henkan_Mode
0xff9b: KP_Next, KP_Page_Down
0xff9a: KP_Page_Up, KP_Prior
0xff3e: Mae_Koho, PreviousCandidate
0xff7e: Mode_switch, script_switch
0xff3d: MultipleCandidate, Zen_Koho
0xff56: Next, Page_Down
0xff55: Page_Up, Prior
0xde: THORN, Thorn
0x27: apostrophe, quoteright
0x60: grave, quoteleft
```

benoit-pierre:
Should be fixed now.

antoniusf:
Thanks for the quick fix! Works for me now :tada:

mtoohey31:
@antoniusf I noticed the mention of `/nix/store` in your logs, would you be able to share the overlay you're using for this branch? I've been trying to get it working on Nixos too but I've run into many issues, the latest of which is related to `/usr/share/wayland/wayland.xml` being missing...

antoniusf:
@mtoohey31 sure! It's not really an overlay though, just a small mess of a few nix derivations and duct tape. I'll see if I can get this into more proper form, but until then, the lines you're missing are probably these?

```
  postPatch = ''
    sed -i /PyQt5/d setup.cfg
+   substituteInPlace plover_build_utils/setup.py \
+    --replace "/usr/share/wayland/wayland.xml" "${wayland}/share/wayland/wayland.xml"
  '';
```

I also had to add the python packages rtf_tokenize and plover_stroke, which aren't in nixpkgs yet, but are very straightforward to write derivations for. (But which is also why this is a bit messy.) So, I'll get back to you with something that's shareable, but if you manage to make something neat I'd love to see it!

mtoohey31:
Thanks, that was it! This is what I ended up with, it works but it's a bit of a mess too. I'll edit this and clean it up if I get around to it:

```nix
(self: super: rec {
  python3Packages = {
    plover-stroke = self.python3Packages.buildPythonPackage rec {
      pname = "plover_stroke";
      version = "1.0.1";
      src = super.python3Packages.fetchPypi {
        inherit pname version;
        sha256 = "t+ZM0oDEwitFDC1L4due5IxCWEPzJbF3fi27HDyto8Q=";
      };
    };
    rtf-tokenize = self.python3Packages.buildPythonPackage rec {
      pname = "rtf_tokenize";
      version = "1.0.0";
      src = super.python3Packages.fetchPypi {
        inherit pname version;
        sha256 = "XD3zkNAEeb12N8gjv81v37Id3RuWroFUY95+HtOS1gg=";
      };
    };
    pywayland_0_4_7 = super.python3Packages.pywayland.overridePythonAttrs
      (oldAttrs: rec {
        pname = "pywayland";
        version = "0.4.7";
        src = super.python3Packages.fetchPypi {
          inherit pname version;
          sha256 = "0IMNOPTmY22JCHccIVuZxDhVr41cDcKNkx8bp+5h2CU=";
        };
      });
  } // super.python3Packages;
  plover.dev = super.plover.dev.overridePythonAttrs
    (oldAttrs: {
      src = self.fetchFromGitHub {
        owner = "openstenoproject";
        repo = "plover";
        rev = "fd5668a3ad9bd091289dd2e5e8e2c1dec063d51f";
        sha256 = "2xvcNcJ07q4BIloGHgmxivqGq1BuXwZY2XWPLbFrdXg=";
      };
      propagatedBuildInputs = oldAttrs.propagatedBuildInputs
        ++ [
        python3Packages.plover-stroke
        python3Packages.rtf-tokenize
        python3Packages.pywayland_0_4_7
      ];
      nativeBuildInputs = (oldAttrs.nativeBuildInputs or [ ]) ++ [
        self.pkg-config
      ];
      doCheck = false; # TODO: get tests working
      postPatch = ''
        sed -i /PyQt5/d setup.cfg
        substituteInPlace plover_build_utils/setup.py \
          --replace "/usr/share/wayland/wayland.xml" "${self.wayland}/share/wayland/wayland.xml"
      '';
    });
})
```

antoniusf:
Oh nice, thank you!

sagehane:
I tried the Nix expression provided by @mtoohey31 and got the following runtime error:
```sh-session
$ plover
<omitted>
  File "/nix/store/sa1wvfw3hcwyc4bd7r23kyvrcshy4p0s-python3.10-xlib-0.31/lib/python3.10/site-packages/Xlib/support/unix_connect.py", line 76, in get_display
    raise error.DisplayNameError(display)
Xlib.error.DisplayNameError: Bad display name ""

Exception ignored in: <function Application.__del__ at 0x7f9ea26303a0>
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/gui_qt/main.py", line 71, in __del__
    del self._win
AttributeError: _win
```
I'm on RiverWM with XWayland disabled. I recall it working a bit better with XWayland, at least up to the point until it complained about [`input_method_unstable_v2` not being supported](https://github.com/riverwm/river/pull/405) when trying GeminiPR.

Does it still need XWayland for the GUI?

antoniusf:
Hmm, so I just tried running plover with DISPLAY unset, and I got a similar error message. Reading the traceback, it seems like the error came from `xwmctrl.py` in my case, so maybe we'd have to look into this as well if we want it to run on pure wayland? Either way, does this match your error? (Given that you're not running xwayland, I assume you have already set `QT_QPA_PLATFORM=wayland` to make sure qt apps know to use wayland.)

sagehane:
I totally forgot to set `QT_QPA_PLATFORM=wayland` because I never needed it before (which is weird because I do use at least one Qt application), but setting it didn't change the output for me. In fact, the only env var I've set so far on NixOS for Wayland is `MOZ_ENABLE_WAYLAND=1` for the `tor-browser-bundle-bin` package.

And yeah, `xwmctrl.py` is among some of the files I mentioned in the `<omitted>` portion from my previous message. I'll just dump the entire thing because it doesn't seem unbearably long.

```sh-session
$ QT_QPA_PLATFORM=wayland plover
Unexpected error: Traceback (most recent call last):
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/scripts/main.py", line 131, in main
    code = gui.main(config, controller)
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/gui_qt/main.py", line 98, in main
    app = Application(config, controller, use_qt_notifications)
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/gui_qt/main.py", line 31, in __init__
    from plover.gui_qt.main_window import MainWindow
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/gui_qt/main_window.py", line 16, in <module>
    from plover.oslayer import wmctrl
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/oslayer/wmctrl.py", line 34, in <module>
    wmctrl = WmCtrl()
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/oslayer/xwmctrl.py", line 9, in __init__
    self._display = display.Display()
  File "/nix/store/sa1wvfw3hcwyc4bd7r23kyvrcshy4p0s-python3.10-xlib-0.31/lib/python3.10/site-packages/Xlib/display.py", line 89, in __init__
    self.display = _BaseDisplay(display)
  File "/nix/store/sa1wvfw3hcwyc4bd7r23kyvrcshy4p0s-python3.10-xlib-0.31/lib/python3.10/site-packages/Xlib/display.py", line 71, in __init__
    protocol_display.Display.__init__(self, *args, **keys)
  File "/nix/store/sa1wvfw3hcwyc4bd7r23kyvrcshy4p0s-python3.10-xlib-0.31/lib/python3.10/site-packages/Xlib/protocol/display.py", line 84, in __init__
    name, protocol, host, displayno, screenno = connect.get_display(display)
  File "/nix/store/sa1wvfw3hcwyc4bd7r23kyvrcshy4p0s-python3.10-xlib-0.31/lib/python3.10/site-packages/Xlib/support/connect.py", line 73, in get_display
    return mod.get_display(display)
  File "/nix/store/sa1wvfw3hcwyc4bd7r23kyvrcshy4p0s-python3.10-xlib-0.31/lib/python3.10/site-packages/Xlib/support/unix_connect.py", line 76, in get_display
    raise error.DisplayNameError(display)
Xlib.error.DisplayNameError: Bad display name ""

Exception ignored in: <function Application.__del__ at 0x7fe2663583a0>
  File "/nix/store/k3f35vmkywwbmj9m8zpywkbj0js1dvv8-python3.10-plover-4.0.0.dev10/lib/python3.10/site-packages/plover/gui_qt/main.py", line 71, in __del__
    del self._win
AttributeError: _win
```

antoniusf:
Huh, maybe QT_QPA_PLATFORM isn't necessary if there isn't an X server to connect to. Either way, your error matches mine, so that module will probably have to be moved to wayland as well (or, at least, disabled so it doesn't cause errors). Most people probably have an xwayland running along though, and I don't think the module is causing any problems as is. I guess that does answer your question though? The gui doesn't need X, but some other bit of code still tries to connect.

user202729:
Regarding the closure of the pull request, I'll quote Ted's announcement in the Discord server here...

> I learned today that Benoit Pierre has decided to move on from the Open Steno Project.
> 
> I'd like to thank him for his years of service. No other dev has had as much impact on Plover as Benoit Pierre. He is responsible for the conversion from python 2 to 3, the introduction of the plugins manager, the QT GUI, adding multilingual support, and so, so much more.
> 
> Thank you Benoit Pierre for your many years supporting Plover and the community. It's been an honor to learn from you and it was always a pleasure to work with you.

matteodelabre:
Thanks for the context.

Mateiadrielrafael:
I have tried running off this branch using the overlay posted above, but I assume something (hyprland?) crashes in the process, because I see a black screen for half a second and then get logged out.

sagehane:
> I have tried running off this branch using the overlay posted above, but I assume something (hyprland?) crashes in the process, because I see a black screen for half a second and then get logged out.

I'm pretty sure this is a bug on Hyprland or whatever login manager you're using (idk what you mean by "logged out") as the compositor probably should never crash as a result of software interaction.

---

Since https://github.com/riverwm/river/issues/351 has been resolved, I finally decided to give Plover a try again.

I modified the Nix expression slightly as I forgot how to get it work. I'm happy to say that Plover seems to work on every app I tried but Element Desktop (Flatpak version is the one I tested) where it refuses to append spaces after words and `*` doesn't delete the previous word.

Expression:
```nix
# Build with `nix-build <name of file>`
with import <nixpkgs> {}; let
  plover-stroke = python3Packages.buildPythonPackage rec {
    pname = "plover_stroke";
    version = "1.1.0";

    src = python3Packages.fetchPypi {
      inherit pname version;
      sha256 = "sha256-3gOyP0ruZrZfaffU7MQjNoG0NUFQLYa/FP3inqpy0VM=";
    };
  };

  rtf-tokenize = python3Packages.buildPythonPackage rec {
    pname = "rtf_tokenize";
    version = "1.0.0";

    src = python3Packages.fetchPypi {
      inherit pname version;
      sha256 = "XD3zkNAEeb12N8gjv81v37Id3RuWroFUY95+HtOS1gg=";
    };
  };

  pywayland_0_4_7 =
    python3Packages.pywayland.overridePythonAttrs
    (oldAttrs: rec {
      pname = "pywayland";
      version = "0.4.7";

      src = python3Packages.fetchPypi {
        inherit pname version;
        sha256 = "0IMNOPTmY22JCHccIVuZxDhVr41cDcKNkx8bp+5h2CU=";
      };

      doCheck = false;
    });
in
  plover.dev.overridePythonAttrs
  (oldAttrs: {
    src = fetchFromGitHub {
      owner = "openstenoproject";
      repo = "plover";
      rev = "fd5668a3ad9bd091289dd2e5e8e2c1dec063d51f";
      sha256 = "2xvcNcJ07q4BIloGHgmxivqGq1BuXwZY2XWPLbFrdXg=";
    };

    propagatedBuildInputs =
      oldAttrs.propagatedBuildInputs
      ++ [
        plover-stroke
        rtf-tokenize
        pywayland_0_4_7
      ];

    nativeBuildInputs = (oldAttrs.nativeBuildInputs or []) ++ [pkg-config];
    doCheck = false; # TODO: get tests working

    postPatch = ''
      sed -i /PyQt5/d setup.cfg
      substituteInPlace plover_build_utils/setup.py \
        --replace "/usr/share/wayland/wayland.xml" "${wayland}/share/wayland/wayland.xml"
    '';
  })
```

sagehane:
I just noticed that with the expression I posted above and the latest commit of River, Gemini PR works fine but setting the machine mode to "Keyboard" causes pretty much all keys but Space and Tab to output nothing. Plover clearly doing something in that it's successfully eating my inputs, but it seems far from functional. I would appreciate it if anyone else could try reproducing it.

greenwyrt:
superseded by #1679

PR #1463, opened 2022-01-05
dictionary: fix KHR*PB stroke

### LINKED ISSUE #1407: ERROR: engine on_stroked failed with KeyError: '^' when typing KHR*PB
**Describe the bug**

Trying to type a colon with no space after (I think that's what `{:^}` is) fails with an error.

**To Reproduce**

1. Enable Plover output
2. Chord KHR*PB
3. See error

**Expected behavior**

A colon should be produced.

**Screenshots**

ERROR: engine on_stroked failed with KeyError: '^'

**Operating system**

 - OS: Windows 10
 - Plover Version 4.0.0.dev10+2.g0501aa3

**Hardware**

If relevant, an EcoSteno v1.0.18 configured as follows:
Machine configuration

lf-:
Currently, this entry gets misparsed as a command and thus throws an
exception. Change it to do a raw attach as suggested in
https://github.com/openstenoproject/plover/issues/1407#issuecomment-914881597

Fix #1407.

Thanks @sammdot 

## Summary of changes

`{:^}` is parsed as the command `^` incorrectly.

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1463.dict.md
+++ b/news.d/bugfix/1463.dict.md
@@ -1,0 +1,1 @@
+Fix `KHR*PB` stroke to not be misinterpreted as a command.


user202729:
There's #400 to suggest edits to the dictionary, but generally speaking the dictionary hasn't been touched for a while.

Also – how was the entry generated in the first place? If it's something that may be repeated – RTF conversion for example – it may be better to find and fix the source

lf-:
Hm! 718f0f3a409d475024fd355fe648e3508050a91d is the one that put it to what it is currently, rather than the raw join, but I have absolutely no idea why it was done since the commit message says nothing of interest. There was probably a comment on the #400 thread that explained it, but it was (very frustratingly) deleted in 2016.

benoit-pierre (approved):

PR #1023, opened 2018-11-11
Update plugins, development and distributions requirements, and the distributions' Python version

benoit-pierre:
Initially, the idea was to stay on PyQt5 5.9.x (since Qt 5.9 is supposed to be a LTS version), and only update to new point releases, but PyQt5 does not follow that, so we might as well update to the latest version.

Note: there's no QtWebEngine support with recent win32 builds of PyQt5, hence the switch to generating a 64bits distribution.

I've tested locally the new AppImage on Arch Linux and Ubuntu Trusty, and the new Windows installer/distribution on a Windows 10 VM, someone need to test the macOS image.

PR #1469, opened 2022-01-17
Improve Windows dev environment documentation

jrowberg:
## Summary of changes

Add explicit documentation for how to get up and running quickly for Plover development in Windows, based on the headache I had to endure today trying fruitlessly with Miniconda and then working through MS Build Tools hurdles.

No code changes are included, only a single block of additional text in the previously near-empty `windows/README.md` file.

### DIFF
--- a/windows/README.md
+++ b/windows/README.md
@@ -5,4 +5,60 @@
 Some of the development helpers rely on `bash` being available. You can use the
 version provided by the official `git` installer.
 
+For the quickest way to get started developing in Windows, use WinPython, *not*
+Anaconda/Miniconda. Some users report impossible dependency/build headaches when
+initializing a development environment starting from a base Miniconda3
+installation that is otherwise working perfectly.
+
+You will also need Microsoft Build Tools and an SDK appropriate for your
+operating system. These are required in order to build the `plover-stroke`
+package, which contains C code and does not currently have cross-platform Python
+wheel packages available. These two tools can be downloaded from their official
+sources here:
+
+- https://winpython.github.io/
+- https://visualstudio.microsoft.com/visual-cpp-build-tools/
+
+**NOTE 1:** You can choose whichever WinPython release you want (32-bit/64-bit),
+including the smallest "dot" releases that contain only the bare minimum
+required for a functional Python interpreter. If you choose this or any other
+pre-built installer that does not include the `tox` package, you will need to
+install it with `pip` later.
+
+**NOTE 2:** There seem to be many sources for various MS Build Tools installers,
+even from Microsoft. The above link is the correct one as of early 2022. When
+you run the installer, make sure you select **both** the SDK for your Windows
+version and the build tools for it. As of early 2022, you should choose
+**"Windows 10 SDK (10.0.19041.0)"** and **"MSVC v142 - VS 2019 C++ x64/x86 build
+tools (Latest)"**, assuming you are running Windows 10. This unfortunately
+requires many gigabytes of data for a task that only needs perhaps 20 MB of
+binary utilities and header files, but such is life.
+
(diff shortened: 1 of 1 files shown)


benoit-pierre:
What version of Windows / Python did you use? Because `plover-stroke` does have wheels for Windows 64 bits and Python versions 3.7 to 3.10.

jrowberg:
I'm running Win10 64-bit, and my Miniconda installation is 64-bit. However, maybe I accidentally made life way harder by absent-mindedly installing the 32-bit version of WinPython 3.9 when I tried going that route.

...yes, yes I did. I just tried with the WinPython 3.9.8.0 64-bit, and it worked without my needing to do any of that mess with MS Build Tools. Good lord.

Maybe I should rework some of the new content so that it notes the differences and provides better help to anyone who actually needs/wants 32-bit Python.

benoit-pierre:
> Maybe I should rework some of the new content so that it notes the differences and provides better help to anyone who actually needs/wants 32-bit Python

Or better yet, we only support 64 bits Python, done. In fact, I'm surprised to see that the latest version of `PyQtWebEngine` has a 32 bits wheel for Windows available (32 bits support was [dropped at some point](https://github.com/openstenoproject/plover/pull/1023)). Anyway, it's baffling to me that 32 bits is still a thing on Windows.

I'm also honestly not a fan of adding specific instructions for Windows, beyond the [standard developer documentation](https://github.com/openstenoproject/plover/blob/master/doc/developer_guide.md#environment-setup).

All those environment specific instructions always end up outdated, plus once you add instructions for one env, everyone want to do it for their env, and things tends to balloon from there...

And not necessarily the simplest way if you just want to develop a Python only plugin (no C extension).

jrowberg:
Thanks for the additional feedback. It's definitely not my call to decide about official architecture support (though I agree about 64-bit target priority these days, especially in Windows). I didn't realize that 32-bit support was dropped before, or I would have been looking out specifically for the right WinPython install. 🤦‍♂️

I've cut down and simplified the additions I made to the Windows doc, which you can review with formatting easily here:

https://github.com/jrowberg/plover/blob/improve-windows-docs/windows/README.md

Having gone through the the dev setup process today, even ignoring the architecture mistake, I think this type of "getting started" content is still valuable. It's the level of detail I was hoping to find when I began, a way to quickly connect a few important dots that a new dev on the project might not know ahead of time.

Even so, if you want to close without merging, no hard feelings.

mkrnr:
...it's been a while 😁 

@jrowberg are you still interested in working on this?

Some thoughts:

* Maybe, instead of explicitly recommending WinPython, I'd just add a warning that Anaconda or Miniconda might cause problems. I personally use the official Python installer to be as close as possible to our CI/CD setup. There are so many ways to set up Python 😅 
* Related to the point by Benoit, we probably should remove any concrete versions (Python >=3.6, e.g. Windows 10 SDK, MSVC v142) since we'd have to maintain them.
* Mentioning `Microsoft Build Tools` is helpful. And that's not only relevant during `plover-stroke` development but when building Plover in general. I stumbled over this several times, even though one can find the solution relatively fast based on the error message.

jrowberg:
Hi @mkrnr,

It has indeed been a while! I'm not in a position to return to this project at the moment; I was initially doing it for my mom (who is, or now at least used to be, a stenographer), but our lives have shifted around since early 2022. I'm trying to step back from as many things as possible, and right now this one is definitely very low on the list. Sorry!

mkrnr:
Hi @jrowberg, that's totally understandable, thanks a lot for your response. I wish you and your family all the best!

I'll close this PR but will consider adding Windows-specific details to the developer guide.

Thanks again for opening this PR back in the day!

PR #1473, opened 2022-01-23
Rename "add dictionaries" options

antistic:
## Summary of changes

image

Rename "open dictionaries" to "load dictionaries", and "new dictionary" to "create dictionary"

I could not find a specific GitHub issue for this, but some people have been struggling to understand which button to press to use their existing dictionaries when coming from non-Plover software. It was suggested that changing the wording might help with this (discord links [here](https://discord.com/channels/136953735426473984/136953809409802240/925834811094990968) and [here](https://discord.com/channels/136953735426473984/850891109140004864/934828102859767879). scroll up for context). Feedback on exact wording is appreciated.

---

I haven't worked with i18n code before, so please let me know if there is a better way to do it than I did.

- There were several `msgid`s:  "&add dictionaries" for the parent menu item, "open dictionaries" for the submenu item, "add dictionaries" for the file picker
- I have merged the latter two (to match the "new dictionary" submenu item and "new dictionary" file picker)
- I named the merged msgid "Load dictionaries"
- ~~In the translations I kept the text that was previously "Open dictionaries", since I don't know those languages and assumed this one would match better. Presumably we should contact the translators to update these translations when they are ready.~~ (Removed modifications to translation files after [SeaLiteral's comment](https://github.com/openstenoproject/plover/pull/1473#issuecomment-1019540399))
- Renamed "new dictionary" to "create dictionary"
- Tested on Linux in English

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/messages/plover.pot
+++ b/plover/messages/plover.pot
@@ -1,21 +1,21 @@
 # Translations template for plover.
-# Copyright (C) 2021 ORGANIZATION
+# Copyright (C) 2022 ORGANIZATION
 # This file is distributed under the same license as the plover project.
-# FIRST AUTHOR <EMAIL@ADDRESS>, 2021.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2022.
 #
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: plover 4.0.0.dev9\n"
+"Project-Id-Version: plover 4.0.0.dev10\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2021-05-23 20:19+0200\n"
+"POT-Creation-Date: 2022-02-06 17:18+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Generated-By: Babel 2.9.0\n"
+"Generated-By: Babel 2.9.1\n"
 
 #: plover/__init__.py:19
 msgid ""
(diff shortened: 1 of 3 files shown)


SeaLiteral:
It might make sense to make an issue or discussion to figure out which wording most people prefer. Some thoughts:

1. I'd rather rename "new dictionary" to "create new dictionary" or "new empty dictionary", to show that it's creating a new file rather than opening an existing one.
2. In some pieces of software, "import" can mean to convert a file into the software's native file format and continue working with the converted file. I think "load" might work better, or maybe even keep the word "open".

I wouldn't say "import" is wrong. Just that there's other ways I think would make just as much sense to solve the same problem.

SeaLiteral:
Also, I mentioned this on Discord, but I'll mention it here too for those who can't see it here: if you aren't changing text in other languages than English, you probably don't need to touch the translation files, especially non-English po files.

benoit-pierre:
It seems to me that `Open` and `New` is pretty standard wording, but sure, why not... Can you cleanup the history?

The way to update the main translations catalog is to use `setup.py extract_messages` (and preferably a dedicated commit, e.g.: `i18n: update main translations catalog`, as there may be other changes that get picked up). This is what is automatically done by the CI's packaging job on new release (to create the `Translations Catalogs` asset).

plover/messages/plover.pot
```diff
@@ -1141,17 +1144,21 @@ msgstr ""
 msgid "no suggestions"
 msgstr ""
 
-#: plover/gui_qt/trayicon.py:115
+#: plover/gui_qt/suggestions_widget_ui.py:31
+msgid "Form"
+msgstr ""
```
benoit-pierre, line 1149: Where is that coming from? Can you run ensure the generated UI files are up-to-date (`setup.py build_ui -f`) before running `./setup.py extract_messages`.

antistic: It's gone now. I had to manually remove `_ui.py` files first (`-f` does not seem to do this?)

benoit-pierre (approved):

plover/messages/plover.pot
benoit-pierre: That's because it's a left-over, the UI file was removed in cc794c242dacd1e0cd13275aafe38bff7e910a8f.

PR #1480, opened 2022-02-04
Don't carry capitalisation when translating {\cxfing }

dnaq:
## Summary of changes
In plovers main.json the fingerspelling entries for lowercase characters
are defined as `{>}{&a}` whereas the uppercase characters are defined
as `{&A}`. This patch makes it so that fingerspelling entries from
RTF dictionaries are translated in the same way.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/news.d/bugfix/1480.core.md
+++ b/news.d/bugfix/1480.core.md
@@ -1,0 +1,2 @@
+Small changes to RTF/CRE:
+ - Don't carry capitalisation when translating lower case `{\cxfing}` entries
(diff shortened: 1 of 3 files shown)


benoit-pierre:
I don't think this is right, there's nothing in the RTF/CRE spec that says that's that how it should work, and if you wanted that behavior you could just add `\cxfl` (the equivalent of Plover's `{>}`) to those entries.

dnaq:
You’re probably correct. I’ve just based this on that the fingerspelling entries in the Phoenix starter dictionary are defined as cxfing and that the documentation explicitly calls them the lowercase alphabet, the uppercase alphabet and so on. So according to the Phoenix documentation they shouldn’t carry capitalisation.

On the other hand I still find the line between the RTF/CRE standard and whatever casecat is doing to be pretty blurry, so it’s probable that the casecat export is incorrect.

PR #1481, opened 2022-02-06
Speedup suggestions widget

### LINKED ISSUE #1477: Suggestions window causes Plover to lag with large scrollback
**Describe the bug**

Keeping the Plover: Suggestions window open causes Plover to eventually noticeably lag when there's a lot of scroll-back. 

This may be related to issue [#778](https://github.com/openstenoproject/plover/issues/778#issue-235042398) although it seems like a separate issue.

This slows down the performance of Plover with other windows, such as the Paper Tape window having delayed updates, along with words appearing like they are being typed slowly.

**To Reproduce**

1.  If the Plover: Suggestions window is not visible, navigate to the Plover tray/menu bar icon and select Tools > Suggestions.
2. Keep the Plover: Suggestions open for your entire of session of Plover running.
3. Wait until the Plover: Suggestions window's vertical scrollbar is approximately 12 logical pixels tall, depending on its window size—keep using Plover with steno inputs until it has a fairly large scroll-back. Usually lag is noticeable as the Suggestions window content builds up after a while.

image

**Expected behavior**

Plover's performance should remain like it is when the Suggestions window is not very populated without any degraded performance. 

Manually clearing the Suggestions window with the 'bin' icon restores Plover's performance, with windows such as the Paper Tape and text appearing instantly as it should.

**Screenshots**

Not applicable.

**Operating system**

 - OS: Windows 11 21H2, Build 22000. 376
 - Plover Version: 4.0.0.dev10+120.g4394ef1 

**Hardware**

Not applicable.

benoit-pierre:
## Summary of changes

Optimize the delegate' `sizeHint` implementation and cache calls to it.

Closes #1477.

### Pull Request Checklist
- ~[ ] Changes have tests~

### DIFF
--- a/news.d/bugfix/1481.ui.md
+++ b/news.d/bugfix/1481.ui.md
@@ -1,0 +1,1 @@
+Speedup suggestions widget implementation: should noticeably improve performance when there's a large scrollback.
--- a/plover/gui_qt/suggestions_widget.py
+++ b/plover/gui_qt/suggestions_widget.py
@@ -32,6 +32,10 @@ class SuggestionsDelegate(QStyledItemDelegate):
         self._translation_char_format = QTextCharFormat()
         self._strokes_char_format = QTextCharFormat()
         self._strokes_char_format.font().setStyleHint(QFont.Monospace)
+        self._size_hint_cache = {}
+
+    def clear_size_hint_cache(self):
+        self._size_hint_cache.clear()
 
     @property
     def text_font(self):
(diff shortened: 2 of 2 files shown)


morinted (approved):

dnaq:
I've been trying out this pull request for a while, and the suggestions widget feels a lot snappier than it used to. I can't say for sure that there isn't a slowdown when the number of widget entries grow, but it's a lot better than it used to be.

PR #1487, opened 2022-02-20
Fix setuptools warning

benoit-pierre:
## Summary of changes

Fix deprecation warning:
```
setup.py:5: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  from distutils import log
```


benoit-pierre:
Updated to keep the only relevant commit...

PR #1488, opened 2022-02-20
drop the dependency on `pkg_resources`

benoit-pierre:
## Summary of changes

Switch to `importlib_metadata`, or simplify the code to remove the need for `pkg_resources`.

Note: we use [`importlib_metadata`](https://pypi.org/project/importlib-metadata/) and not [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html?highlight=importlib%20metadata#module-importlib.metadata) until the version included by our minimum support Python version is recent enough.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/plover/i18n.py
+++ b/plover/i18n.py
@@ -2,9 +2,8 @@ import os
 import locale
 import gettext
 
-import pkg_resources
-
 from plover.oslayer.config import CONFIG_DIR, PLATFORM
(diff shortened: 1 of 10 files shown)


benoit-pierre:
*ugh*… The entrypoint objects returned by `importlib_metadata` does not actually check requirements on load (see https://github.com/python/importlib_metadata/issues/368).

PR #1489, opened 2022-02-20
Partially drop the dependency on `pkg_resources`

benoit-pierre:
## Summary of changes

Salvage parts of #1488: do not switch to `importlib_metadata`, keep using `pkg_resources` in `scripts.main` and for the `registry` implementation, remove other uses.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/plover/i18n.py
+++ b/plover/i18n.py
@@ -2,9 +2,8 @@ import os
 import locale
 import gettext
 
-import pkg_resources
-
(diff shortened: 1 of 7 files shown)


benoit-pierre:
Another benefit if this change, is how much faster the keyboard machine initialization is on Windows: from 1.4s (“initializing” → “connected”), down to 0.4s.

PR #1491, opened 2022-02-26
system/english_stenotype: improve RTF/CRE compatibility

### LINKED ISSUE #1486: ValueError for strokes of the form `2#`
**Describe the bug**
 
My rtf-dictionary for Phoenix contains a bunch of strokes that end with the number sign. These are parsed by the rtf parses, but plover-stroke considers them invalid strokes since it wants them to start with the number bar sign.

I wasn't sure if I should file the bug here or in plover-stroke, since it could either be seen as a normalisation isse when parsing the rtf-file, or as a bug in plover-stroke.

**To Reproduce**

1. The following rtf-dictionary reproduces the bug:
```
{\rtf1\ansi{\*\cxrev100}{\*\cxdict}
{\*\cxsystem caseCATalyst4 version 19.06 by Stenograph L.L.C.}
{\info{\creatim\yr2011\mo8\dy8}
{\revtim\yr2019\mo11\dy25}
{\buptim\yr0\mo0\dy0}
}{\stylesheet{\s0 Normal;}{\s1 Question;}{\s2 Answer;}{\s3 Continuation;}{\s4 Colloquy;}{\s5 Paren;}{\s6 By Line;}{\s7 UserDef}}
{\*\cxsgstenotable \cxsgstnormal STKPWHRAO*EUFRPBLGTSDZ#(!
\cxsgstshifted 12K3W4R50*EU6R7B8G9SDZ#(!
\cxsgstenoleft7\cxsgstenomiddle5\cxsgstenoright10}
{\*\cxs 2#}2
}
```

**Expected behavior**

The chapter on the number bar in the RTF CR Extensions document says the following:

> IV.C.1.iii            Number Bar
If the number bar was used in a stroke, the stroke contains a number sign to indicate that.  The number sign may be anywhere within the stroke, but should probably be at one end or the other (e.g., /18# or /#18 but not /1#8).  Note that the number bar may be used even if no numeral-equivalent keys are pressed.  For instance, the stroke /K# is perfectly valid even though the 'K' steno character doesn't correspond to a numeral, and it is not the same as the stroke /K without the number bar.

benoit-pierre:
## Summary of changes

Support the arbitrary placement of the number sign when parsing steno. 

Closes #1486.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/news.d/feature/1491.core.md
+++ b/news.d/feature/1491.core.md
@@ -1,0 +1,1 @@
+Improve “English stenotype” system compatibility with RTF/CRE spec: support arbitrary placement of the number sign when parsing steno (e.g. `18#`, `#18`, and `1#8` are all valid and equivalent).
--- a/plover/steno.py
+++ b/plover/steno.py
@@ -30,11 +30,12 @@ class Stroke(BaseStroke):
     UNDO_STROKE = None
 
     @classmethod
-    def setup(cls, keys, implicit_hyphen_keys, number_key, numbers, undo_stroke):
+    def setup(cls, keys, implicit_hyphen_keys, number_key,
+              numbers, feral_number_key, undo_stroke):
         if number_key is None:
             assert not numbers
             numbers = None
-        super().setup(keys, implicit_hyphen_keys, number_key, numbers)
+        super().setup(keys, implicit_hyphen_keys, number_key, numbers, feral_number_key)
         cls._class = type(cls.__name__, (cls,), {'_helper': cls._helper})
         cls._class._class = cls._class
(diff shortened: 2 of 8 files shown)


benoit-pierre:
Note: this behavior (arbitrary placement of the number sign) is conditional on a new optional system plugin flag (`FERAL_NUMBER_KEY = True`).

Rational:
- it's specific to the RTF/CRE spec, and so the English system
- consequently it does not make sense to have it as default for all systems
- it's also only possible if the number key is not an implicit hyphen key, and its corresponding letter is unique

PR #1501, opened 2022-03-12
Better steno errors handling

benoit-pierre:
## Summary of changes

- change `normalize_stroke`, `normalize_steno`, and `steno_to_sort_key` default to strict mode, and stop logging errors.

- add an helper for dictionary implementations to log invalid steno errors on load: there is only one notification in case of errors (instead of potentially spamming the notification area), and the message includes the dictionary name and number of errors:

2022-03-12-014539_303x200_scrot

- add an error icon to invalid steno entries in the editor to make them stand out:

2022-03-12-014234_464x379_scrot

### Pull Request Checklist
- [x] Changes have tests


benoit-pierre:
I changed the error into a warning, since it's not critical.

PR #1509, opened 2022-03-20
`osslayer` reorg

benoit-pierre:
## Summary of changes

In preparation for output plugins / locking keys, among other things:

* tweak keyboard suppression handling to avoid calling methods of the capture instance before it has been started
* add base output and keyboard capture classes
* move platform specific `oslayer` modules to dedicated platform packages

Tested on Linux and Windows, the macOS code is completely untested.

### Pull Request Checklist
- [x] Changes have tests
~- [ ] News fragment added in news.d. See [documentation](https://github.com/openstenoproject/plover/blob/master/doc/developer_guide.md#making-a-pull-request) for details~


user202729 (commented):
With these changes it appears to work.

(there's a remaining issue I don't seem to see the errors, but I don't with some older version either)

benoit-pierre:
> With these changes it appears to work.

Some very stupid mistakes on my part… Thanks for testing.

> (there's a remaining issue I don't seem to see the errors, but I don't with some older version either)

Can you elaborate?

user202729:
I just don't see any error message associated with e.g. invalid steno, or dictionary not found. (although there's a possibility I just don't know how to configure Mac properly. Normally I guess there should be notification in the notification center (top right corner, click on the clock)?)

benoit-pierre:
> I just don't see any error message associated with e.g. invalid steno, or dictionary not found. (although there's a possibility I just don't know how to configure Mac properly. Normally I guess there should be notification in the notification center (top right corner, click on the clock)?)

Can you start from a terminal with `-l info`, maybe there's an issue with setting up the platform's log handler.

benoit-pierre:
> …maybe there's an issue with setting up the platform's log handler.

Note that if that's the case,  we should still fallback on using Qt for notifications.

user202729:
(regarding notification) Alright, seems to be my fault.

I installed latest continuous disk image and go to "System Preferences → Notifications → Plover → Allow Notifications"

(Does Mac users normally do this? If this is unusual maybe we could add to installation guide)

benoit-pierre:
> (for some reason it's by default off for me. // Does Mac users normally do this? If this is unusual maybe we could add to installation guide)

No idea.

PR #1512, opened 2022-03-26
steno_dictionary: fix `StenoDictionaryCollection.longest_key`

benoit-pierre:
## Summary of changes

Ignore disabled dictionaries!

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/news.d/bugfix/1512.core.md
+++ b/news.d/bugfix/1512.core.md
@@ -1,0 +1,1 @@
+Fix `StenoDictionaryCollection.longest_key` implementation: ignore disabled dictionaries!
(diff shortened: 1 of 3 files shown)


user202729:
Looks like previously it was necessary (only) because it's harder to notify listener otherwise. (Listeners are gone now.)

(Unrelated, but reading through the code it looks like that there are some redundant code

https://github.com/openstenoproject/plover/blob/2ada7c71cd25a114e1439817a44206ff0da8b70e/plover/steno_dictionary.py#L84-L87

https://github.com/openstenoproject/plover/blob/2ada7c71cd25a114e1439817a44206ff0da8b70e/plover/steno_dictionary.py#L166-L168

)

benoit-pierre:
@user202729 : good catch, fixed.

PR #1528, opened 2022-05-24
Added feature to read and apply Qt style sheet from file

### LINKED ISSUE #1514: Load and use Qt application style sheets from Plover.qss file

A use case would be to customize Plover plugins act/look more like OSDs. Specifically,  to have the word tray window displayed with a fully customized style, e.g. transparent background (can also be achieved using window rules), a different font size and foreground color than the defaults, minimizing the padding around the word list, ....

A straightforward solution is to implement looking for a Qt style sheet file in default locations, defining some sensible fallback values (e,g, the ones Plover uses currently) and setting the style sheet with the `QApplication:setStyleSheet()` function. In principle this would be a few lines only, but I expect the error/exception/fallback handling to make up most of the lines for this change. This solution would also provide the means for users to experiment with the style sheets on this [Gallery Of Qt QSS Based Styles](https://qss-stock.devsecstudio.com/templates.php)

**Describe alternatives you've considered**

I've considered KWin window rules to set e.g. the window background transparency, but that only goes part of the way.

MarkoSchuetz:
## Summary of changes

If a file 'plover.qss' is present in the config directory then it gets read at the initialization of the Application.

Closes #1514 
### Pull Request Checklist
- [ ] Changes have tests
- [x ] News fragment added in news.d. See [documentation](https://github.com/openstenoproject/plover/blob/master/doc/developer_guide.md#making-a-pull-request) for details

### DIFF
--- a/news.d/bugfix/1274.windows.md
+++ b/news.d/bugfix/1274.windows.md
@@ -1,0 +1,1 @@
+Fix some key combinations being sent incorrectly.
--- a/plover/oslayer/windows/keyboardcontrol.py
+++ b/plover/oslayer/windows/keyboardcontrol.py
@@ -50,14 +50,6 @@ SCANCODE_TO_KEY = {
 
 KEY_TO_SCANCODE = dict(zip(SCANCODE_TO_KEY.values(), SCANCODE_TO_KEY.keys()))
 
-# Keys that need an extended key flag for Windows input
-EXTENDED_KEYS = {
-    # Control   Alt         INS   DEL   HOME  END   PG UP PG DN Arrows
-    0xA2, 0xA3, 0xA4, 0xA5, 0x2D, 0x2E, 0x21, 0x22, 0x24, 0x23, 0x25, 0x26,
-    # Arrow     NmLk  Break PtSc  Divide
-    0x27, 0x28, 0x90, 0x03, 0x2C, 0x6F
-}
-
 PASSTHROUGH_KEYS = {
     0XA2, 0XA3, # Control
     0XA0, 0XA1, # Shift
@@ -477,8 +469,6 @@ class KeyboardEmulation(Output):
 
     def _key_event(self, keycode, pressed):
         flags = 0 if pressed else KEYEVENTF_KEYUP
-        if keycode in EXTENDED_KEYS:
-            flags |= KEYEVENTF_EXTENDEDKEY
         self._send_input(self._keyboard(keycode, flags))
 
     # Press and release a key


plover/gui_qt/main.py
```diff
@@ -43,6 +45,11 @@ def __init__(self, config, controller, use_qt_notifications):
         self._app = QApplication([sys.argv[0], '-name', 'plover'])
         self._app.setAttribute(Qt.AA_UseHighDpiPixmaps)
 
+        styleSheetPath = os.path.join(CONFIG_DIR, 'plover.qss')
+        if os.path.isfile(styleSheetPath):
+            with open(styleSheetPath, 'r') as styleSheetFile:
+                log.info('stylesheet: %s', styleSheetPath)
+                self._app.setStyleSheet(styleSheetFile.read())
```
benoit-pierre, line 52:
```suggestion
        stylesheet = os.path.join(CONFIG_DIR, 'plover.qss')
        if os.path.isfile(stylesheet):
            log.info('applying stylesheet: %s', stylesheet)
            with open(stylesheet, 'r') as fp:
                self._app.setStyleSheet(fp.read())
```

plover/gui_qt/main.py
```diff
@@ -5,6 +5,7 @@
     QCoreApplication,
     QLibraryInfo,
     QTimer,
+    QFile,
```
benoit-pierre, line 8: This is not used:
```suggestion
```

plover/gui_qt/main.py
```diff
@@ -13,7 +14,8 @@
 
 from plover import _, __name__ as __software_name__, __version__, log
 from plover.oslayer.keyboardcontrol import KeyboardEmulation
-
+from plover.oslayer.config import CONFIG_DIR
+import os
```
benoit-pierre, line 18: Keep the blank line between standard and GUI specific imports (and keep them sorted). Move the `os` import where it belongs (together with the other Python standard modules at the top).

news.d/feature/1514.ui.md
```diff
@@ -0,0 +1 @@
+Read and apply Qt style sheet from file in config directory.
```
benoit-pierre, line 1: This can probably be reworded to be clearer, and specify the name of the name of the file Plover will look for.

plover/gui_qt/main.py
user202729: Actually, since this is Python it's probably more idiomatic to use EAFP. Open it and catch FileNotFoundError.

MarkoSchuetz: I like to keep the _Path_ part in the name: it is not the style sheet itself after all.

user202729:
I have no comment, but I don't have write permission anyway.

user202729:
I have no comment, but I don't have write permission anyway.

benoit-pierre:
OK, I messed up trying to update the PR... Next time please do your work in a dedicated branch (not master). New PR here: #1535.

I've added #1534 to fix the error handling.

PR #1558, opened 2022-08-29
plover/machine/*: DRY up / remove code duplication

sol:
## Summary of changes

This removes some code duplication. I'm not yet sure if it's worth the effort (e.g. would we want to update the plugin documentation accordingly?).

As this is a refactoring, it technically should not require any new tests.  However, from what I understand only `test/test_passport.py` tests the relevant code right now.

Regarding news fragments + docstring, is this relevant for the plugins api?  If yes, I guess we would want both?

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/machine/base.py
+++ b/plover/machine/base.py
@@ -90,6 +90,11 @@ class StenotypeBase:
         for callback in self.stroke_subscribers:
             callback(steno_keys)
 
+    def _notify_keys(self, steno_keys):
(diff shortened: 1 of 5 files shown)


sol:
@benoit-pierre can I get your input on this?

user202729 (approved):

sammdot (approved):

PR #1560, opened 2022-09-01
Update machine state on unhandled exceptions

### LINKED ISSUE #1559: Unplugging a serial-over-USB steno machine leaves Plover in an undefined state
**Description**

When you unplug a serial-over-USB steno machine then:

- The thread that handles the serial connection dies silently.
- Other parts of the system never learn about this.
- The tray icon is never updated, the disconnect is never indicated to the user.

**Steps to reproduce**

1. Plug a GeminiPR serial-over-USB machine into a free USB port of your computer (e.g. a USB keyboard running QMK).
3. In Plover, configure and connect the machine.
4. Unplug the machine from your computer.

**Expected behavior**

Tray icon and mouseover text should indicate that the machine is disconnected:

expected

**Actual behavior**

Tray icon and mouse overtext **wrongly indicate** that the machine is connected:

actual

**Operating system**

 - OS: Linux 5.19.4 (zen patches)
 - Plover Version: https://github.com/openstenoproject/plover/commit/3066a9a47269861ac1d66f8f05cdb26f7251b98d (master)
 
 **Discussion**

Plover handles serial connections in a dedicated *machine thread*. The machine thread communicates events to other parts of the system via callbacks.

When you unplug a serial-over-USB device then `pyserial` indicates the disconnect by throwing an exception:
```
serial.serialutil.SerialException: device reports readiness to read but returned no data
(device disconnected or multiple access on port?)
```

Plover does not handle these exceptions. Instead the machine thread dies silently.

This has been discussed before in #596 and more recently in #1273. There is also a related PR that address the issue #1054.

From what I can tell, this is straightforward to fix. I’m not sure what exactly is holding a fix back. Two things I noticed:
1. Reading through all the related discussions gets unwieldy quickly. To address this I hope that this issue provides a more compact summary.
2. The exciting PR is technically a breaking change. To address this I opened #1560, an alternative approach that does not introduce any breaking changes.

sol:
## Summary of changes

When the *machine thread* dies due to an unhadled exception then indicate this to the engine (and by extension to the user) by setting the machine state to "disconnected".

This is a first step.  Possible follow-ups:
1. We may choose to handle `serial.serialutil.SerialException` more explicitly as a "disconnect" (say, not even print a stack trace to `stderr`).
2. What I really want from a users perspective is that the machine re-connects automatically (I think it's feasible to do so, stay tuned for a follow-up PR).
3. For `SerialStenotypeBase` specifically, we also want to close the serial port on unhandled exceptions.

Even if we do (1) and/or (2) later, I think we still want this PR as a fallback for arbitrary exceptions.

I will have a separate PR for (3), but again, I think this PR provides merit on its own. So let’s focus on this first.

Closes #1559.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/news.d/bugfix/1560.ui.md
+++ b/news.d/bugfix/1560.ui.md
@@ -1,0 +1,1 @@
+Update the tray icon to "disconnected" when a serial-over-USB machine is unplugged.
--- a/plover/machine/base.py
+++ b/plover/machine/base.py
@@ -149,9 +149,17 @@ class ThreadedStenotypeBase(StenotypeBase, threading.Thread):
     """
     def __init__(self):
         threading.Thread.__init__(self)
+        self._on_unhandled_exception(self._error)
         self.name += '-machine'
         StenotypeBase.__init__(self)
         self.finished = threading.Event()
+
+    def _on_unhandled_exception(self, action):
+        super_invoke_excepthook = self._invoke_excepthook
+        def invoke_excepthook(self):
+            action()
+            super_invoke_excepthook(self)
+        self._invoke_excepthook = invoke_excepthook
 
     def run(self):
         """This method should be overridden by a subclass."""
--- a/test/test_machine.py
+++ b/test/test_machine.py
@@ -1,0 +1,14 @@
+from unittest.mock import Mock
+from plover.machine.base import ThreadedStenotypeBase
+
+class MyMachine(ThreadedStenotypeBase):
+    def run(self):
+        raise "some unexpected error"
+
+def test_update_machine_staten_on_unhandled_exception():
+    machine = MyMachine()
+    callback = Mock()
+    machine.add_state_callback(callback)
+    machine.start_capture()
+    machine.join()
+    callback.assert_called_with('disconnected')


user202729:
Somewhat related to #1054, although I can't immediately tell if they're identical.

sol:
@user202729  thanks for taking the time to look at this.

> Somewhat related to #1054, although I can't immediately tell if they're identical.

They are addressing the same underlying issue.  However, #1054 is technically a breaking change as it changes the base class of `ThreadedStenotypeBase` (see https://github.com/openstenoproject/plover/issues/1559).

sammdot (approved):

sol:
@sammdot thanks for merging this.  I had more improvements in the pipeline, but the inertia of this project let me eventually give up on it, and on steno as a whole.

The thing is that if things are fundamentally broken, there have been bug reports for years and several PRs that address these, but they don't get merged, then from a user  perspective that's not very attractive. I certainly was not up for maintaining my own fork indefinitely + upstream did not work for me.

sol:
Also, don't understand me wrong, I have patience + I completely understand if nobody has time to look at some thing. I myself suffer from RSI and I have not been able to keep up with PRs to my projects in the past.  I even chose to ignore PRs that I considered not crucial, or that had severe  issues, like no tests or other code issues.

I tried very hard to make it as easy as possible for an interested maintainer to merge this by keeping it as focused as possible.

It was only after I realized that  people are actually looking at this, and still choose not to merge it, after I realize that other PRs get merged, but not this one, when I decide to cut my looses and move on.

sammdot:
I understand the frustration. 😞 It's not exactly easy not having enough bandwidth (especially after the two lead devs stepped down) to take care of this on top of all my existing commitments, and as much as I would have liked to contribute more, even the process to set up a dev environment was painful for me, and I only just got more permissions to review and merge stuff a few days ago.

I can't promise that I'll be especially active moving forward, but I'm doing my best in my current situation.

sammdot:
(For what it's worth, having ~~had this project dropped on me~~ officially picked this project up just recently, I didn't exactly have a proper system for triaging PRs; the main perceived difficulty with this one was deciding between this and #1054, and being unable to test either locally I had to rely on CI-built artifacts. This then broke the build which I had to patch in #1634, etc. It was a whole thing 🙃)

sol:
@sammdot to be clear, that criticism was not directed towards you.

PR #1571, opened 2022-11-12
Fix "Add Translation" formatted translation & raw strokes always using a white background color

### LINKED ISSUE #1569: Some of the text in "Add Translation" doesn't use the style sheet's background color
**Describe the bug**

The strokes & translation text doesn't use the background color of the text field, and instead renders with a pure white background.

**To Reproduce**

1. Add a `plover.qss` stylesheet to the config directory with the following content: `QTextEdit { background-color: #1d1d1d; }`
3. Start Plover
4. Open the "Add Translation" window
5. Fill in both of the strokes and translation text fields

**Expected behavior**

The strokes & translations should inherit the provided background color, instead of rendering as pure white.

**Screenshots**

image

**Operating system**

 - OS: Windows 10
 - Plover Version 4.0.0-dev12

**Hardware**

N/A

DropDemBits:
## Summary of changes

Setting the stylesheet resets the palette colors to the system default. For the base color (which was used by the span for the formatted translation & raw stroke text), that'd mean that it'd be reset to `#ffffff`.

While setting the palette from the stylesheet could be done, it's much simpler to instead let the formatted text inherit the background color (i.e. removing the explicit setting of the background color).

Closes #1569

### Pull Request Checklist
- [ ] ~~Changes have tests~~ N/A, UI change

### DIFF
--- a/news.d/bugfix/1571.ui.md
+++ b/news.d/bugfix/1571.ui.md
@@ -1,0 +1,1 @@
+Fix "add translation" dialog ignoring the stylesheet's background color for the translation and stroke text.
--- a/plover/gui_qt/add_translation_widget.py
+++ b/plover/gui_qt/add_translation_widget.py
@@ -42,14 +42,12 @@ class AddTranslationWidget(QWidget, Ui_AddTranslationWidget):
 
         self._special_fmt = (
             '<span style="' +
-            'background-color:' + self.palette().base().color().name() +';' +
             'font-family:monospace;' +
             '">%s</span>'
         )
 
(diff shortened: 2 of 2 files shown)


user202729:
For reference the commit that adds that part was 9f1bdde658f8887e10d651961d5c2eafe9d8742d. Seems there's no specific rationale for the background part in particular.

Side note, even with that removed the style is still hard coded i.e. not configurable with qss I think? Probably not a big issue either way

user202729 (approved):

sammdot (approved):

PR #1584, opened 2022-12-19
Add Option for Inserting Delay between Key Press Events

DropDemBits:
## Summary of changes

On Pop_OS! 22.04, typing in Firefox often drops backspaces when undoing words (likely due to #1030). While it doesn't fix that issue, this is a suitable workaround in the mean time (there are also other input issues for other users on other OSes that are fixed by adding a delay between key presses).

Supersedes #1132 (see [this comment](https://github.com/openstenoproject/plover/pull/1132#issuecomment-1229173322)).

Also keeps the same option name as the original PR so that anyone using the old PR can use this one without having to change anything.

### Notable changes from the original PR

- Also includes [this change](https://github.com/user202729/plover/commit/f4cb21c5f6d178b06340a77c68629d370d0b5944) to delay when changing keymaps in Linux.

- Adds `set_key_press_delay` to Output. When not implemented it currently raises a `NotImplementedError`, although for some output interfaces it may be okay to just leave it as a stub.

Tested on:

- [x] Linux
- [ ] macOS
- [ ] Windows

### Pull Request Checklist
- [x] Changes have tests
    - Note: only for if an int option is missing, not for validating if the delay option is not negative

### DIFF
--- a/plover/config.py
+++ b/plover/config.py
@@ -105,12 +105,14 @@ def json_option(name, default, section, option, validate):
 def int_option(name, default, minimum, maximum, section, option=None):
     option = option or name
     def getter(config, key):
-        return config._config[section].getint(option)
+        return config._config[section][option]
     def setter(config, key, value):
         config._set(section, option, str(value))
     def validate(config, key, value):
-        if not isinstance(value, int):
-            raise InvalidConfigOption(value, default)
+        try:
+            value = int(value)
+        except ValueError as e:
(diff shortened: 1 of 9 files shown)


user202729:
(while this was the approach that the original code use also) It feels like redundant duplication of code to repeat the delay code in Windows/Linux/Mac. May be a better idea to handle it in the layer above.

PR #1611, opened 2023-06-27
First-up chord send for keyboard machine

user202729:
## Summary of changes

See title.

Refer to http://plover.stenoknight.com/2021/09/qmk-extensions-for-hobbyist-stenoboards.html for some detail what first-up chord send is -- basically this allow you to, e.g. if you have STPH-G bound to [→] (`{#Right}{^}`), then you can hold down STPH and repeatedly press -G to press [→] multiple times.

--------

Later it may be useful for things like "repeatedly send a chord with frequency X if it's held for longer than time Y", so making a check box doesn't feel like the best idea.

Currently it gives an error (that can only be fixed by clicking the refresh button) if both options are checked. Maybe it's better to either make a combo box or make the GUI automatically untick the other button if one is ticked (this may be confusing though)?

There's a quirk: currently if I hold down PHRO, then press -LG repeatedly, it should toggle the machine repeatedly, but the problem is it deletes 2 characters on each subsequent enable (when the other keys are not released then somehow the un-suppress does not "register", not sure). I'm not sure what the situation is on other platforms, but if it is then some modification would be in order.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/news.d/feature/1611.core.md
+++ b/news.d/feature/1611.core.md
@@ -1,0 +1,1 @@
+Implement first-up chord send for keyboard machine.
--- a/plover/gui_qt/config_keyboard_widget.ui
+++ b/plover/gui_qt/config_keyboard_widget.ui
@@ -6,14 +6,25 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>117</width>
-    <height>38</height>
+    <width>159</width>
+    <height>66</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string notr="true"/>
   </property>
-  <layout class="QGridLayout" name="gridLayout">
+  <layout class="QGridLayout">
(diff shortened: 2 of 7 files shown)


Robberduckzilla:
Tested on M2 Mac (Mac os 13.1) and seems to be working fine.
(I have merged this with Greg's Plover5.0/pyqt6 branch [on plover5-firstup in my own plover fork] for testing on M2 mac)

> There's a quirk: currently if I hold down PHRO, then press -LG repeatedly, it should toggle the machine repeatedly, but the problem is it deletes 2 characters on each subsequent enable (when the other keys are not released then somehow the un-suppress does not "register", not sure)

This is not the case for me when testing as above - it toggles machine on and off without any issues.

sammdot (approved):

PR #1636, opened 2023-09-28
Close serial port when disconnected to ensure clean reconnection

sammdot:
In addition to the changes from #1560, this PR closes a serial port on error to handle this issue from #1054:
> Now an exception in the run method is reported as an error to the system
and additionally the SerialStenotypeBase class does close the com port on
error.
> 
> **Otherwise the com port resource might still be taken by plover upon
reconnect of the writer, and a different com port might be taken by the OS
and plover might not be able to reconnect to that machine.**

Closes #1054.

- [x] News fragment added in news.d

### DIFF
--- a/news.d/bugfix/1636.core.md
+++ b/news.d/bugfix/1636.core.md
@@ -1,0 +1,1 @@
+Closes serial ports upon disconnection to ensure clean reconnections.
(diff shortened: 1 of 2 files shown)


pianohacker:
I've been running a fork to work around this issue for literal years. Thank you _so much_.

PR #1658, opened 2024-02-05
Allow system plugins to customize stroke display

lynn:
## Summary of changes

System definitions/plugins can now define a function `display`, which is used to render outlines in the UI (suggestions, lookup, dictionary view, etc).

For example, users of Lapwing theory (which uses `#` for proper nouns) may enjoy an override that doesn't convert digits:

```py
# english_stenotype_no_digits.py

from plover.system.english_stenotype import *

UNDIGIT_TABLE = str.maketrans('1234567890', 'STPHAOFPLT')

def display_stroke(stroke):
    undigit = stroke.translate(UNDIGIT_TABLE)
    return stroke if stroke == undigit else '#' + undigit

def display(strokes):
    return '/'.join(map(display_stroke, strokes))
```

Of course, this example shows that this isn't quite the right abstraction: `English Stenotype` and `English Stenotype (no digits)` are not really two different steno systems. But tying the display function to the system keeps things architecturally simple, and it's easy for power users to edit the `display` function in the file that defines their system.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/doc/api/system.md
+++ b/doc/api/system.md
@@ -254,6 +254,17 @@ Each line in this word list consists of a word and a number, separated by
 a space. See {data}`ORTHOGRAPHY_WORDS` for more information.
 ```
 
+```{data} display
+:type: Callable[[Tuple[str, ...]], str]
+
+A function called to display a steno outline in this system. The input is a
+tuple of normalized strokes. The output is a string shown to the user when
+making stroke suggestions or browsing a dictionary.
+
+If not defined, the default is `"/".join`, i.e. simply join the normalized
+strokes by slashes.
+```
+
 ## Computed Fields
 
 The fields below are automatically calculated from the values defined by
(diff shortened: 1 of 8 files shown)


mkrnr:
Hey @lynn, thanks for this PR!

I was thinking a bit about this, also related to your note above with the abstraction.

Maybe first of all: Is the Lapwing example still relevant? By setting the `NUMBER_KEY = None` and `NUMBERS = {}` (see [here](https://github.com/aerickt/plover-lapwing-aio/blob/d4516a9a85b6d858e4be5175efa45b664e27bb2d/plover_lapwing/system.py#L3-L5)), this should be resolved for the Lapwing system?

Alternative ways of displaying, for example `#L*IN` instead of `#HR*EUPB`, might be difficult to implement since there will always be "false positives", for example in briefs.

What are your thoughts?

mkrnr:
Closing this PR, see comment above.

PR #1663, opened 2024-02-18
Add keyboard emulation init after machine reset

### LINKED ISSUE #1585: Disconnecting and reconnecting the machine causes some diacritic entries to stop working.
**Describe the bug**

Disconnecting the steno machine (tested on Gemini PR and keyboard with a Multisteno) will cause letters with diacritics to not be outputted by Plover.

**To Reproduce**

Steps to reproduce the behaviour:

1. Add the following entry:
```
"A": "người",
```
2. Writing `A` will output the translation as expected (`người`)
3. Unplug the steno machine
4. Plug the steno machine back in
5. Press the "reconnect machine button"
6. Writing `A` will output `ngi`

Additionally, trying to undo this stroke with `*` will still result in 6 characters being deleted.

It appears that this bug only affects previously written translation. For example, skipping step 2 in the above process won't result in the malformed translation in step 6. I'm using [a Vietnamese dictionary](https://raw.githubusercontent.com/aerickt/steno-dictionaries/main/viet.json) and this also shows up there: writing words with diacritics will result in the above behaviour only if they had been written before the machine was reconnected.

Restarting Plover will fix this issue temporarily.

**Operating system**

 - Fedora 37 running GNOME 43.2 on x.org
 - Plover v4.0.0.dev12

Cl4ryty:
## Summary of changes

Disconnecting and reconnecting the machine causes certain characters to not be output in the translation anymore. Specifically, this affects all characters that are not in the default keymap and that have been used output by Plover, meaning that Plover has added them to the keymap, before the machine was disconnected. 
Unplugging and then plugging in a USB device causes the operating system to reset the keymap. To avoid Plover sending keypresses for keys it had previously configured but which are not mapped after the reset (leading to empty outputs) it is necessary to reset the keyboard emulator when resetting the machine after reconnecting so that Plover assumes the default keymap and any unmapped symbols will be added again.

Closes #1585

### Pull Request Checklist
- [ ] Changes have tests


user202729:
This will not fix all the cases where the internal and external keymap are different (e.g. the user manually reset the keymap), but I suppose it's an improvement.

There's also the case where the keymap of different keyboards are different (I haven't thought carefully about this possibility yet)

Cl4ryty:
You're right, this is only aimed to deal with this issue in the context of dis- and reconnecting the machine, because that definitely leads to a change in the keymap and can easily be addressed. For the other cases it would be necessary to listen for keymap changes and then react accordingly, but that is of course more complicated and OS specific. It's definitely something that should be addressed at some point, but I think that for now this change at least improves the situation, fixing the problem in a situation that people commonly encounter.

The case of different keymaps for different keyboards/machines is interesting and worth a look, although I can imagine that it would only matter when using plover with the machine set to use a keyboard as input because for all other protocols I think it can only be connected to one machine. 

I'm not sure I understand what you mean with the engine being refreshed "automatically" when the keyboard is plugged in, can you clarify that?

user202729:
I was thinking about something like https://github.com/openstenoproject/plover_plugins_registry/pull/46 , but looks like at the moment it doesn't matter for the keyboard machine.

(Later on, there may be feature to filter to only keyboards with specific names, so when that keyboard is disconnected and reconnected this would apply?)

Cl4ryty:
That plug-in is just calling the same function that is used when reconnecting manually once the serial port that was previously used becomes available, so this fix would also apply then. If the filtering for a specific keyboard is done by selecting a port according to the device id that would still entail reconnecting in the same way by calling the reset machine function as soon as the port with the associated device id becomes available. Although, I have to admit that I'm not sure if filtering serial ports by device id is something that exists for Windows or OSX, but then it's also a hypothetical future feature and at the moment the keyboard machine is different from the other machines configurable in plover

mkrnr (approved):
Thanks for the fix! LGTM

mkrnr:
Sadly I had to revert this commit since it caused issues on macOS because of the `_layout_watcher` in `KeyboardLayout` which now crashes due to the second __init__ of that class.

So this change needs more manual testing across all operating systems and some solution for the macOS issue.

Feel free to open another PR.

PR #1680, opened 2024-07-19
Update +ify orthography rule

### LINKED ISSUE #1678: [Grammar] Stroking PREUT/TPAOEU outputs 'prettyify' even though the correct form is "prettify"
**Describe the bug**
When I stroke PREUT/TPAOEU, I get 'prettyify' even though the correct form is "prettify". I suspact that there's a mistake in the grammar rules that does not account for this case.

**To Reproduce**
1. Stroke PREUT/TPAOEU

**Expected behavior**
I get the word "prettify"

**Screenshots**
image
image

**Operating system**
 - OS: Linux Mint 21.3 Cinnamon
 - Plover Version: Plover 4.0.0rc2

Shrimpy48:
## Summary of changes

Orthography rule for words ending in i + ify suffix updated to also cover words ending in y.

Closes #1678 

### Pull Request Checklist
- [x] Changes have tests


Shrimpy48:
I'm not sure which category of news fragment this should have - dict feels the closest.

mkrnr:
Hey @Shrimpy48 — thanks a ton for opening this!
(And sorry it’s taken me a while to get to it.)

I'm afraid the PR needs to be updated to resolve the conflicts. We very recently added code formatting with Ruff.

While we're touching the rule, might make sense to further extend the regex so it only fires when the final i/y is preceded by a consonant? Something like:
`(r"^(.+[^aeiou])[iy] \^ if(y|ying|ied|ies|ication|ications)$", r"\1if\2")`
Examples that we can add to the tests would be:
* Hawaii + -ification should be Hawaiification
* play + -ification should be playification

If you’re busy or no longer working on this, just let me know — I’m happy to pick it up and push it over the line. Either way, really appreciate the contribution!

Shrimpy48:
Hey there, thanks for getting back to me on this! No worries about the delay.
I'm happy to carry on with it.

The preceding vowel case is tricky. Maybe the rule should fire for vowel+"i" (Hawaiification, ASCIIfication, Dubaification) but not for vowel+"y" (playification, joyification)?

mkrnr:
Yes, that sounds good. There'll always be exceptions to this (and many other) rule but it will cover the majority of cases.

mkrnr:
The changes look very good, thanks for that! Regarding the news file, I'd vote for "core" to make it more clear that it's changing the system rules and not adding some entries to the dictionary.

Shrimpy48:
OK, I've added a news fragment now, thank you!

Issue #904, opened 2018-02-28
Include libusb in the Mac App bundle


morinted:
## Summary

The [Stenograph USB](https://github.com/morinted/plover_stenograph_usb) plugin requires translation of third-party library libusb to work on Mac. On Windows, the user uses the manufacturer's driver, on Linux a compatible backend is usually installed.

## Implementation Notes

I've been playing with trying to get this to work on my local machine. It seems it's not possible to point the python executable in the app bundle to libusb using the preferred method, because pyusb uses [find_library()](https://docs.python.org/3/library/ctypes.html#finding-shared-libraries) which just checks certain directories.

It might be possible to add our app's bundle to the checked directories by setting `$DYLD_LIBRARY_PATH` to the app's Frameworks folder with the libusb dylib in it.

I also found that the dylib has some static links which might be good to set after moving the libusb file: `install_name_tool -id "@executable_path/libusb.dylib" Plover.app/Contents/Frameworks/libusb.dylib` (though it's not exactly clear to me what the benefit is, here)

The other option if we can't control find_library with environment variables is to provide a function to pyusb that knows where to look. The disadvantage here is that the code would move into the plugin.

morinted:
I modified the launch script to have:

`DYLD_LIBRARY_PATH="$DIR/../Frameworks" exec "$DIR/../Frameworks/python3.6" -m plover.main "$@"`

with libusb.dylib copied into Frameworks, and it appears to work.

benoit-pierre:
Regarding static links: doesn't running macholib takes care of that? As for the `DYLD_LIBRARY_PATH` it looks like the equivalent of `LD_LIBRARY_PATH` on Linux, if it works, then let's use it.

morinted:
Macholib fixes the links within the python3.6 executable. I don't believe
that it handles direct dylibs that are loaded dynamically at runtime.

morinted:
Okay, that does look promising, just copy the libusb Dylib, run macholib,
and it copies it over and fixes the references? I will be trying that out.

benoit-pierre:
Yep, exactly.

morinted:
Okay, so I've got my proof-of-concept going

```diff
diff --git a/osx/app_resources/plover_launcher.sh b/osx/app_resources/plover_launcher.sh
index 91fba813..5be904aa 100644
--- a/osx/app_resources/plover_launcher.sh
+++ b/osx/app_resources/plover_launcher.sh
@@ -1,5 +1,5 @@
[...]

PR #1693, opened 2024-12-09
Add support for the HID Protocol.

CrazyH2:
## Summary of changes
This PR adds a new protocol called Hid which works over BLE and USB.
Other protocols can't be used over BLE which this aims to solve.

Closes 1693

### Pull Request Checklist
- [x] HID protocol added.
- [x] HID protocol documented.


gdwaynewarner:
I'm pretty sure I'm not alone when I say I have no idea what this is supposed to do.  Lots of acronyms without specifying what they stand for doesn't help anybody.

user202729:
If I understood correctly this is just copy pasted from the plugin https://github.com/dnaq/plover-machine-hid/blob/main/plover_machine_hid.py (with the goal to make it available by default in Plover).

Technically nothing wrong (the original repository uses MIT license which allows code to be used in HID)

petercpark:
I support this PR. Let's get this merged!

dnaq:
This PR needs adding the hidapi libraries to the release files as well for it to work.

morinted:
What about adding the plugin to the default plugins and shipping it with releases? Same effect, but keeps the code ownership and parallel development intact?

morinted:
Yes, if I remember correctly, that was the original plan. I was having trouble bundling HID with the Mac release—someone would need to take that on.

lynn:
You'll also want to copy over [hid_option.py](https://github.com/dnaq/plover-machine-hid/blob/main/hid_option.py), so that you can configure things like repeat delay from the Plover UI. It'll look like this:

image

petercpark:
> Yes, if I remember correctly, that was the original plan. I was having trouble bundling HID with the Mac release—someone would need to take that on.

@morinted I have a Mac, is there any way I can help?

pagedpenguin251:
Any update on this? (I support this PR)

CrazyH2:
Yeah, any updates?

pagedpenguin251:
@TheaMorin  What is needed to be done for this to progress further?

lynn:
> You'll also want to copy over [hid_option.py](https://github.com/dnaq/plover-machine-hid/blob/main/hid_option.py), so that you can configure things like repeat delay from the Plover UI. It'll look like this:
> 
> image

@CrazyH2 Sorry to repeat my comment, but I really do think you should update the PR with this. You'll also need to add a line to `setup.cfg` to include this file, something like

```
plover.gui.qt.machine_option =
  plover.machine.hid:HidMachine = plover.machine.hid_option:HidOption
```

Otherwise the protocol won't be configurable from the GUI.

dnaq:
I would say, do not merge this PR. Bundle the official plugin as well as hidapi in the default plugins instead as @TheaMorin suggested. This PR should be closed.

On a side note I find it a bit annoying that someone would rip out code I have written without giving any credit whatsoever. But I have full confidence in that that is not how the plover project wants to work.

TheaMorin:
Hopefully it was more ignorance than malicious intent!

Maybe I'm wrong and we already addressed the driver issue: https://github.com/openstenoproject/plover/issues/904 (Sorry, it's been a minute since I was in the weeds on this.)

If so, bundling stenograph and HID support should "just work"? Someone would have to add it to the default plugins list, make a build, and have people test it out.

PR #1698, opened 2025-01-28
Fix GitHub actions

mkrnr:
## Summary of changes

Fixes the currently broken GitHub actions (see e.g. [this run](https://github.com/openstenoproject/plover/actions/runs/12678789290)):

1. Bump the GitHub action runner macOS version since macOS-12 is deprecated, resulting in the test being skipped
2. Bump the ubuntu version to 24.04 since it's not causing issues (Windows 2025 does create issues so leaving that out for now)
3. Bump twine to 3.7.0 to support metadata 2.2 which is otherwise breaking the packaging build
4. Bump pkginfo to 1.8.1 which is required by twine 3.7.0
5. Bump the GitHub actions versions to remove deprecation warnings
6. Update when the GitHub actions are triggered to prevent double executions (both triggered by `push` and `pull_request`)

I don't think it makes sense to add a news fragment since this is a technical change related to CI/CD.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,15 @@ name: CI
 on:
   push:
     branches:
-      - '*'
+      - 'main'
     tags-ignore:
       # Ignore continuous releases' tag.
       - 'continuous'
   pull_request:
+    types:
+        - opened
+        - reopened
+        - synchronize
 
 defaults:
   run:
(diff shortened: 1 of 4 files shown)


greenwyrt (approved):
since the change in #1687 doesn't make sense (now closed), I don't think there are any potential conflicts

PR #1702, opened 2025-02-22
Update hardware_communication.md with Stenograph USB format info from…

ThaiWood:
… plugin docs

## Summary of changes

This is the explanation of the protocol extracted from the [stenograph_usb](https://github.com/TheaMorin/plover_stenograph_usb) plugin, just with `Mira` replaced with writer since it applies to all of them.

Somewhere I have the doc where this was originally copied from and a draft of some extra abilities (e.g. packet IDs to do different things), but I figure this is a good start.

### DIFF
--- a/doc/hardware_communication.md
+++ b/doc/hardware_communication.md
@@ -91,8 +91,57 @@ class MySerialMachine(SerialStenotypeBase):
 
 ## USB-based Protocols
 
-```{todo}
-Complete this section.
+### Stenograph USB
+```
+Packet format:
+--------------
+Name:     Size:       Value Range:
+----------------------------------
+Sync          2 bytes "SG"
+Sequence #    4 bytes 0 - 0xFFFFFFFF (will wrap from 0xFFFFFFFF to 0x00000000)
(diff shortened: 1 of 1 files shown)


mkrnr:
Thank you so much @ThaiWood for the contribution!

I'm wondering, however, whether we want to include this info in the plover documentation since it's regarding logic of a plugin and not plover itself. To me it feels more natural to link to the plugin documentation instead.

@sammdot what are your thoughts on this?

sammdot:
Agree on this being included in the plugin documentation, at least *for now*. If we ever decide to bring this and other machine protocols into Plover's codebase then sure, but for right now I think keep it with the plugin.

ThaiWood:
I think I'm missing something here.  Is the disucssion to move this section or the hardware communication doc?  If the former, what would go under the USB-based Protocols section if not things like the packet format of a USB protocol?

mkrnr:
Sorry for the confusion! We should only include USB protocols in this  documentation that are implemented in this repo which is not the case for stenograph.

Issue #1703, opened 2025-03-04
pkg_resources is deprecated as an API


user202729:
**Describe the bug**
Currently `pkg_resources` is used by `plover.registry`, however it's deprecated

```
DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```

although this is mostly invisible to the user, it's better to refactor it out.

PR #1704, opened 2025-03-04
Remove deprecated extra gui_qt from setup.cfg

user202729:
According to https://github.com/pypa/setuptools/issues/4862 , specifying extra for entrypoint is deprecated. And as I explained there, it breaks some features (in particular if that is used then gui_qt no longer work on my machine).

The simplest way seems to be just remove the extra specification.

### Pull Request Checklist
- [ ] Changes have tests


mkrnr (approved):
LGTM 👍 

Also did a quick test on the PySide6 branch and removing this didn't seem to have any impact.

Feel free to merge the PR.

tsprlng:
This change seems to break local builds on MacOS, causing the `check_requirements` step at the end of `osx/make_app.sh` to fail with  `pkg_resources.UnknownExtra: plover 4.0.0 has no such extra feature 'gui_qt'.` -- not sure why.

tsprlng:
(Is this PySide6 branch up anywhere, out of interest?)

user202729:
Understandable, given Plover uses quite a lot of `pkg_resources` sanity check internally. (Sometimes the sanity check is a bit too strict though. This one probably qualifies.)

See also https://github.com/openstenoproject/plover/issues/1703

Given the above, and the fact that dependency on entry point itself is going to be deprecated, it's probably sufficient if we just remove the relevant check.

PR #1711, opened 2025-04-04
On linux wayland, warn if ibus or fcitx5 isn't running.

### LINKED ISSUE #1709: incorrect input on wayland
**Describe the bug**

some combinations of inputs trigger incorrect output. for example H-T outputs numbers, instead of the expected symbol `?`

**To Reproduce**

1. Open KDE wayland session 
2. Open a native wayland program, e.g. Kate or telegram 
3. type H-T
4. see output

**Expected behavior**

By default should produce the question mark symbol

actual behaviour is 

**Screenshots**

Image

**Operating system**

 - OS: Arch Linux, up to date as of writing
 - Plover Version 4.0 Appimage (other versions not installable yet)

**Hardware**

confidential

752986:
## Summary of changes
When initializing uinput keyboard emulation, this runs the `ps` command to check whether ibus or fcitx5 is running. If not, it logs a warning.

It doesn't strike me as a super efficient or elegant way to acheive this, but it's the only way I could think of for fcitx5. ibus has the `ibus engine` command, but if I'm scanning every process anyway, I see no reason to split the implementation. It also makes it easy to add other supported input methods later.

Closes #1709 

### Pull Request Checklist
- [ ] Changes have tests
    - I'm not really sure how to test this, since it's platform specific and depends on other running processes.

### DIFF
--- a/news.d/feature/1711.linux.md
+++ b/news.d/feature/1711.linux.md
@@ -1,0 +1,1 @@
+On linux wayland, warn if ibus or fcitx5 isn't running.
--- a/plover/oslayer/linux/keyboardcontrol_uinput.py
+++ b/plover/oslayer/linux/keyboardcontrol_uinput.py
@@ -1,6 +1,7 @@
 from evdev import UInput, ecodes as e, util, InputDevice, list_devices
 import threading
 from select import select
+import subprocess
 
 from plover.output.keyboard import GenericKeyboardEmulation
 from plover.machine.keyboard_capture import Capture
@@ -271,6 +272,14 @@ class KeyboardEmulation(GenericKeyboardEmulation):
         # Initialize UInput with all keys available
         self._res = util.find_ecodes_by_regex(r"KEY_.*")
         self._ui = UInput(self._res)
+
+        # Check that ibus or fcitx5 is running
+        processes = subprocess.run(["ps", "x"], capture_output=True)
+        for line in processes.stdout.decode().splitlines():
+            if line.endswith(("ibus", "fcitx5")):
+                break
+        else:
+            log.warning("It appears that an input method, such as ibus or fcitx5, is not running on your system. Without this, some text may not be output correctly.")
 
     def _update_layout(self, layout):
         if not layout in LAYOUTS:


mkrnr:
Thanks a lot @752986 for your contribution!

The build was failing because it started using a version of setuptools with metadata 2.4 which we don't support. One solution would be to bump our dependencies but I wouldn't want that in this PR (or plover 4 for that matter). So I took the liberty to set an upper limit for setuptools in the `pyproject.toml`.

I'd like to ask @LilleAila for a code review if you have time 🙂

user202729:
* I'd rather `if not any(line.endswith(…) for line in …):`. Both shorter and equally clean.
* `ps -e -o cmd --no-headers`.
* Check if user doesn't have `ps` installed. Or if `ps` fails.
* in the meantime consider using `psutil`? Extra dependency though.

752986:
- Agreed on `any(...)`
- Agreed, although I'm thinking of using `-o comm` instead and then matching the entire line with exact equality.
- Since `ps` is a posix command, I'd expect it to be present on pretty much every linux install in existence. I'll at a try/except to catch it though, just in case. I'll also add a check for a non-zero return code.
- Since this is only an issue on linux on the first place, I think something cross-platform like psutil seems like overkill.

appetrosyan:
+1 re: `psutil`. 

The `ps` program has a lot of non-trivial interactions on different systems. Pure Python implementations almost always result in more predictable outcomes.

LilleAila:
I also think this should use psutil instead of calling ps. Otherwise, LGTM.

user202729:
Technically psutil uses some compiled C library too right… it's not exactly pure Python like that.

On the other hand psutil is sufficiently popular that it's likely it won't introduce much issue if you choose that way.

Performance-wise it probably doesn't matter since it's only ever invoked once at start anyway.

752986:
Hmm, I'm guessing I maybe didn't add the psutil dependency correctly (I just added it to `reqs/dist.txt`). There's also a chance I messed something up when I tried to resolve the merge conflict. Any thoughts on why it's failing?

mkrnr (approved):
Thanks a lot for your contribution! 🙂
LGTM

Issue #1714, opened 2025-04-21
Can't Access Plugins Manager


Viinnnn:
**Describe your issue**

The Plugin Manager doesn't load (Error Message below)

**Steps you've taken**

1. Actually just installing Plover 4.0.0 - I uninstalled and reinstalled Plover after this happened for the first time, but it didn't seem to fix the issue. Either Plover doesn't actually delete everthing it stores on the PC, or its a bug with the main version?
2. Click on 'Plugin Manager'
3. Fails to load

**Screenshots**

![Image](https://github.com/user-attachments/assets/ec822575-faa1-4536-bd78-1536861a15a1)

This is the error message. Any idea what it means? I can't decipher it haha :D

**Operating system**

 - OS: Windows 10
 - Plover Version 4.0.0

**Hardware**

- The Polyglot (Gemini PR) - but disconnecting it didn't seem to change anything.
- If it's of note I did install the Display Plugin and the Lapwing-aio, however the error stays even when they are not installed anymore.

Any idea what I can do?

greenwyrt:
Which specific version of Plover are you using? Is this the latest continuous build or [v4.0.0rc5](https://github.com/openstenoproject/plover/releases/tag/v4.0.0rc5)?

Viinnnn:
I'm using Plover 4.0.0

greenwyrt:
Did you have any plugins installed before you uninstalled Plover the first time? Plover keeps dictionaries and installed plugins in the AppData folder for Windows and it may be having a problem trying to load the existing plugins. 

One possibility is deleting/relocating the plover folder and then starting Plover.

Grahp:
This may have been fixed in v4.0.1, my plugins manager problems seem to have gone away after updating.

dylanwulf:
I think I found the issue here. It looks like Plover uses an older version of `pkginfo` which only supports `Metadata-Version <= 2.2`, but `plover-lapwing-aio` is built with `Metadata-Version = 2.4`.

mkrnr:
Thanks @dylanwulf for fixing the issue in `plover-lapwing-aio`!

As you wrote, in Plover v5 we support Metadata-Version 2.4 and the latest setuptools. The first dev version of Plover v5 should be released in the next few weeks.

Unless more people run into this issue, we won't fix it for Plover v4.

Closing the ticket since a fix is already on main and just needs to be released.

PR #1720, opened 2025-04-24
Bump python to 3.13 and update dependencies

mkrnr:
## Summary of changes

Updates the python version that is packaged into the Plover installers to 3.13 and updates all python dependencies to the latest supported versions.

### Pull Request Checklist
- [ ] Changes have tests


greenwyrt:
Could we resolve #1714  with the version/dependency update?

mkrnr:
As @Viinnnn wrote in #1714, the pkginfo version is already updated on main. So I'd say we can close #1714 and should soon release the first dev version of v5.0.0. My goal is to do that after merging this PR.

Issue #1707, opened 2025-03-14
uinput output method doesn't completely work yet


user202729:
**Describe the bug**

I use Plover with latest version merged (including https://github.com/openstenoproject/plover/pull/1679) 

Then I modify `/plover/oslayer/linux/keyboardcontrol.py` from `if DISPLAY_SERVER == "x11":` to `if False:` to force using uinput (since I'm on x11).

I see a few issues

1. normally the display shows "connected", but in uinput it shows "disconnected"

![Image](https://github.com/user-attachments/assets/6ed08b67-c41b-4c5d-a217-dd1d2409404a)

2. in one case I type a stroke defined as `{#backslash}` and get

```
 [Dummy-1] WARNING: Key backslash is not valid!
```

3. in another case I type a stroke defined as `{^}+{^}` and get `2b`. Presumably because I don't have ctrl-U.

https://github.com/openstenoproject/plover/pull/1679/files#diff-6b4281e6a4006cf3e22f4cfbcaf9fcc8ff7337cf721c33d5051f1e6cbd369844R296-R303

In my plugin I implement some feature to try to get a keycode if possible:

https://github.com/user202729/plover-uinput-output/blob/main/plover_uinput_output/__init__.py#L111-L122

That said, I think it's a good idea to try to use "native" key as much as possible, because there may be some application that cares about the scan code as well as keycode. 

**Operating system**

 - OS: Arch Linux, X11
 - Plover Version: latest master

user202729:
4. when the user e.g. hold Ctrl and press Q, elsewhere the Q is passed through verbatim. Here the Q is still interpreted

PR #1721, opened 2025-04-27
Add more keys to uinput output

user202729:
As mentioned in https://github.com/openstenoproject/plover/issues/1707 , this tries to use hardware key as much as possible.

Things like `?` or `*` (which requires shift key) doesn't work yet. I haven't tested with layouts other than Qwerty.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/oslayer/linux/keyboardcontrol_uinput.py
+++ b/plover/oslayer/linux/keyboardcontrol_uinput.py
@@ -38,8 +38,6 @@ BASE_LAYOUT = {
     "0": e.KEY_0,
     # Symbols
     " ": e.KEY_SPACE,
-    ".": e.KEY_DOT,
-    ",": e.KEY_COMMA,
     "\b": e.KEY_BACKSPACE,
     "\n": e.KEY_ENTER,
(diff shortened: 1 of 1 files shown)


plover/oslayer/linux/keyboardcontrol_uinput.py
```diff
@@ -256,13 +276,7 @@
...
 }
 
-us_qwerty = {
-    **LAYOUTS[DEFAULT_LAYOUT],
-    ";": e.KEY_SEMICOLON,
-    "'": e.KEY_APOSTROPHE,
-    "[": e.KEY_LEFTBRACE,
-}
-KEYCODE_TO_KEY = dict(zip(us_qwerty.values(), us_qwerty.keys()))
+KEYCODE_TO_KEY = dict(zip(LAYOUTS[DEFAULT_LAYOUT].values(), LAYOUTS[DEFAULT_LAYOUT].keys()))
```
user202729, line 279: actually this doesn't even make sense, it should have just been a set and named `KEYCODES_TO_SUPPRESS` or something. But the original code is like that

mkrnr (approved):
Thanks for adding the keys! Also, nice improvement to `KEYCODE_TO_KEY`, getting rid of the `us_qwerty`.

PR #1728, opened 2025-05-15
Wait for all keys to be released before grabbing

iansw246:
## Summary of changes

Makes uinput module wait for all keys on each input device to be released before grabbing the device. This prevents keys from being stuck when keys are held while Plover is starting.

Device grabbing was moved from `_run` to `start` to show that the keyboard is still connecting in the UI while waiting for keys to be released (`self._running` is only set when keys are grabbed)

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/oslayer/linux/keyboardcontrol_uinput.py
+++ b/plover/oslayer/linux/keyboardcontrol_uinput.py
@@ -393,7 +393,27 @@ class KeyboardCapture(Capture):
         )
         return not is_uinput and keyboard_keys_present
 
+    def _grab_devices(self):
+        """Grab all devices, waiting for each device to stop having keys pressed.
+
+        If a device is grabbed when keys are being pressed, the key will
+        appear to be always pressed down until the device is ungrabbed and the
+        key is pressed again.
+        See https://stackoverflow.com/questions/41995349/why-does-ioctlfd-eviocgrab-1-cause-key-spam-sometimes
+        There is likely a race condition here between checking active keys and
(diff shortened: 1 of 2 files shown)


plover/oslayer/linux/keyboardcontrol_uinput.py
```diff
@@ -393,7 +393,27 @@ def _filter_devices(self, device):
...
+        """Grab all devices, waiting for each device to stop having keys pressed.
+
+        If a device is grabbed when keys are being pressed, the key will
+        appear to be always pressed down until the device is ungrabbed and the
+        key is pressed again.
+        See https://stackoverflow.com/questions/41995349/why-does-ioctlfd-eviocgrab-1-cause-key-spam-sometimes
+        There is likely a race condition here between checking active keys and
+        actually grabbing the device, but it appears to work fine.
+        """
+        for device in self._devices.values():
```
LilleAila, line 406: Should this be wrapped in a loop?

LilleAila (approved):
LGTM. haven't tried running the code locally, but it looks good.

plover/oslayer/linux/keyboardcontrol_uinput.py
iansw246: Once a device is grabbed, keys can't get stuck anymore, so there's no need to check a device again once grabbed. I don't think another loop around the existing `self._devices` for loop over is necessary.

PR #1729, opened 2025-05-21
UInput: ungrab devices on exception

iansw246:
## Summary of changes

Improve uinput exception handling by wrapping the keyboard capture loop in a try finally. All devices are ungrabbed in the finally block so that exceptions do not block all grabbed input devices. Currently, if an exception occurs in `_run`, the loop exits and no device inputs are handled. This issue can be reproduced by having multiple keyboards and unplugging one.

A possible UI improvement is to show an error/disconnected state when an exception occurs. I haven't figured a clean way to do this yet.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/oslayer/linux/keyboardcontrol_uinput.py
+++ b/plover/oslayer/linux/keyboardcontrol_uinput.py
@@ -411,8 +411,20 @@ class KeyboardCapture(Capture):
                         break
             device.grab()
 
+    def _ungrab_devices(self):
+        """Ungrab all devices. Handles all exceptions when ungrabbing."""
+        for device in self._devices.values():
+            try:
+                device.ungrab()
+            except:
+                log.debug("failed to ungrab device", exc_info=True)
 
     def start(self):
+        try:
+            self._grab_devices()
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


plover/oslayer/linux/keyboardcontrol_uinput.py
```diff
@@ -414,23 +412,35 @@ def suppress(self, suppressed_keys=()):
...
+                        self._ui.syn()
+        except:
+            log.error("keyboard capture error", exc_info=True)
+        finally:
+            # Always ungrab devices to prevent exceptions in the _run loop
+            # from causing grabbed input devices to be blocked
+            for device in self._devices.values():
+                try:
+                    device.ungrab()
+                except:
```
user202729, line 444: does the try-except have any value, I think a normally raised exception will get logged and made visible to the user anyway?

(also which exception can be raised here?)

mkrnr:
@iansw246 is this PR ready to be merged from your perspective? (If not, better mark it as Draft)

iansw246:
It's ready to be merged. I just wanted to note the UI improvements for future reference.

PR #1734, opened 2025-06-07
Resolve regex library warnings

emmanuel-ferdman:
## Summary of changes
This small PR resolves the annoying regex library warnings starting Python3.11:
```python
/tmp/plover/plover_build_utils/pyqt.py:9: DeprecationWarning: 'count' is passed as positional argument
```

### DIFF
--- a/plover_build_utils/pyqt.py
+++ b/plover_build_utils/pyqt.py
@@ -6,7 +6,7 @@ def gettext(contents):
     contents = re.sub(r'\n', (
         '\n'
         '_ = __import__(__package__.split(".", 1)[0])._\n'
-    ), contents, 1)
+    ), contents, count=1)
     contents = re.sub(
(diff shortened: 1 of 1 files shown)


mkrnr:
After looking into this some more, and checking your activity on GitHub, I need to be honest with you: I'm not a fan of your shotgun approach, creating all those tiny PRs all over the place to boost your GitHub stats. If more people would do this, we'd drown in such PRs. Closing this PR since the noise of this commit is not worth it.

user202729:
> Somehow I'm not running into it when building or running Plover with my setup.

I think you need Python ≥ 3.11 (I haven't tested this either)

mkrnr:
> > Somehow I'm not running into it when building or running Plover with my setup.
> 
> I think you need Python ≥ 3.11 (I haven't tested this either)

I'm using Python 3.13.3

Issue #1754, opened 2025-07-16
Key code not correctly sent in Windows


user202729:
**Describe the bug**

Visit https://keyjs.dev and type some keys e.g. `a` using normal keyboard versus Plover.

See that if Plover is used, `e.code` is empty, but if normal keyboard is used, `e.code` is `KeyA`.

This affects some things e.g. Plover doesn't work over Anydesk.

**Operating system**

 - OS: Windows 10.
 - Plover Version: 4.0.3.

PR #1755, opened 2025-07-16
Send correct scancode on Windows

user202729:
## Summary of changes

Closes https://github.com/openstenoproject/plover/issues/1754

The functions to calculate the mapping from virtual key code to scancode are already there, we just need to pass the correct values.

Manually tested on Windows 10 (after cherry-picked to v4.0.3).

### Pull Request Checklist
- [ ] Changes have tests


user202729:
actually looks like this may still fail for uppercase characters, needs more testing. still, this is better than the status quo.

edit; the issue with uppercase character appears to be `code` being `shift` instead of `shiftL` or `shiftR` which leads to the scan code being zero.

PR #1760, opened 2025-08-10
Uinput: Use pipe instead of select() with timeout

iansw246:
## Summary of changes

For Linux Uinput, signal the evdev thread to stop using a pipe. Currently, the thread periodically checks a variable to know when to stop.

This eliminates the slight delay between clicking the reconnect button and the reconnection actually happening.

Closes 

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/oslayer/linux/keyboardcontrol_uinput.py
+++ b/plover/oslayer/linux/keyboardcontrol_uinput.py
@@ -480,8 +480,8 @@ class KeyboardCapture(Capture):
                 log.debug("failed to ungrab device", exc_info=True)
 
     def start(self):
-        # Exception note: cancel() will eventually by called when the machine
-        # reconnect machine is pressed, or when the machine is changed.
+        # Exception handling note: cancel() will eventually be called when the
+        # machine reconnect button is pressed or when the machine is changed.
         # Therefore, cancel() does not need to be called in the except block.
         try:
             self._grab_devices()


plover/oslayer/linux/keyboardcontrol_uinput.py
```diff
@@ -1,6 +1,16 @@
-from evdev import UInput, ecodes as e, util, InputDevice, list_devices
+from evdev import (
+    UInput,
+    ecodes as e,
+    util,
+    InputDevice,
+    list_devices,
+    InputEvent,
```
mkrnr, line 7: Trivial: InputEvent and KeyEvent are not used and can be removed from the import.

mkrnr (approved):
Thanks for this improvement!

I tested it on Ubuntu and it works well.

GPT-5 flagged some potential lifecycle concerns, but given how this class is used, I don’t see any real issues.

Just a minor note about some unused imports.

Feel free to merge whenever you’re ready.

mkrnr (approved):

mkrnr:
@iansw246 are you allowed to merge on your own? If yes, feel free to do so. If not, let me know when ready and I merge.

iansw246:
I didn't see a merge button after you approved merging (before these new changes), so I assume am not able to merge myself.

I think it's ready to merge now.

mkrnr (approved):

user202729:
I guess the exception handling counts for defensive programming (a warning message and program carry on is better than Plover hangs/crashes)

there's a disadvantage with the extra exception handling though, it makes the code harder to read for developers, and the text says the exact same thing as the code (thus doesn't save the developer any time).

I don't know in which case can the exception be thrown at all (if it cannot be thrown the exception handling is obviously redundant), but maybe one case is race condition between the two threads? In that case it may be a better idea to communicate with queues instead, or locks if the above is insufficient.

PR #1769, opened 2025-09-22
Add code signing and notarization for macOS app

### LINKED ISSUE #1667: Please, sign macos app
Hello! Plover can not run on Mac by default due to security reasons (it is not signed, so MacOS cannot validate app's developer). However there is a workaround: you can add Plover to exception list so that OS still runs it (despite untrusted developer).
My problem is the following: my corporate policies forbid launching unsigned applications (there isn't even button like "still run potentially dangerous software"). So it seems there is no way to run Plover on my Mac.
If it is possible, please, sign the app.

mkrnr:
## Summary of changes

Add code signing and notarization for macOS app.

Code signing and notarization can be enabled via environment variable (`MACOS_NOTARIZE_ENABLED=1`) and will require a number of other environment variables as well, including the certificates.

Opening the app installed with the .dmg file from [this run](https://github.com/openstenoproject/plover/actions/runs/18029417130) looks like this:
image

In [this run](https://github.com/openstenoproject/plover/actions/runs/18029660516), the notarization steps were skipped which is the behavior when not pushing to `main`, `maintenance`, or a `v*` tag.

The tests were performed on the `main` environment but it's now changed to `production` to avoid confusion. After the PR is merged, I will delete the `main` environment.

Closes #1667 

### Pull Request Checklist
- [ ] Changes have tests


dnaq:
I think it might be prudent to separate the different jobs into different workflow files, and only have the release jobs run triggered on changes to the main branch. This would protect somewhat against malicious pull requests leaking the contents of the macOS dev certificate/private key (e.g. if there were to be an injection point in one of the CI jobs)

mkrnr:
Very good point! I was thinking that the PR action runs need approval but we shouldn't rely on that alone. I'll check how to best restrict the signing and notarization to main.

dnaq:
Might be easiest to create a release environment: https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments and only allow that environment to run on the main branch and move the secrets there

mkrnr:
Very good input! That's what I did. We already have a release environment and I'd keep that for actual tagged releases. So I added a `dev` and a `main` environment. The `main` environment is protected of course.

mkrnr:
Since I'm not a fan of deploying to a `dev` environment all the time, I'll extract the signing and notarizing into a separate job that only runs on `main`, `maintenance/*`, and tagged refs. I'll also rename the `main` environment to prevent confusion, maybe to `prod`.

PR #1773, opened 2025-09-29
Update URL's in index.html

JoshyLPHD:
## Summary of changes

-updated links which directed to the github wiki and not to the new one at plover.wiki
-changed one alt in the HTML as it mentioned github


mkrnr:
The deployment worked 🎉

PR #1777, opened 2025-10-25
Add Plover HID support based on plover-machine-hid plugin

mkrnr:
## Summary of changes

Adds support for the Plover HID protocol based on the great [plover-machine-hid](https://github.com/dnaq/plover-machine-hid) plugin by dnaq.

Also adds handling of multiple attached keyboards as well as hot-plugging of devices.

Like the original plugin, this uses the [hid](https://pypi.org/project/hid/) python module. This requires building [hidapi](https://github.com/libusb/hidapi) from source. I tried the [PyPI hidapi wrapper](https://pypi.org/project/hidapi/) but it's not well maintained and has a number of issues, namely:
* support for the `hid_darwin_set_open_exclusive` flag on macOS, see https://github.com/trezor/cython-hidapi/pull/194
* The usage_page and usage was always `0` during my tests on Ubuntu

Once this is merged and works as expected, I'll extend the dev documentation.

### Pull Request Checklist
- [x] hidapi added to macOS build
- [x] hidapi added to linux build
- [x] hidapi added to windows build
- [ ] Changes have tests

### DIFF
--- a/dev/build_hidapi.sh
+++ b/dev/build_hidapi.sh
@@ -15,7 +15,7 @@ set -euo pipefail
 . ./plover_build_utils/deps.sh
 . ./plover_build_utils/functions.sh
 
-python='python3'
+python='python'
 
 MACOS_UNIVERSAL2="${MACOS_UNIVERSAL2:-1}"
 
--- a/dev/write_hidapi_pth.py
+++ b/dev/write_hidapi_pth.py
@@ -32,10 +32,11 @@ def main(argv: list[str]) -> int:
     pth_path = site_packages / "plover_hidapi_add_dll_dir.pth"
 
     # Keep it one line: .pth executes arbitrary Python on import.
+    # Include defensive checks for directory existence and DLL loading
     code = (
         "import os,sys;"
         "p=r%r;"
-        "getattr(os,'add_dll_directory',lambda *_:None)(p) if sys.platform=='win32' else None\n"
+        "os.add_dll_directory(p) if sys.platform=='win32' and hasattr(os,'add_dll_directory') and os.path.isdir(p) else None\n"
         % (str(dll_dir),)
     )
 
--- a/plover_build_utils/functions.sh
+++ b/plover_build_utils/functions.sh
@@ -243,7 +243,8 @@ fetch_hidapi() {
   local src_dir="$1" download_dir="$2"
   local file_name="hidapi-${hidapi_version}.tar.gz"
   local url="https://github.com/libusb/hidapi/archive/refs/tags/${file_name}"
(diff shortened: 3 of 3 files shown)


plover/machine/procat.py
```diff
@@ -22,7 +22,7 @@
 BYTES_PER_STROKE = 4
 
 
-class ProCAT(SerialStenotypeBase):
+class ProCat(SerialStenotypeBase):
```
dnaq: seems unrelated to this change, but nbd

plover/system/english_stenotype.py
```diff
@@ -238,6 +238,59 @@
         "-Z": "Z",
         "no-op": ("!", "^", "+"),
     },
+    "Plover HID": {
```
dnaq: The configuration gui popped up the keys in a bit of a weird order, unsure if it's related to this or something else

linux/build_hidapi.sh
```diff
@@ -0,0 +1,21 @@
+need() { command -v "$1" >/dev/null 2>&1 || { echo "Error: '$1' not found" >&2; exit 2; }; }
+
+fetch_hidapi_linux() {
```
dnaq: This is a genric comment on all the fetchers, no matter what platform. I believe it would be good to do two things when fetching deps as part of a build.

1. Pin the hashes of the fetches as well. This would help in those rare cases were a dep gets retagged without you knowing.
2. Add an update script (that can be run manually, or in ci) to automatically test the build with later version of those deps, and update the hashes accordingly. This helps out with supporting new versions. Bonus points for making it integrate with dependabot.

dnaq (commented):
Some small comments, but all in all, great work.

Also tested with a modified version of the plover-flake to pull in this branch of plover, and hid seemed to work well for me.

dnaq (approved):

plover/machine/procat.py
mkrnr: That was done to align the machine class names

plover/system/english_stenotype.py
mkrnr: They currently are shown row-by-row, and not in steno order, right? It's the same with Gemini PR.
I assume it's using the `KEYS_LAYOUT` order from `plover_hid` but I'd leave it like this for now.

linux/build_hidapi.sh
mkrnr: Great idea to pin the hash! Added that and used the existing `plover_build_utils.download` to validate the hash.

While doing so, I noticed that all the `fetch_hidapi_*` methods are identical across all operating systems so I moved the method to a central place.

The update script sounds nice but for me it's enough if that `plover_build_utils.download` fails after a version bump. Also the dependabot integration will have to wait.

Issue #1726, opened 2025-05-13
Uinput output for other layouts than US


LilleAila:
**Describe the bug**

Using the modified qwerty layout from #1721 will not work with non-US layouts, as many symbols such as "-", "/", "?" etc. will not be at the same locations. For example, in norwegian the / is at shift+7 rather than at a dedicated key as in US.

The reason i did not include any special characters at all in #1679, was because it should be possible to use any layout. I suggest having separate layout options for generic (minimal) qwerty, and another for US qwerty. I also think the characters should be not in `BASE_LAYOUT` but in specifically `qwerty-us`, as `BASE_LAYOUT` should be the keys which are the same regardless of layout.

Relevant commit: https://github.com/openstenoproject/plover/pull/1679/files/18aaf5174a0feaa5b4e3fea2fbce72bcc1d9f561..5279766d11ad6405eb08e168a01c438f25d103de

If we would want to have all possible keyboards emulated with direct keys, we would have to implement something similar to the code that was removed, but the xkbcommon dependency caused some problems as it requires a system-wide dependency and i experienced version conflicts.

---

Also, i saw some confusion about this line: https://github.com/openstenoproject/plover/pull/1721/files#diff-6b4281e6a4006cf3e22f4cfbcaf9fcc8ff7337cf721c33d5051f1e6cbd369844R279

The reason it was done like that, is because that is used by the keyboard capture, which does not care about what the actual layout is, but which keys were pressed.

**To Reproduce**

1. Use another layout than US (for example norwegian)
2. Type some of the characters added in #1721 

**Expected behavior**

The correct character is output

**Operating system**

Linux / Wayland on latest main branch

CC @user202729 @mkrnr

mkrnr:
Thanks a lot for raising this, makes a lot of sense to me!

Regarding `xkbcommon` causing problems: it might be worth giving it another shot after #1720 is merged. Implementing the mapping for all the different keyboard layouts sounds painful.

user202729:
I thought the library is used to magically determine the current layout in use. If it's just for pre-generate the mapping can't it be hardcoded?

LilleAila:
> I thought the library is used to magically determine the current layout in use. If it's just for pre-generate the mapping can't it be hardcoded?

Do you mean to generate the full mapping first and save it to a file? That would be possible, but i’m not sure if such a big file of keyboard layouts should be included in plover.

user202729:
Isn't it included in the source code of the X server anyway?

[...]

PR #1807, opened 2026-01-07
Add determining keyboard layout from Wayland

iansw246:
## Summary of changes

Adds a Linux keyboard layout option "wayland-auto" that retrieves the current keyboard layout from the running Wayland compositor. Uses the xkbcommon package to parse the keymap, as well as to generate the other keyboard layout options.

Solves #1726 but only for Wayland users.

Another limitation is that keyboard capture isn't using the retrieved layout. This means keyboard modifiers other than the standard QWERTY keys (for example mapping CapsLock to Ctrl) won't be handled properly when passing through keyboard shortcuts. This issue exists currently so new issues shouldn't be added and this PR should be ready to merge.

### Pull Request Checklist
- [ ] ~~Changes have tests~~ Linux specific changes 

### DIFF
--- a/plover/gui_qt/config_window.py
+++ b/plover/gui_qt/config_window.py
@@ -306,9 +306,23 @@ class BooleanAsDualChoiceOption(ChoiceOption):
         super().__init__(choices)
 
 
+class TextWrapQLabel(QLabel):
+    """Simple QLabel wrapper that enables wordWrap"""
+
+    def __init__(self, *args):
+        super().__init__(*args)
+        self.setWordWrap(True)
+
+
 class ConfigOption:
     def __init__(
-        self, display_name, option_name, widget_class, help_text="", dependents=()
+        self,
+        display_name,
+        option_name,
+        widget_class,
+        help_text="",
+        dependents=(),
+        additional_widget_classes=[],
     ):
         self.display_name = display_name
         self.option_name = option_name
@@ -318,6 +332,9 @@ class ConfigOption:
         self.layout = None
         self.widget = None
         self.label = None
+        # Other widgets to be added immediately after the main widget_class
+        # Does not work in dependents
+        self.additional_widget_classes = additional_widget_classes
 
 
 class ConfigWindow(QDialog, Ui_ConfigWindow, WindowStateMixin):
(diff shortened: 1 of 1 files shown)


mkrnr:
Thank you so much @iansw246 for this important contribution!

I tested it on my Ubuntu 24.04 machine having a German and English layout. For me, Plover always uses the German layout when selecting `wayland-auto`. I'm not deep into this but maybe its because I always get "0" here:

```
martin@martin-ubuntu-X230:~$ gsettings get org.gnome.desktop.input-sources sources
[('xkb', 'de'), ('xkb', 'us')]
martin@martin-ubuntu-X230:~$ gsettings get org.gnome.desktop.input-sources current
uint32 0
```
Even when setting my keyboard to `en` in the Ubuntu taskbar, I still get that 0. Maybe because of that, Wayland also doesn't get the info that I switched to `en`?

Apart from that problem, the detected German layout does work.

One question: Have you tested how Plover reacts when the `libxkbcommon` package is missing? I'm thinking of people with very minimal Linux installations that don't use Wayland.

Given my limited knowledge in this area, I'm not able to a thorough review.  @LilleAila do you possibly have time to look into this approach?

LilleAila:
LGTM from a glance at the code. I think it is olah that this only applies to wayland as that is what the uinput output was primaily intended from. I have switched to MacOS since initially writing the output layer, but i can see if i can get it working on my secondary computer. Other than that it looks good (apart from the issue @mkrnr mentioned) and it seems like it is ready to merge. Although i haven’t tested it personally, i didn’t see anything specific in the code that i disagree with.

mkrnr:
Thanks a lot @LilleAila!

@iansw246 there's now some minor conflicts since #1812 also modified the `apt_get_install` in the CI files. Once those are resolved, I'm happy to approve and merge.

mkrnr:
@iansw246 let me know when the PR is ready to be merged

iansw246:
@mkrnr About the multiple layouts, there's not a way to detect which layout is active over Wayland without receiving the key events themselves. Plover always assumes the first layout is active. I'll add a warning in the UI before merging. Is there somewhere else this be documented too?

Plover will crash without libxkcommon if it tries to use Uinput. I'll work on fixing that later. If you think supporting that use case is important, we can wait until that's fixed as well.

mkrnr:
@iansw246 thanks! Maybe you find a place in the docs regarding selecting the first layout but a note in the UI is already very good. I'd suggest not relying on tooltips though since not everyone reads them.

You can decide if you want to fix the crash when libxkcommon is missing now or later. I can't really judge how common that situation would be.

mkrnr (approved):

iansw246:
Added text below the keyboard layout selection. Should be good to merge now.

I will fix the libxkbcommon crash another time.

mkrnr (approved):
Thanks for adding the info text! Tested it on Ubuntu and it works as expected.

PR #1808, opened 2026-01-10
Make Plover HID not fail on startup

user202729:
## Summary of changes

From the [documentation of `hid`](https://pypi.org/project/hid/) (typo not mine):

> pyhidapi is dependant upon the [hidapi library](https://github.com/libusb/hidapi), which must be installed separately.

This may not be [installed](https://github.com/libusb/hidapi?tab=readme-ov-file#installing-hidapi), which means the import may raise `ImportError`.

For me, if that library is not installed, `import hid` fails with

```
ImportError: Unable to load any of the following libraries:libhidapi-hidraw.so libhidapi-hidraw.so.0 libhidapi-libusb.so libhidapi-libusb.so.0 libhidapi-iohidmanager.so libhidapi-iohidmanager.so.0 lib hidapi.dylib hidapi.dll libhidapi-0.dll
```

at startup. This change makes it so that it only fail when the user tries to select `Plover HID` as the input source.

Note: we don't actually need to depend on `hid`, that can be left as an optional dependency.

The `from __future__ import annotations` is just in case, because there are several type annotations that use `hid.something`, and on sufficiently-old Python versions they will actually be evaluated and raise errors (depends on what is the minimum-supported version this may not be necessary)

Note: only tested on Linux.

Note: I consider the case where the first `import hid` fails but the second succeed _extremely_ unlikely, but it may look cleaner to capture the error message in the first import and `log.error()` it when started.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1808.core.md
+++ b/news.d/bugfix/1808.core.md
@@ -1,0 +1,1 @@
+Avoid error at startup if `hidapi` library is not available.
--- a/plover/machine/plover_hid.py
+++ b/plover/machine/plover_hid.py
@@ -7,12 +7,12 @@ https://github.com/dnaq/plover-machine-hid
 Plover HID is a simple HID-based protocol that sends the current state
 of the steno machine every time that state changes.
 """
+from __future__ import annotations
 
 from plover.machine.base import ThreadedStenotypeBase
 from plover.misc import boolean
 from plover import log
 
-import hid
 import time
 import platform
 import ctypes
@@ -21,6 +21,10 @@ from queue import Queue, Empty
 from typing import Dict
 from dataclasses import dataclass
 
+try:
+    import hid
+except ImportError:
+    hid = None
 
 def _darwin_disable_exclusive_open():
     lib = getattr(hid, "hidapi", None)
@@ -36,7 +40,7 @@ def _darwin_disable_exclusive_open():
 
 
 # Invoke once at import time (before any devices are opened)
-if platform.system() == "Darwin":
+if platform.system() == "Darwin" and hid is not None:
     _darwin_disable_exclusive_open()
 
 USAGE_PAGE: int = 0xFF50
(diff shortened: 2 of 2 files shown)


mkrnr:
Thank you @user202729!

The current way of installing `hidapi` is indeed a pain and can result in the `hid` package missing.

I noticed, however, that [hidapi 0.15.0](https://pypi.org/project/hidapi/) was recently released on PyPI. I did some tests and this version does fix the issue with macOS (`hid_darwin_set_open_exclusive` not being available) so I will try to migrate to that version next days. That would simplify our setup a lot and result in hid always being there.

If it's fine for you, let's park your PR in case I'm not successful with moving to that `hidapi` package.

user202729:
https://github.com/opensteno/plover/pull/1812 fixes the issue on the machine I tested. Caveat is that the user **must** make sure that `hid` package is not also installed, otherwise the error remains. (probably plover can help with this by declaring conflicting package or something.)

That said, I don't understand how the linked pull request is supposed to fix the issue raised in this pull request. Quoting documentation of `hidapi`:

> ## [Software Dependencies](https://pypi.org/project/hidapi/#toc-entry-2)
> 
> [...]
> 
> * hidraw or libusb+libudev on Linux

if the last entry is not installed, we would still get some sort of error?

I notice the dependency installs a `hidraw.cpython-310-x86_64-linux-gnu.so` file however.

mkrnr:
Thanks @user202729 for testing with #1812. I assume you only checked if Plover works but not the Plover HID functionality itself? I'm asking because I'm having trouble getting the Plover HID keyboard to work on Linux. Windows and maCOS work perfectly.

user202729:
Indeed, I don't need the functionality (currently I use the Keyboard machine with some local modifications to only activate on one keyboard but not another. I may upstream this feature eventually, but it's somewhat nontrivial…), only check that the error doesn't show up.

PR #1809, opened 2026-01-10
Move plover.machine.keymap documentation to docstring

user202729:
**Note**: this needs some discussion!

The problem: do we want to have documentation in a separate `.md` file (i.e. the status quo), or do we want it in the docstrings?

In both cases, the rendered documentation looks exactly the same (*). The considerations are:

- Developers preference. I prefer docstrings because
    - Documentation ought to be close to the source, otherwise it is easy to make code changes and forget to update the documentation.
    - When I use `ctags`, I go to the function definition directly, it would be extra work to lookup the documentation  at a separate place.
- Dependency. Doing this requires `sphinx-autodoc`.
- If we do this, we need to either
    - use reStructuredText instead of Markdown (what this pull request does),  or
    - Implement a solution in https://stackoverflow.com/questions/56062402/force-sphinx-to-interpret-markdown-in-python-docstrings-instead-of-restructuredt.

(*): there are some differences.

- One is `KeysView` (but reading the source code, I believe it is indeed returning a `KeysView` instead of a `list` i.e. the work in this PR already uncovers a small inaccuracy in the documentation)
- the other is ordering of members changed slightly compared to previous version, but this is trivial to fix by reordering them in the source code.

(**): After the change to Google style, some of the documentation formatting are changed slightly. Probably for the better.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -18,6 +18,7 @@ extensions = [
     "sphinx_plover",
     "myst_parser",
     "sphinx.ext.autodoc",
+    "sphinx.ext.napoleon",
     "sphinx.ext.todo",
 ]
 
--- a/plover/machine/keymap.py
+++ b/plover/machine/keymap.py
@@ -22,24 +22,18 @@ from plover import log
 
 class Keymap:
     """
-    .. attribute:: keys
-       :type: List[str]
+    Args:
+        keys: A list of possible keys on the physical machine. For a serial steno
+            protocol like TX Bolt, this would be the steno keys themselves
+            (``S-``, ``T-``, etc.); for the keyboard, this would be keyboard keys
+            (``q``, ``w``, etc.).
+        actions: A list of possible actions in the steno system.
 
-       A list of possible keys on the physical machine. For a serial steno
-       protocol like TX Bolt, this would be the steno keys themselves
-       (``S-``, ``T-``, etc.); for the keyboard, this would be keyboard keys
-       (``q``, ``w``, etc.).
-
-    .. attribute:: actions
-       :type: List[str]
-
-       A list of possible actions in the steno system.
-
-       In addition to the steno keys in the current system (see
-       :attr:`system.KEYS<plover.system.KEYS>` for more information),
-       the actions may include ``no-op``, a special action that does nothing,
-       and ``arpeggiate``, a special action for arpeggiate mode that is only
-       available if the current machine is a keyboard.
+            In addition to the steno keys in the current system (see
+            :attr:`system.KEYS<plover.system.KEYS>` for more information),
+            the actions may include ``no-op``, a special action that does nothing,
+            and ``arpeggiate``, a special action for arpeggiate mode that is only
+            available if the current machine is a keyboard.
     """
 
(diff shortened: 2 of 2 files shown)


mkrnr:
I really like this approach! This makes it way more likely that the documentation stays in sync with the code.

I've seen [sphinx-autodoc2](https://sphinx-autodoc2.readthedocs.io/en/latest/) which would support markdown but seems like it's not maintained.

We could use the [napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#module-sphinx.ext.napoleon) extension to go for NumPy or Google style docstrings?

user202729:
Okay good.

Do you have any preference with respect to the 2 points you raised?

I don't think argument formatting matters too much, especially because with type annotations and variable names, almost all of them should be self-documenting. But can't hurt to have either, especially since napoleon is built-in in sphinx so no additional dependency needed. (Google style looks cleaner than Numpy style to me.)

If no changes needed, we can start with merging this first, then move the remaining documentation later in subsequent pull requests.

mkrnr:
I'd also go for Google style for the same reason.

Would it be much work to already add napoleon with this PR and reformat the keymap file to the Google style? Then we don't have to touch it a second time.

user202729:
Done.

Caveat: in some functions such as `get_keys()`, the docstring is "Return (something)." Instead of "Return:(line break)something." In this specific case I think there's no point using the latter, and there's especially no point having both.

I also change `keys` from attr to args, which necessitate a change to the reference.

mkrnr (approved):
Great work, thanks a lot! Looking forward to expanding this to other files.

As always, feel free to merge when ready.

user202729:
I'm looking at a few other files. One issue I have with napoleon is that it renders as the following. (Open issue: https://github.com/sphinx-doc/sphinx/issues/7582)

bad:

image

instead of

good:

image

this is because of the generated RST:

```
# good:
.. attribute:: attach
   :type: bool

   Whether to delete the space before the translation when the
   initial stroke is translated.

# bad:
.. attribute:: capitalize

   Whether to capitalize the translation when the initial
   stroke is translated.

   :type: bool
```

What's your opinion on how to deal with this?

- use raw RST instead of Napoleon for attributes (looks slightly uglier in source code, but actually I don't mind, the difference isn't large enough)
- fix it upstream and wait for a few months (most reliable but may takes forever)
- use some postprocessing hack in `doc/conf.py`, or monkey-patch the relevant methods (might be fragile with respect to Sphinx version upgrade)
- leave it as is and make the generated doc looks slightly uglier
- something else?

mkrnr:
I think it would be great to enforce Google style doc strings at some point, e.g. via `pydocstyle` in `Ruff`.

So I wouldn't start adding exceptions like using raw RST because Ruff would start complaining about those.

In this concrete case, I'd accept the slightly uglier doc for now. If not too much effort, also providing a fix upstream would be amazing of course.

What do you think?

user202729:
That's fine for me. https://github.com/opensteno/plover/pull/1814 updated accordingly.

PR #1814, opened 2026-01-18
Move more documentation to docstring

user202729:
## Summary of changes

Follow-up on https://github.com/opensteno/plover/pull/1809

One issue is discussed in the linked pull request above. The conclusion is to stick to Google docstring style, even though the rendered documentation looks slightly worse

image

another issue is that there's a section (all configuration parameters), should it be

- left as Markdown (which is actually possible with some `{eval-rst}`
- convert to rst (current option)
- move to Python file as class docstring

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/doc/api/engine.rst
+++ b/doc/api/engine.rst
@@ -6,61 +6,61 @@
 
 .. autoclass:: plover.engine.StenoEngine
 
-   .. automethod:: plover.engine.StenoEngine._in_engine_thread
+   .. automethod:: _in_engine_thread
 
-   .. automethod:: plover.engine.StenoEngine.start
+   .. automethod:: start
 
-   .. automethod:: plover.engine.StenoEngine.quit
+   .. automethod:: quit
 
-   .. automethod:: plover.engine.StenoEngine.restart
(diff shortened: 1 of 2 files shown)


user202729:
there's also the method of `napoleon_use_ivar = True` which leads to slightly different rendering. Not sure what's the preference here. (suggested in https://github.com/sphinx-doc/sphinx/issues/7582)

there are https://github.com/sphinx-doc/sphinx/pull/7583 which already make the change, but it causes https://github.com/sphinx-doc/sphinx/issues/7808 , so it gets reverted.

I guess if `napoleon_use_ivar = True` looks good enough then we can go with that.

user202729:
plover.engine moved. This one suffers particularly badly from the attribute issue.

`napoleon_use_ivar = True` would have looked like this.

image

Not sure what to feel about this.

------

Also, for some of the `namedtuple`s, explicitly assigning the docstring to `__doc__` is unnecessary (you can just write it as a docstring right after the variable), but apparently not assigning will make the default value of the fields disappear. No idea why.

PR #1815, opened 2026-01-19
Add selection list for Keyboard selection when on Linux in uinput

### LINKED ISSUE #1767: Logitech MX Master mouse is stuck when upgrading Plover from 4 to 5
**Describe the bug**

When starting Plover using machine keyboard the mouse input freezes and cannot be used (no clicking, scrolling, pointing) unless the keyboard is disconnected physically. Regardless of whether the output in Plover is disabled or enabled.

This happens on Hyprland using the latest 5.0.0 appimage, the rc1 and dev appimage are both also effected. Connecting the mx master mouse via bluetooth or dongle both freeze. Connecting a USB dumb mouse works fine.

**To Reproduce**

1. Run plover through command-line.
2. Attempt to move mouse and observe it is stuck.
3. Unplug keyboard
4. Mouse is no longer stuck, plover should error saying keyboard capture error, no such device.

**Expected behavior**

Mouse should not be stuck.

**Screenshots**
N/A

**Operating system**

 - OS: Linux
 - Plover Version 5.0.0

**Hardware**

Standard USB keyboard, logitech MX master mouse via dongle and bluetooth.

Edit: This issue was recreated on both Hyprland (wayland) and i3 (X11).

NijeboerFrank:
## Summary of changes

Add a Regex field which can be used to filter some keyboards. This prevents e.g. the situation where a mouse with a Logitech Unifying receiver is incorrectly captured on Linux in Wayland (that was bothering me). 

I currently don't have access to other operating systems, so this PR only works for the uinput keyboard capture.

 I will try to add some tests later. I'm also not very familiar with qt, so all feedback is appreciated.

Closes #1767 

### Pull Request Checklist
- [x] Changes have tests


NijeboerFrank:
In my case, I would add `^(?!Logitech)` to the Keyboard Regex field to exclude the unifying receiver with my mouse from being captured by Plover.

mkrnr:
Thanks a lot @NijeboerFrank for working on this!

I now approved the GitHub actions run. Some tests will need the new parameter to be included.

Regarding the GUI: It would be good to make it more clear that this option is Linux only. Adding this to the tooltip is probably not enough because most people don't read that. You could either add "Linux only" to the option name or, even better, only show this option on Linux. See `config_window.py`:
```
        if PLATFORM != "linux":
            # Keyboard layout is only relevant on Linux
            keyboard_layout_option = option_by_name.get("keyboard_layout")
            if keyboard_layout_option is not None:
                keyboard_layout_option.label.hide()
                keyboard_layout_option.widget.hide()
```

NijeboerFrank:
Thanks for the tips @mkrnr! I'm trying to find some time this week to fix the tests and improve the GUI.

user202729:
I've a more fancy UI suggestion in https://github.com/opensteno/plover/issues/1187#issuecomment-1908992553 (in my case it was for X implementation), but haven't found the time to properly integrate and upstream it. Having just a regex field works too (kind of).

only for Linux uinput though. Having it silently do nothing might not be such a good idea.

For the specific issue https://github.com/opensteno/plover/issues/1767:

In principle, uinput shouldn't be used when X is available, but maybe depends on certain environment variables it may not be correctly selected. Relying on the user to manually provide a regex is certainly a hack instead of a proper fix.

More proper fix would be

- Distinguish between keyboard and mouse, and don't capture mice (mouses?). I think it's possible for a device to show up as both a keyboard and a mouse, maybe this is the case here.
- More reliable way to see if X works.


------

@mkrnr actually I don't think hiding the option is a particularly good idea, might be better to gray it out (with a tooltip explaining why it's disabled on hover). Otherwise there may be some users who check a screenshot online, see the feature, not seeing it in their local install, and get confused.

NijeboerFrank:
@user202729 I was also thinking of something like that as well today. I could try to implement something like that to properly fix it, because I also think that the Regex version might be too hacky for non-technical users who run into this issue. I'm thinking of something like a dropdown _if_ I can get it working.

My initial reasoning for Regex was that if you leave the field empty it will just work as it did previously and it won't interfere with the current way that uinput works for the people that don't have similar issues as the one that I had.

NijeboerFrank:
I changed the implementation to a list selection. Currently you can only select 1 keyboard from the list, because I also don't really see the usecase for selecting multiple as @user202729 mentioned in [#118](https://github.com/opensteno/plover/issues/1187#issuecomment-1908992553). 

So now if you have the "Keyboard" option selected, you can select which Keyboard will be captured with uinput.

Let me know what you think, I'm happy to implement any feedback when I have the time.

user202729:
Actually, after thinking about it, a use case is if if you have multiple steno keyboards and may use one of them.

Anyway, I think the current implementation breaks the oslayer abstraction layer. Not to mention https://github.com/opensteno/plover/issues/1742 . And in X, you "need" the keyboard capture object to get the list of keyboards.

(up to @mkrnr what to do with it. On one hand there's the above, on the other hand there's the "the perfect is the enemy of the good" saying.)

NijeboerFrank:
@user202729 do you mean that you think that the implementation in this PR breaks the `oslayer` abstraction layer? Because I think I only changed the `keyboardcontrol_uinput.py` `KeyboardCapture` class. I didn't change anything in or related to `keyboardcontrol_x11` AFAIK. I also added if-statements around the parts so that this code is only carried out on Wayland and not X11. 

So if there's anything in this PR that needs fixing, let me know. 

In the future I can also try to get multi-selection working so that people with multiple steno keyboards can use both and if there's need for such a feature.

mkrnr:
@NijeboerFrank thanks again for your work on this!

I tested this on my Ubuntu laptop (wayland) with an external keyboard. Not sure why but non of the keyboards show up in the dropdown menu. Might be something about my setup though so I trust you if you tested it.

We can go ahead and merge this after you resolved the conflict in the `tox.ini`.

VasilisThePikachu:
image
I tested this PR myself and resolved the conflict because the author does not seem active on github anymore.

I can confirm it works on Arch Linux with kernel 7.0.0-rc7-3-cachyos-rc and KDE plasma 6.6.4 running on wayland (Although i did just notice it is picking up my headphones as a keyboard? hmmm)

I tried making a git patch for the conflict, but I am unsure if this will apply correctly so double check it, otherwise this is a very simple conflict it took me 2 mins to resolve it.

If you are unable to apply a patch for @NijeboerFrank behaf please either directly pull their changes into the branch or close this PR and let someone else reopen it.

[conflict.patch](https://github.com/user-attachments/files/27030814/conflict.patch)

mkrnr (dismissed):
Thanks @VasilisThePikachu for testing and for bumping this PR. I resolved the conflicts and updated the translation files.

mkrnr (approved):

Issue #1554, opened 2022-08-22
With space placement set to 'after output', trailing '/' characters blocks OK button in add translation window


Anodynous:
**Describe the bug**

With space placement set to "after output" a trailing '/' character is appended to the end of the outline, causing the OK-button to be grayed out. Manually deleting it enables the OK button making it possible to add the new translation. With spaces added before output the issue is not present.

**To Reproduce**

1. Set space placement to "after output" in configuration
2. Open the add translations GUI
3. Add stroke(s)
4. Add translation
5. Remove trailing '/' character from strokes outline in order to enable the OK-button 

**Expected behavior**

Translation is possible to add regardless of space placement without the need to remove the trailing stroke-separator character from the outline.

**Screenshots**

Trailing '/' character prevents adding translation:
![image](https://user-images.githubusercontent.com/17461433/185953745-fc96a75b-9bed-47c7-b6d2-d2f6ed5c21ef.png)

Deleting the last '/' character of the outline enables addition of translation:
![image](https://user-images.githubusercontent.com/17461433/185953788-0b338dd2-b8c1-48f6-bcd1-158fe6d248bc.png)

**Operating system**

 - OS: Windows 10
 - Plover Version: v4.0.0.dev12

greenwyrt:
Closed by #1836

PR #1836, opened 2026-03-28
fixes for add translation dialog

greenwyrt:
## Summary of changes

This mitigates the trailing backslash in add translations dialog (#1554) when space placement is set to `After Output` by temporarily setting it to `Before Output`. 

Also a minor change to amend #1833. In the process of making translation dialog tab-navigable, the translation input had become un-clickable.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1836.ui.md
+++ b/news.d/bugfix/1836.ui.md
@@ -1,1 +1,1 @@
-Fixes trailing backslash when space placement is set to `After Output` in the add translations dialog.
+Fix trailing slash when space placement is set to `After Output` in the Add Translation dialog.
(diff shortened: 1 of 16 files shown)


mkrnr:
@greenwyrt I think it's better to not modify the engine state from inside the Add Translation dialog since it would change the global engine state which could lead to unexpected behavior. I pushed a commit that extends StartingStrokeState (and also adds constants for `Before Output` and `After Output`). Let me know your thoughts on this alternative approach.

greenwyrt:
Oh that is so much better!

PR #1837, opened 2026-03-29
correct encoding of generated *_ui.py files on windows system

greenwyrt:
## Summary of changes

tox i18n test was failing on windows because generated *_ui.py files from pyside6-uic were not utf-8.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1836.ui.md
+++ b/news.d/bugfix/1836.ui.md
@@ -1,0 +1,1 @@
+Fix trailing slash when space placement is set to `After Output` in the Add Translation dialog.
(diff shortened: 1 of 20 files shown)


mkrnr:
I tested PySide6 6.11.0 on macOS and got a lot of the log entries:
```
_pythonToCppCopy: Cannot copy-convert 0x1118bfd80 (set) to C++.
```

Let's bump the PySide6 version in a separate PR since it requires some more testing and possibly fixing.

mkrnr (approved):

greenwyrt:
I get the same entries about cannot copy-convert but with ` Cannot copy-convert 00000263E5E60E40`. 

Not sure what the impact is since keyboard and translation seem to work properly.

PR #1838, opened 2026-04-04
Hook up modifier aliases for wayland output

olivia-fl:
Without this, aliases like "control" don't work.

## Summary of changes

Without this, aliases like "control" don't work, resulting in an error message like: "Key control is not valid!".

### Pull Request Checklist
- [ ] Changes have tests


mkrnr (approved):
Thanks a lot for your contribution! 🙂

Issue #1629, opened 2023-09-19
[Bug] When an unconfigured key is used, the output stops working, even on the paper tape and won't work again until reconnect


Monniasza:
**Describe the bug**

**To Reproduce**

1. Configure a key to be empty (not no-op)
2. Use that key
3. Write something else on that machine
4. See that nothing is written

**Expected behavior**
Nothing happens, and machine works without interruption

**Screenshots**

**Operating system**

 - OS: Windows 10
 - Plover Version 4.0.0dev12

**Hardware**
I'm using Picosteno, but I've forgot to map S1-, so the issue occurred when I wanted to write some numbers

mkrnr:
Fixed with https://github.com/opensteno/plover/pull/1855

PR #1855, opened 2026-07-21
Treat unbound keys as no-op

### LINKED ISSUE #1850: `KeyError` when pressing a physical key is pressed that isn't in the system keymap.
**Describe the bug**

If you have a system where a certain physical key is not present in the system `KEYMAPS`, then pressing that key throws a `KeyError` and causes the device to disconnect (tested only with Gemini PR, but may apply to other devices too).

**To Reproduce**

1. Create a system with a keymap like the following:
```py
KEYMAPS = { "Gemini PR": { } }
# NOTE: this one maps no keys, however as long as there's at least one unmapped key, this will work.
```
2. Press an key that is not mapped in `KEYMAPS`.
3. Device will disconnect, throwing a stack trace like the following:
```
Exception in thread Thread-2-machine:
  File "/home/runner/work/plover/plover/build/appimage/plover.AppDir/usr/lib/python3.13/threading.py", line 1041, in _bootstrap_inner
  File "/home/runner/work/plover/plover/build/appimage/plover.AppDir/usr/lib/python3.13/site-packages/plover/machine/gemini_pr.py", line 56, in run
  File "/home/runner/work/plover/plover/build/appimage/plover.AppDir/usr/lib/python3.13/site-packages/plover/machine/base.py", line 94, in _notify_keys
  File "/home/runner/work/plover/plover/build/appimage/plover.AppDir/usr/lib/python3.13/site-packages/plover/machine/keymap.py", line 164, in keys_to_actions
KeyError: '#1'
```

**Expected behavior**

Most users or system developers probably don't expect a device to disconnect because they pressed an unmapped key. This can be an error in the system, however the [documentation here](https://plover.readthedocs.io/en/latest/api/system.html#plover.system.KEYMAPS) doesn't explicitly say you have to map all keys, and use `no-op` for keys that aren't used by the system. It only says you _can_ use `no-op`.

There's multiple solutions here:
1. Automatically map all unmapped to `no-op`.
2. State in the documentation that all keys must be mapped. Ideally there should be a run-time check when loading a system that refuses to load it if this invariant isn't upheld.

I prefer solution (1), as it makes it less annoying (and if, for whatever reason, a newer version of a protocol gains more keys, then you don't have to go modify every custom keymap).

**Operating system**

 - OS: Fedora 42
 - Plover Version: 5.3.0

I'm sure this is replicable on any system though.

toyboot4e:
## Summary of changes

Treat unbound keys as no-op. It's _slightly_ kinder to machine developers.

Closes #1850, #1629 

- If the user machine emits a key out of range, it would still fail like `AssertionError: 'k9' not in OrderedDict({...})`. Maybe that's ok for now.

### Pull Request Checklist
- [x] Changes have tests

### DIFF
--- a/news.d/bugfix/1850.dict
+++ b/news.d/bugfix/1850.dict
@@ -1,1 +1,0 @@
-Treat unbound keys as no-op
--- a/news.d/bugfix/1850.dict
+++ b/news.d/bugfix/1850.dict
@@ -1,0 +1,1 @@
+Treat unbound keys as no-op
--- a/news.d/bugfix/1850.core.md
+++ b/news.d/bugfix/1850.core.md
@@ -1,0 +1,1 @@
+Treat unbound keys as no-op.
--- a/plover/machine/keymap.py
+++ b/plover/machine/keymap.py
@@ -93,7 +93,7 @@ class Keymap:
         action_list = []
         for key in key_list:
             assert key in self._keys, "'%s' not in %s" % (key, self._keys)
-            action = self._bindings[key]
+            action = self._bindings.get(key, "no-op")
             if "no-op" != action:
                 action_list.append(action)
         return action_list
--- a/test/test_keymap.py
+++ b/test/test_keymap.py
@@ -58,6 +58,18 @@ def test_keymap_set_mappings():
     assert k.get_mappings() == MAPPINGS_FULL
 
 
+def test_keymap_keys_to_actions_for_unbound_key():
+    # An unbound key is treated as no-op.
+    k = new_keymap()
+    k.set_bindings(BINDINGS_DICT)
+    assert "k2" not in k.get_bindings()
+    assert k.keys_to_actions(["k0", "k2", "k4"]) == ["a0", "a1"]
+    # If the machine emits a key out of range, it's an error:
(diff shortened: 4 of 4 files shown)


news.d/bugfix/1850.dict
```diff
@@ -0,0 +1 @@
+Treat unbound keys as no-op
```
mkrnr: Some trivial points: 
* the file is missing the `.md` ending
* I'd rather put it in `core` instead of `dict` since it's not modifying the dictionary 
* The text should end with a period

mkrnr (changes requested):
Thanks a lot! I like the approach and it's already a good improvement.

Just some small remarks regarding the news file.

news.d/bugfix/1850.dict
toyboot4e: Thank you.. I've applied the three points in https://github.com/opensteno/plover/pull/1855/changes/0104f1028cb3a88aaaaef52112a2735cd55447e3.

mkrnr (approved):

toyboot4e:
Sorry, it closes #1629, too. It's still open due to my PR comment mistake.

PR #1857, opened 2026-07-26
fix(rtfcre): escape characters outside code page 1252 on export

### LINKED ISSUE #1705: RTF dictionary fails with encoding error
**Describe the bug**

The export of a dictionary to the RTF format can fail with an encoding error.

First reported by @jthlim on discord.

**To Reproduce**

1. Right-click on a larger dictionary like main.json
2. Click `Save dictionaries as...`
3. Click `Create a copy of each dictionary`
4. In the last drop-down, select `RTF dictionaries`
5. An encoding error occurs: `Exception: creating dictionary /Users/martin/Downloads/main - Copy.rtf failed. 'charmap' codec can't encode character '\u0301' in position 24: character maps to <undefined>`

**Expected behavior**

The dictionary is exported.

**Screenshots**

**Operating system**

 - OS: macOS 15.3.1
 - Plover Version 4.0.0

Sanjays2402:
## Summary of changes

Saving a dictionary as RTF encodes the file as code page 1252, so any translation containing a character outside that code page (e.g. `ở` in `phở`, or a combining acute accent) raised `UnicodeEncodeError` and the export failed.

Characters that cp1252 cannot represent are now escaped with the RTF `\uN` control word using UTF-16 code units, so characters outside the BMP are written as a surrogate pair — the approach @sammdot described on the issue (surrogate pairs, positive numbers, wrapped in `\uc0` so no ANSI fallback is expected). The parser learned `\uN`/`\ucN` to match, so RTF dictionaries still round-trip.

Closes #1705

### Pull Request Checklist
- [x] Changes have tests


plover/dictionary/rtfcre_parse.py
```diff
@@ -241,6 +243,24 @@ def parse_rtfcre(text, normalize=lambda s: s, skip_errors=True):
...
+                    # Low surrogate: combine with the preceding high one.
+                    high = ord(g_text[-1]) - 0xD800
+                    g_text = g_text[:-1] + chr(
+                        0x10000 + (high << 10) + (code_unit - 0xDC00)
+                    )
+                else:
+                    g_text += chr(code_unit)
+            # Number of fallback characters following a unicode escape.
+            elif uc_rx.fullmatch(ctrl):
+                pass
```
mkrnr: Small thing: pass here means the fallback characters after `\uN` still end up in the text. Our own files are fine (we write `\uc0`), but `\uc1` is the RTF default, so `\u7903 ?` from another program parses as `ở?`. Worth skipping N characters after each escape; or just a comment noting only `\uc0` is handled for now.

Sanjays2402: Good catch, that would have been a real regression for files from other programs. Fixed in 33f5780 rather than leaving a comment.

The parser now tracks `\ucN` (default 1, per the spec) as group state, pushes/restores it alongside the destination on group start/end, and after each `\uN` escape skips that many following characters — both control-word substitutions and plain text, since the fallback can be either.

Before: `{\*\cxs TPA*}ph\u7903 ?` parsed to `phở?`, and `\uc2 \u7903 ??x` to `ở??x`.
After: `phở` and `ởx`. The existing `\uc0` cases are unchanged, and I added the `\uc1` default case as a load test. `test/test_rtfcre_dict.py` is 184 passed.

mkrnr (approved):
Thank you for this contribution and for implementing the suggestion!

PR #1860, opened 2026-07-26
Fix Traditional Chinese translation on Linux

### LINKED ISSUE #1859: The Chinese translation implicitly falls back to English on Linux
**Describe the bug**

As in the title, we cannot start plover with Chinese on Linux.

**To Reproduce**

In either way, plover launches with English:

```sh
$ LANGUAGE=zh_TW python setup.py launch
$ LANGUAGE=zh_tw python setup.py launch
```

**Expected behavior**

It launches in Chinese.

**Screenshots**

N/A

**Operating system**

 - OS: NixOS
 - Plover Version: HEAD

**Hardware**

N/A

toyboot4e:
## Summary of changes

Change the language from `zh_tw` to `zh_TW`. Probably, it was broken on Linux only, because it uses a case-sensitive file system.

Closes #1859 

### Pull Request Checklist
- [ ] Changes have tests


mkrnr (approved):
Thanks a lot!

PR #1862, opened 2026-08-06
fix: detect ibus/fcitx5 via D-Bus instead of process name

Pandapip1:
## Summary of changes

On certain distributions, certain programs (such as the ibus or fcitx daemons) get wrapped. This causes the process name to be different and the check with the hardcoded name to fail.

This instead checks for the presence of the actual D-Bus APIs, which is the correct way of determining if the service is running or not.

Currently uses raw FFI to avoid pulling in any additional dependencies, LMK if you'd prefer I just bring in one of the python D-Bus libraries (which would simplify things a lot).

### Pull Request Checklist

- [ ] Changes have tests

### DIFF
--- a/news.d/bugfix/1862.linux.md
+++ b/news.d/bugfix/1862.linux.md
@@ -1,0 +1,1 @@
+Fix ibus/fcitx5 detection incorrectly warning about a missing input method by probing D-Bus for the running service instead of matching on process name.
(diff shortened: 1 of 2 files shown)


Pandapip1:
Bump, it's been ~1 month

Pandapip1:
Fixed ruff failure

mkrnr:
Hi, thank you so much for this PR! I'm currently moving so I won't find time for this the next few weeks, but it's not forgotten.

Any review or comment from other contributors is very welcome of course!

PR #1864, opened 2026-08-11
downgrade InvalidReport from logging error to debug

greenwyrt:
## Summary of changes

I have a keyboard that switches between PloverHID and mouse keys.

On Windows, I can use PloverHID and switch with the mouse keys with no errors. 

On Linux, I get the invalid report error if I try to use mouse keys.

In Linux, HID devices share the same path even though they may have multiple usage and usage pages (unlike Windows which seems to generate unique paths). Plover will attempt to read non-PloverHID data packets (mousekey events in my case), and raise InvalidReport errors, generating multiple system notifications. 

The change in log level will stop the notifications, but debug output will still show the exception.

### Pull Request Checklist
- [ ] Changes have tests

### DIFF
--- a/plover/machine/plover_hid.py
+++ b/plover/machine/plover_hid.py
@@ -245,7 +245,7 @@ class PloverHid(ThreadedStenotypeBase):
             try:
                 current = self._parse(report)
             except InvalidReport:
-                log.error("invalid report")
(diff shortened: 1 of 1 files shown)


mkrnr (approved):
Thanks! I'd say we don't need a news entry for this fix.

PR #1868, opened 2026-09-18
Port to xcffib

user202729:
Requires https://github.com/tych0/xcffib/pull/205 . DO NOT MERGE until new xcffib version available.

Though, as mentioned in that pull request, it _might_ be possible to use `xkbcommon` instead, which Plover already depends on.

----

## Summary of changes

Delete dependency on python-xlib, depend on xcffib instead.

Reason: python-xlib is not very active nowadays, and it has a few bugs/missing features that I come across.

With a nontrivial amount of LLM assistance. Changes look relatively mechanical (some magic constants come from `X11/X.h` that `xcffib` doesn't include, such as https://github.com/tych0/xcffib/issues/206), but I should test it first.

### Pull Request Checklist
- [ ] Changes have tests


user202729:
Build failure is puzzling. I'd expect a import failure but...?

> missing URL for zsync2

Edit: looks like transient failure, passes now.

user202729:
Fixed LLM hallucinations and tested by hand the keyboard capture/uncapture part.

I'm not entirely sure what the wmctrl part is about (tests done on a window manager that doesn't support that feature, so it returns None…?), but it at least doesn't error out. Second force-push is to fit code in (restrictive) 80-column limit.

Update: wmctrl also manually (programmatically) tested now.

